PR #70, opened 2024-07-04
Allow deletion of the `ServerClaim` when `Server` does not exist

hardikdr:
# Proposed Changes

Allow deletion of the ServerClaim incase Server doesnt exist. 

Fixes #

### DIFF
--- a/internal/controller/serverclaim_controller.go
+++ b/internal/controller/serverclaim_controller.go
@@ -62,7 +62,10 @@ func (r *ServerClaimReconciler) delete(ctx context.Context, log logr.Logger, cla
 
 	server := &metalv1alpha1.Server{}
 	if err := r.Client.Get(ctx, client.ObjectKey{Name: claim.Spec.ServerRef.Name}, server); err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to get server: %w", err)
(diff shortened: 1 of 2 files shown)


internal/controller/serverclaim_controller.go
```diff
@@ -61,7 +61,7 @@ func (r *ServerClaimReconciler) delete(ctx context.Context, log logr.Logger, cla
 	log.V(1).Info("Deleting server claim")
 
 	server := &metalv1alpha1.Server{}
-	if err := r.Client.Get(ctx, client.ObjectKey{Name: claim.Spec.ServerRef.Name}, server); err != nil {
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: claim.Spec.ServerRef.Name}, server); err != nil && !apierrors.IsNotFound(err) {
```
afritzler, line 64: I would add here an if statement + logging. Something along the line: `log.Info("Server gone")`

afritzler (changes requested):

internal/controller/serverclaim_controller.go
hardikdr: Sounds good, made changes !!

afritzler (approved):

PR #84, opened 2024-07-16
Add support for inline BMC access in `Server` reconciler

### LINKED ISSUE #66: Implement inline `BMC` access for `Server`
# Summary

The `Server` resource has 2 ways to interact with a `BMC`: you either reference a `BMC` object via `.spec.bmcRef` or you directly provide the `BMC` access information via `.spec.bmc` in the `Server` spec. Currently only the `bmcRef` is honoured for client instantiation.

afritzler:
# Proposed Changes

Instead of relying on an `Endpoint` and `BMC` resouce to be present to interact with a `Server` you can now register a `Server` directly by providing the BMC access information via `.spec.bmc` in the `Server` spec.

Regenerated stale manifests.

Fixes #66

### DIFF
--- a/internal/controller/bmcutils.go
+++ b/internal/controller/bmcutils.go
@@ -8,8 +8,6 @@ import (
 	"encoding/base64"
 	"fmt"
 
-	"k8s.io/apimachinery/pkg/api/errors"
-
 	metalv1alpha1 "github.com/ironcore-dev/metal-operator/api/v1alpha1"
 	"github.com/ironcore-dev/metal-operator/bmc"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -20,10 +18,7 @@ func GetBMCClientForServer(ctx context.Context, c client.Client, server *metalv1
 		b := &metalv1alpha1.BMC{}
 		bmcName := server.Spec.BMCRef.Name
 		if err := c.Get(ctx, client.ObjectKey{Name: bmcName}, b); err != nil {
-			if errors.IsNotFound(err) {
-				return nil, fmt.Errorf("BMC %q not found", bmcName)
-			}
-			return nil, err
+			return nil, fmt.Errorf("failed to get BMC: %w", err)
 		}
 
 		return GetBMCClientFromBMC(ctx, c, b, insecure)
@@ -45,7 +40,7 @@ func GetBMCClientForServer(ctx context.Context, c client.Client, server *metalv1
 		)
 	}
 
-	return nil, fmt.Errorf("server %s has nor a BMCRef nor an BMC configured", server.Name)
+	return nil, fmt.Errorf("server %s has neither a BMCRef nor a BMC configured", server.Name)
 }
 
 func GetBMCClientFromBMC(ctx context.Context, c client.Client, bmcObj *metalv1alpha1.BMC, insecure bool) (bmc.BMC, error) {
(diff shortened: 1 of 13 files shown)


internal/controller/bmcutils.go
```diff
@@ -15,15 +15,37 @@ import (
...
+			insecure,
+			server.Spec.BMC.Protocol.Name,
+			metalv1alpha1.MustParseIP(server.Spec.BMC.Endpoint),
+			server.Spec.BMC.Protocol.Port,
+			bmcSecret,
+		)
 	}
-	return GetBMCClientFromBMC(ctx, c, bmc, insecure)
+
+	return nil, fmt.Errorf("server %s has nor a BMCRef nor an BMC configured", server.Name)
```
hardikdr, line 48:
```suggestion
	return nil, fmt.Errorf("server %s has neither a BMCRef nor a BMC configured", server.Name)
```

internal/controller/bmcutils.go
```diff
@@ -15,15 +15,37 @@ import (
...
+	if server.Spec.BMC != nil {
+		bmcSecret := &metalv1alpha1.BMCSecret{}
+		if err := c.Get(ctx, client.ObjectKey{Name: server.Spec.BMC.BMCSecretRef.Name}, bmcSecret); err != nil {
+			return nil, fmt.Errorf("failed to get BMC secret: %w", err)
+		}
+
+		return CreateBMCClient(
+			ctx,
+			insecure,
+			server.Spec.BMC.Protocol.Name,
```
hardikdr, line 41: Could this cause nil pointer if not set, same for also Port ?

internal/controller/server_controller.go
```diff
@@ -512,7 +513,7 @@ func (r *ServerReconciler) ensureServerPowerState(ctx context.Context, log logr.
 		return nil
 	}
 
-	bmcClient, err := GetBMCClientFromBMCName(ctx, r.Client, server.Spec.BMCRef.Name, r.Insecure)
+	bmcClient, err := GetBMCClientForServer(ctx, r.Client, server, r.Insecure)
```
hardikdr, line 516: Just curious, if it's intentional, this should be "FromServer" or "ForServer", both are semantically correct though.

internal/controller/bmcutils.go
```diff
@@ -15,15 +15,37 @@ import (
...
 		}
-		return nil, err
+
+		return GetBMCClientFromBMC(ctx, c, b, insecure)
+	}
+
+	if server.Spec.BMC != nil {
+		bmcSecret := &metalv1alpha1.BMCSecret{}
+		if err := c.Get(ctx, client.ObjectKey{Name: server.Spec.BMC.BMCSecretRef.Name}, bmcSecret); err != nil {
+			return nil, fmt.Errorf("failed to get BMC secret: %w", err)
```
hardikdr, line 35: Should we handle the NotFound error gracefully here, just to avoid plenty of reconciles in case the object doesn't exist?

hardikdr (commented):
Looks good, just minor comments inline.

internal/controller/bmcutils.go
afritzler: `.spec.BMC` is checked for nil above and the `Protocol` field is not a pointer ref, so that should be fine.

internal/controller/server_controller.go
afritzler: I somehow found `For` a better naming here 🙈

hardikdr (approved):

internal/controller/bmcutils.go
afritzler: We need to fail fast here as we can't construct the BMC client if we can't get the credentials.

PR #86, opened 2024-07-17
Implement `ServerSelector` in `ServerClaimReconciler`

### LINKED ISSUE #68: Implement `ServerSelector` in `ServerClaimReconciler`
# Summary

We expose the `ServerSelector` in the `ServerClaim` typ to schedule a claim to a `Server` which matches the label selector in the `ServerSelector` field. However currently the selector is not implemented in the claim reconciler.

## Expected outcome

- Implementation handling the `serverSelector` in the `ServerClaim`
- Unit tests ensuring the correct behaviour

aobort:
# Proposed Changes

- added `ClaimedServerRef` to `.status`. Since `.spec.serverRef` is an optional field it's reasonable to store reference to claimed server in the claim's status
- added handling of `.spec.serverSelector` field

Fixes #68


api/v1alpha1/serverclaim_types.go
```diff
@@ -44,11 +44,15 @@ const (
 type ServerClaimStatus struct {
 	// Phase represents the current phase of the server claim.
 	Phase Phase `json:"phase,omitempty"`
+
+	// ClaimedServerRef is a reference to a specific server which is claimed.
+	// This field is optional and can be omitted if the server is to be selected using ServerSelector.
+	ClaimedServerRef *v1.LocalObjectReference `json:"claimedServerRef,omitempty"`
```
afritzler, line 50: Can we name this status field `ServerRef`? This would represent the effective `Server` which the claim is owning.

internal/controller/serverclaim_controller.go
```diff
@@ -251,6 +261,81 @@ func (r *ServerClaimReconciler) removeBootConfigRefFromServerAndPowerOff(ctx con
 	return nil
 }
 
+func (r *ServerClaimReconciler) getClaimableServer(ctx context.Context, log logr.Logger, claim *metalv1alpha1.ServerClaim) (*metalv1alpha1.Server, error) {
```
afritzler, line 264: Can we name this method `claimServer`.

internal/controller/serverclaim_controller.go
```diff
@@ -251,6 +261,81 @@ func (r *ServerClaimReconciler) removeBootConfigRefFromServerAndPowerOff(ctx con
...
+	}
+	// Check the server state and only proceed if it is Available or Reserved
+	if server.Status.State != metalv1alpha1.ServerStateAvailable && server.Status.State != metalv1alpha1.ServerStateReserved {
+		log.V(1).Info("Server not in a claimable state", "Server", server.Name, "ServerState", server.Status.State)
+		return nil, nil
+	}
+	return server, nil
+}
+
+func (r *ServerClaimReconciler) serverByReference(ctx context.Context, log logr.Logger, claim *metalv1alpha1.ServerClaim) (*metalv1alpha1.Server, error) {
```
afritzler, line 289: Can we name this method `claimServerByReference`.

internal/controller/serverclaim_controller.go
```diff
@@ -251,6 +261,81 @@ func (r *ServerClaimReconciler) removeBootConfigRefFromServerAndPowerOff(ctx con
...
+		return nil, err
+	}
+	if !selector.Matches(labels.Set(server.ObjectMeta.Labels)) {
+		log.V(1).Info("Specified server does not match label selector", "Server", server.Name, "Claim", claim.Name)
+		return nil, nil
+	}
+	return server, nil
+}
+
+func (r *ServerClaimReconciler) serverBySelector(ctx context.Context, log logr.Logger, claim *metalv1alpha1.ServerClaim) (*metalv1alpha1.Server, error) {
```
afritzler, line 312: Can we name this method `claimServerBySelector`.

internal/controller/serverclaim_controller.go
```diff
@@ -251,6 +261,81 @@ func (r *ServerClaimReconciler) removeBootConfigRefFromServerAndPowerOff(ctx con
...
+		return nil, err
+	}
+	if claimRef := server.Spec.ServerClaimRef; claimRef != nil && claimRef.UID != claim.UID {
+		log.V(1).Info("Server claim ref UID does not match claim", "Server", server.Name, "ClaimUID", claimRef.UID)
+		return nil, nil
+	}
+	if claim.Spec.ServerSelector == nil {
+		return server, nil
+	}
+	selector, err := metav1.LabelSelectorAsSelector(claim.Spec.ServerSelector)
```
afritzler, line 301: Why do we need to compare the selector labels in the case where we directly want to claim a `Server` via a reference?

internal/controller/serverclaim_controller.go
```diff
@@ -251,6 +261,81 @@ func (r *ServerClaimReconciler) removeBootConfigRefFromServerAndPowerOff(ctx con
...
+func (r *ServerClaimReconciler) serverBySelector(ctx context.Context, log logr.Logger, claim *metalv1alpha1.ServerClaim) (*metalv1alpha1.Server, error) {
+	selector, err := metav1.LabelSelectorAsSelector(claim.Spec.ServerSelector)
+	if err != nil {
+		return nil, err
+	}
+	opts := client.ListOptions{
+		LabelSelector: selector,
+	}
+	serverList := &metalv1alpha1.ServerList{}
+	if err := r.List(ctx, serverList, &opts); err != nil {
```
afritzler, line 321: Please use a `MatchLabels` in the `List` call. That way you don't have to loop over the result set later.

api/v1alpha1/serverclaim_types.go
afritzler, line 50: I guess it would make sense to have here a `name` + `UID` combination here as this is the technical representation of the resource we are claiming. You could avoid situation were a `Server` with the same name but a different `UID` is claimed.

afritzler: We could use here something like this
```
// LocalUIDReference is a reference to another entity including its UID
// +structType=atomic
type LocalUIDReference struct {
	// Name is the name of the referenced entity.
	Name string `json:"name"`
	// UID is the UID of the referenced entity.
	UID types.UID `json:"uid"`
}
```

/ref https://github.com/ironcore-dev/ironcore/blob/main/api/common/v1alpha1/common_types.go#L69

I would declare this type in the `common` file.

internal/controller/serverclaim_controller_test.go
```diff
@@ -146,6 +146,127 @@ var _ = Describe("ServerClaim Controller", func() {
...
+		By("Patching the Server to available state")
+		Eventually(UpdateStatus(server, func() {
+			server.Status.State = metalv1alpha1.ServerStateAvailable
+		})).Should(Succeed())
+
+		By("Ensuring that the Server has the correct claim ref")
+		Eventually(Object(server)).Should(SatisfyAll(
+			HaveField("Spec.ServerClaimRef.Name", claim.Name),
+			HaveField("Spec.Power", metalv1alpha1.PowerOff),
+			HaveField("Status.State", metalv1alpha1.ServerStateReserved),
```
afritzler, line 249: Shouldn't we also compare the `Status` field here for the correct `Server`?

afritzler (changes requested):
Thanks for the PR @aobort! Please find my comments above.

The core question I have here:
When I claim a `Server` via a Ref do I want to also allow a `Selector` or do we ignore it? My understanding of a Ref is: if the `Server` is `Available` - claim it! No matter what it's labels are. On the other side the `ServerSelector` allows you to select the next best `Server` which has the correct set of labels. So the Ref should have precedence over the selector. On the other side if there is no Ref, try to use the selector.

internal/controller/serverclaim_controller.go
```diff
@@ -61,9 +62,18 @@ func (r *ServerClaimReconciler) reconcileExists(ctx context.Context, log logr.Lo
 
 func (r *ServerClaimReconciler) delete(ctx context.Context, log logr.Logger, claim *metalv1alpha1.ServerClaim) (ctrl.Result, error) {
 	log.V(1).Info("Deleting server claim")
+	if claim.Status.ClaimedServerRef == nil {
```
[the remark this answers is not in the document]
aobort: `.spec.serverRef` is an optional field, hence it will be nil, if the server selection is made using `.spec.serverSelector`

internal/controller/serverclaim_controller.go
aobort: That's a good question. I'm not sure whether we should check labels or not. Let's say, it was done to initiate discussion. It's perfectly fine for me to remove this part.

internal/controller/serverclaim_controller_test.go
aobort: We definitely should. Missed check.

aobort:
> Thanks for the PR @aobort! Please find my comments above.
> 
> The core question I have here: When I claim a `Server` via a Ref do I want to also allow a `Selector` or do we ignore it? My understanding of a Ref is: if the `Server` is `Available` - claim it! No matter what it's labels are. On the other side the `ServerSelector` allows you to select the next best `Server` which has the correct set of labels. So the Ref should have precedence over the selector. On the other side if there is no Ref, try to use the selector.

I'd prefer not to make it overcomplicated and stick to two options:
- use reference if set
- use selector if reference is not set
without implementing complex decision making like "if server by reference is not claimable, then let's try to find another one by selector"

afritzler:
You are right: we should stick to the following semantic: 
- If `.spec.serverRef` and `.spec.serverSelector` is both set, we expect that those labels are present on the `Server`
- If only `.spec.serverRef` is set we just pick the `Server` if it is in an `Available` state
- If only `.spec.serverSelector` is set, we `List` all matching `Servers` and pick the first `Available` one

internal/controller/serverclaim_controller.go
aobort: Resolved according to agreed approach https://github.com/ironcore-dev/metal-operator/pull/86#issuecomment-2233025341

internal/controller/serverclaim_controller.go
aobort: Used `client.MatchingLabelsSelector` instead of `client.MatchingLabels`. The "matcher" was chosen due to the fact, that `client.MatchingLabels` is an alias to the `map[string]string` type, however label selector might also contain `MatchExpressions` field, which is a list of more complex structs. Thus `client.MatchingLabelsSelector`, which has nested `labels.Selector` field would fit better to filter out fitting server objects.

afritzler:
Can you also run `make docs` to regenerate the API reference documentation?

aobort:
@afritzler  Test (not included into PR) for the case when neither `.spec.serverRef` nor `.spec.serverSelector` defined fails due to ginkgo runs tests in parallel:

it is impossible to define which server will be claimed in case ref and selector are empty. Thus I've made a try to just get server from the claim after `.spec.serverRef` is fulfilled - works fine. However on some runs - I've run tests 5-10 times in a row - controller claims server which was supposed to be claimed in another test case and that test case fails.

In total: some test case might fail occasionally. The solution which works is to:
1. add `Ordered` option to the test suite. This will grant serial run for all nested containers, so the test cases will run one by one;
2. destroy servers after each test. This will grant absence of side effects on any objects which are not belong to current test case.

afritzler:
To test the case, that `.spec.serverRef` and `.spec.serverSelector` are empty, just create a couple of `Server` resources and ensure, that the `.spec.serverRef` has been mutated to a non `nil` value. Also you could then try to get the `.spec.serverRef` object and ensure that it's state is set to `Reserved`.

In general each suite test (`*_test.go`) should ensure the correct clean up of any `Server` which has been created or derived in any of the `It` statements.

internal/controller/serverclaim_controller.go
```diff
@@ -61,18 +62,39 @@ func (r *ServerClaimReconciler) reconcileExists(ctx context.Context, log logr.Lo
...
+	if modified, err := clientutils.PatchEnsureNoFinalizer(ctx, r.Client, claim, ServerClaimFinalizer); !apierrors.IsNotFound(err) || modified {
+		return ctrl.Result{}, err
+	}
+	log.V(1).Info("Ensured that the finalizer has been removed")
+
+	log.V(1).Info("Deleted server claim")
+	return ctrl.Result{}, nil
+}
+
+func (r *ServerClaimReconciler) finalize(ctx context.Context, log logr.Logger, claim *metalv1alpha1.ServerClaim) error {
```
afritzler, line 82: Can we call this method `cleanupAndShutdownServer`?

afritzler (changes requested):
The PR looks good to me now. Just one minor naming issue I have.

aobort:
> The PR looks good to me now. Just one minor naming issue I have.

@afritzler fixed. Pls, review

afritzler (approved):

PR #89, opened 2024-07-24
Expose BIOS configuration in `Server` type

### LINKED ISSUE #85: Expose BIOS configuration in `Server` type
# Summary

We need to extend the `Server` type by supporting the following BIOS related settings:

- Specify the boot order of a `Server`
- Change arbitrary BIOS settings

The `Server` type will expose two new fields which will allow the user to specify those settings:

- `.spec.bootOrder`
- `.spec.BIOS`

Upon `Server` reconciliation the current BIOS settings have to be read out via RedFish and put into the `Server` status. In case a drift is detected, e.g. the `.spec.BIOS` does not match the `.status.BIOS` settings, a `condition` "Restart/Reboot needed" has to be added to the `conditions` field of the `Server` status.

An example manifest can be found below:

```
kind: Server
metadata:
  name: foo
  annotations: 
    metal.ironcore.dev/op: PowerCycle ## set by somebody else
spec:
  bmcRef:
  #	-> some BMC of which we know the vendor/family
  bootOrder:
  - name: foo
    priority: 1
    type: "netboot"
  - name: disk
    priority: 100 
    type: "disk (hd0)"
  BIOS: # danger zone!!!
  - version: 1.0.1
    settings:
      foo3: bar3
  - version: 1.0.0
    settings:
      foo: bar
      foo2: bar2
status:
  BIOS:
    version: 1.0.0
    settings: # read first top prio settings
      foo: bar
      foo2: bar2
  condition:
    type: "Reboot needed" # if drift is detected
```

## Tasks:

- Extend the current `Server` type with the fields mentioned above
- Implement the corresponding `Reconciler` logic to reflect those changes in the BMC
- Provider unit tests ensuring the correct behavior

stefanhipfel:
# Proposed changes

- Add BIOS configuration to `Server` type
- Implement `ServerReconciler` logic to set BIOS settings
- Improve boot order handling in `Server` type

Fixes #85


bmc/redfish.go
```diff
@@ -46,45 +49,30 @@ func (r *RedfishBMC) Logout() {
 }
 
 // PowerOn powers on the system using Redfish.
-func (r *RedfishBMC) PowerOn(systemID string) error {
-	service := r.client.GetService()
-	systems, err := service.Systems()
+func (r *RedfishBMC) PowerOn(systemUUID string) error {
+	system, err := r.getSystemByID(systemUUID)
```
afritzler, line 53: Can we name this method `getSystemByUUID`?

api/v1alpha1/server_types.go
```diff
@@ -55,6 +55,18 @@ type BMCAccess struct {
...
 
+type BootOrder struct {
+	Name     string `json:"name"`
+	Priority int    `json:"priority"`
+	Device   string `json:"device"`
+}
+
+type BIOSSettings struct {
+	Version string `json:"version"`
+	//maybe use map[string]intstr.IntOrString?!; interface not possible
```
afritzler, line 66: I am not sure that this will be serializable and if we can generate a proper OpenAPI spec out of this. We can check if kube is using `IntOrString` somewhere in their types.

bmc/redfish_local.go
```diff
@@ -182,3 +182,23 @@ func (r RedfishLocalBMC) GetManager() (*Manager, error) {
 
 	return nil, err
 }
+
+func (r *RedfishLocalBMC) GetBootOrder(systemUUID string) ([]string, error) {
```
afritzler, line 186: Would it make sense to also implement the BIOS and boot order settings for the local provider? That way we could add unit tests against the BMC emulator and have at least some test coverage here.

bmc/redfish.go
```diff
@@ -166,27 +145,144 @@ func (r *RedfishBMC) GetManager() (*Manager, error) {
...
+		}
+	}
+	// filter out immutable, readonly and hidden attributes
+	filtered := make(map[string]RegistryEntryAttributes)
+	for _, entry := range biosRegistry.RegistryEntries.Attributes {
+		if !entry.Immutable && !entry.ReadOnly && !entry.Hidden {
+			filtered[entry.AttributeName] = entry
+		}
+	}
+	//TODO: add more tyes like maps and Enumerations
```
afritzler, line 254: Typo in the comment.

afritzler (changes requested):
Thanks @stefanhipfel for your PR. Please find my comments below. Could you also rebase your PR branch please.

bmc/redfish_local.go
stefanhipfel: sure I am not done. forgot to mark this as a draft

api/v1alpha1/server_types.go
stefanhipfel: I think we may need to use rawjson here somehow, if we want to support all types. Dell for example also has type map or enumeration:

```
"Type": "Enumeration",
"Value": [
{
"ValueDisplayName": "Enabled",
"ValueName": "Enabled"
},
{
"ValueDisplayName": "Disabled",
"ValueName": "Disabled"
}
],
```

afritzler: Do we need then some kind of a `type` field in the `BIOSSettings` so we can cast it into a corresponding vendor specific struct?

stefanhipfel: I was thinking about that. It would make it more complicated for the user though. Should we maybe just start with strings? so we would support numbers and strings for now.

Currently I am casting strings to int, if the type is an Integer according to the registry

bmc/redfish_local.go
stefanhipfel: @afritzler I made a proposal to change the bmclient. now you would only need to implement the interface methods if they differ from the default `RedfishBMCClient`. if not overwritten the methods of `RedfishBMCClient` are being called.

internal/controller/server_controller.go
```diff
@@ -613,3 +645,87 @@ func (r *ServerReconciler) enqueueServerByServerBootConfiguration() handler.Even
...
+				}
+			}
+			reset, err := bmcClient.SetBiosAttributes(server.Spec.UUID, diff)
+			if err != nil {
+				return err
+			}
+			if reset {
+				server.Status.Conditions = append(server.Status.Conditions, metav1.Condition{
+					Type: "Reboot needed",
+				})
```
aobort, line 718: instead of direct appending condition, I'd be better to use `k8s.io/apimachinery/pkg/api/meta.SetStatusCondition` 

```
SetStatusCondition sets the corresponding condition in conditions to newCondition and returns true if the conditions are changed by this call. conditions must be non-nil.
1. if the condition of the specified type already exists (all fields of the existing condition are updated to newCondition, LastTransitionTime is set to now if the new status differs from the old status)
2. if a condition of the specified type does not exist (LastTransitionTime is set to now() if unset, and newCondition is appended)
```

stefanhipfel: added it

api/v1alpha1/server_types.go
afritzler: So are we ok to use `map[string]string` for now?

api/v1alpha1/server_types.go
```diff
@@ -82,6 +94,9 @@ type ServerSpec struct {
 	// the boot configuration for this server. This field is optional and can be omitted
 	// if no boot configuration is specified.
 	BootConfigurationRef *v1.ObjectReference `json:"bootConfigurationRef,omitempty"`
+
+	BootOrder []BootOrder    `json:"bootOrder,omitempty"`
```
afritzler, line 98: Can you add some documentation to the newly added fields?

api/v1alpha1/server_types.go
```diff
@@ -139,6 +154,8 @@ type ServerStatus struct {
 	// NetworkInterfaces is a list of network interfaces associated with the server.
 	NetworkInterfaces []NetworkInterface `json:"networkInterfaces,omitempty"`
 
+	BIOS BIOSSettings `json:"BIOS,omitempty"`
```
afritzler, line 157: Please add some docs here.

internal/controller/server_controller.go
```diff
@@ -613,3 +645,87 @@ func (r *ServerReconciler) enqueueServerByServerBootConfiguration() handler.Even
...
+	newOrder := []string{}
+	change := false
+	for i, boot := range server.Spec.BootOrder {
+		newOrder = append(newOrder, boot.Device)
+		if order[i] != boot.Device {
+			change = true
+		}
+	}
+	if change {
+		return bmcClient.SetBootOrder(server.Spec.UUID, newOrder)
```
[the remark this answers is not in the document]
afritzler: With "one time boot" you mean "once on next boot"?

afritzler (changes requested):
Some minor issues. Could you also run `make docs` once documentation has been added to the newly added fields?

internal/controller/server_controller.go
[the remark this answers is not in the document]
stefanhipfel: yes, or should it rather be the same boot order on any reboot

afritzler (approved):

hardikdr (approved):

PR #95, opened 2024-07-31
Add `config/dev` kustomize

defo89:
# Proposed Changes
`config/dev` can be used for local dev setup and for now it will disable the `kube-rbac-proxy` sidecar


afritzler:
Should we also remove the `Webhook` configuration for the `dev` kustomization?

defo89:
> Should we also remove the `Webhook` configuration for the `dev` kustomization?

At least for my `tilt` case it does not matter since cert-manager is part of the stack.

afritzler (approved):
Do we w

Issue #97, opened 2024-08-02
Improve new `Server` discovery handling


defo89:
# Summary

When new `Server` is added and is in `Initial` state we set PXE boot on every reconciliation. While the server boots up or in Dell's case a Lifecycle Controller reconfiguration job is running, we constantly send same instructions to PXE boot. 

Also, when powered server is added and is already powered on, it won't be restarted resulting in no PXE boot.

## Proposal

When new `Server` is added:
- send PXE boot instruction
- set the intermediate state to `Discovery` where we expect server to PXE boot and post details to registry
- if previous step is successful, we set the server to `Available`

## Motivation

Having `Discovery` state we could catch servers that are stuck in PXE boot (network, cabling issues) and if they don't post anything after e.g. 1 hour, we could send them back to `Initial` state.

PR #98, opened 2024-08-02
add discovery state and rework initial power handling

defo89:
# Proposed Changes

- add `Discovery` state where we expect server to PXE boot and post to registry
- On the `Initial` state we power off the server regardless, hence no need to cover server that is [powered on](https://github.com/ironcore-dev/metal-operator/blob/main/internal/controller/server_controller.go#L438)

Fixes https://github.com/ironcore-dev/metal-operator/issues/97

### DIFF
--- a/api/v1alpha1/server_types.go
+++ b/api/v1alpha1/server_types.go
@@ -91,6 +91,9 @@ const (
 	// ServerStateInitial indicates that the server is in its initial state.
 	ServerStateInitial ServerState = "Initial"
 
+	// ServerStateDiscovery indicates that the server is in its discovery state.
+	ServerStateDiscovery ServerState = "Discovery"
+
 	// ServerStateAvailable indicates that the server is available for use.
 	ServerStateAvailable ServerState = "Available"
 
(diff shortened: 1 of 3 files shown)


stefanhipfel (approved):

hardikdr (commented):
I suggest we have a quick offline discussion next week before we go ahead here. I have a few points I’d like to cover:

- Could we leverage the NodeConditions like API instead of introducing a new state? This might streamline the integration without altering the state machine significantly. 

- The rationale for always using PXEBootOnce is to ensure that the Metal Machine consistently fetches the latest PXE script and Ignition during Initial state. Since the OS runs in-memory, especially with Gardenlinux, avoiding PXE could result in the machine booting from disk, which might lead to unexpected behaviors.

- Regarding the issue with machines that are already powered on, perhaps a hint in the Endpoint object about whether Servers should undergo a Firstboot or be directly moved to Available during onboarding might offer a more holistic solution.

afritzler:
@defo89 can you also run `make docs` to regenerate the API reference documentation?

defo89:
> @defo89 can you also run `make docs` to regenerate the API reference documentation?

thanks for the hint - done

hardikdr (approved):

afritzler (approved):

Issue #81, opened 2024-07-12
Dell IDRAC9: BIOS reconfiguration job triggered on each one time boot override


defo89:
**Describe the bug**
Each time `"BootSourceOverrideMode": "UEFI"` is set - despite it's already configured - config job is queued and applied, making the server to reboot twice (1st boot to reconfigure BIOS).

**To Reproduce**
Steps to reproduce the behavior e.g. provide example action definition.

**Expected behavior**
Probably 2 options to fix:
1. Investigate with Dell why reconfig config job is done despite it's already being configured (no config diff)
2. Operator reads out current setting and does not apply `"BootSourceOverrideMode": "UEFI"`, but rather just sets `BootSourceOverrideTarget` and `BootSourceOverrideEnabled`.

**Screenshots**
Configuration - BIOS settings

![image](https://github.com/user-attachments/assets/99a9fc66-9a5a-4c82-a515-694888708443)

On 1st reboot:

![image](https://github.com/user-attachments/assets/1ba1b0f8-98a0-4748-ac3a-edabce5480d8)

![image](https://github.com/user-attachments/assets/318c6bc0-e513-4eba-9c36-045720bb73ef)

**Additional context**
Encountered this on 2 available lab servers:
```
Model: PowerEdge R660
BIOS Version: 1.6.6
iDRAC Firmware Version: 7.10.30.05
```

stefanhipfel:
added it to our internal GCS Compute DELL board

defo89:
@afritzler @hardikdr the Dell boots into lifecycle controller and causes extra delay/instability if we always set [redfish.UEFIBootSourceOverrideMode](https://github.com/ironcore-dev/metal-operator/blob/main/bmc/redfish.go#L113)

Proposal for now would be to read out the setting and only set it if's not equal to UEFI. Later on, I rather see this kind of parameter part of BIOS setting.

Issue #80, opened 2024-07-12
Dell IDRAC9: unreliable boot override


defo89:
**Describe the bug**
Despite one time boot params being set via Redfish, server would still boot from hard drive.

```
2024-07-12T07:50:33.409624088Z 2024-07-12T07:50:33Z	DEBUG	Booted Server in PXE	{"controller": "server", "controllerGroup": "metal.ironcore.dev", "controllerKind": "Server", "Server": {"name":"compute-0-bmc-node003-ap052"}, "namespace": "", "name": "compute-0-bmc-node003-ap052", "reconcileID": "5f7131f8-89a0-4000-949a-35201cb7b18c"}
2024-07-12T07:50:33.419884725Z 2024-07-12T07:50:33Z	DEBUG	Extracted Server details	{"controller": "server", "controllerGroup": "metal.ironcore.dev", "controllerKind": "Server", "Server": {"name":"compute-0-bmc-node003-ap052"}, "namespace": "", "name": "compute-0-bmc-node003-ap052", "reconcileID": "5f7131f8-89a0-4000-949a-35201cb7b18c"}
2024-07-12T07:50:33.420478758Z 2024-07-12T07:50:33Z	DEBUG	Reconciling BMC	{"controller": "bmc", "controllerGroup": "metal.ironcore.dev", "controllerKind": "BMC", "BMC": {"name":"bmc-node003-ap052"}, "namespace": "", "name": "bmc-node003-ap052", "reconcileID": "2f25026e-723f-4b56-9a9f-a25de0282b00"}
2024-07-12T07:50:33.420499572Z 2024-07-12T07:50:33Z	DEBUG	Got Endpoints for BMC	{"controller": "bmc", "controllerGroup": "metal.ironcore.dev", "controllerKind": "BMC", "BMC": {"name":"bmc-node003-ap052"}, "namespace": "", "name": "bmc-node003-ap052", "reconcileID": "2f25026e-723f-4b56-9a9f-a25de0282b00", "Endpoints": "node003-ap052"}
[...]

PR #107, opened 2024-08-13
Do not always set `BootSourceOverrideMode`

defo89:
# Proposed Changes

- to avoid triggering Lifecycle Manager on Dell servers and to improve PXE boot stability, we set `BootSourceOverrideMode` only if it's not already set to `UEFI`

Fixes https://github.com/ironcore-dev/metal-operator/issues/81 and https://github.com/ironcore-dev/metal-operator/issues/80

### DIFF
--- a/bmc/redfish.go
+++ b/bmc/redfish.go
@@ -21,6 +21,16 @@ type RedfishBMC struct {
 	client *gofish.APIClient
 }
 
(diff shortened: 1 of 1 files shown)


afritzler (approved):

hardikdr (approved):
lgtm !!

Issue #108, opened 2024-08-13
Sporadic PR tests failures 


afritzler:
**Describe the bug**

We currently seem to have a sporadic test failure issue when running `make test` as part of a PR check. The issue can't be reproduced locally (both Linux and MacOS). 

Here is an example of such a failed job:
https://github.com/ironcore-dev/metal-operator/actions/runs/10372575072/job/28716111328?pr=107

afritzler:
Fixed with #109

PR #109, opened 2024-08-15
Allow the discovery of `Servers` which are in the powered `On` state.

### LINKED ISSUE #103: Improve new `Server` power handling
# Summary

When we add a `Server` which is already powered on, it won't be restarted resulting in no PXE boot. 

Pros:
- if server which is running productive payload  is added by mistake, there is no disruption right away

Cons:
- when adding multiple servers (e.g. a rack/pod) and they are already powered on, manual action is required to start the machinery

afritzler:
# Proposed Changes

Currently we are not running the discovery of `Servers` which are already powered `On`. A new flag (`--enforce-first-boot`) has been introduced to enforce the discovery of already powered on `Servers`. The default of this flag is set to `false`.

Additionally if we encounter a `Server` with `.spec.power == ""` in the initial state, we mutate the `Server` spec to match the detected `PowerState` in the `Servers` status.

Also some sporadic test failures where addressed.

Fixes #103 #108


internal/controller/serverclaim_controller_test.go
```diff
@@ -244,13 +243,19 @@ var _ = Describe("ServerClaim Controller", func() {
...
 		By("Patching the Server to available state")
-		Eventually(UpdateStatus(&server, func() {
+		Eventually(UpdateStatus(server, func() {
 			server.Status.State = metalv1alpha1.ServerStateAvailable
 		})).Should(Succeed())
 
 		By("Ensuring that the Server has the correct claim ref")
-		Eventually(Object(&server)).Should(SatisfyAll(
-			HaveField("Spec.ServerClaimRef.Name", claim.Name),
+		Eventually(Object(server)).Should(SatisfyAll(
```
afritzler, line 251: @defo89 that was the reason for the sporadic test failures which we encountered. Eventually `.spec.ServerClaimRef.Name` was set but if not the test ran into a `nil` pointer deref.

afritzler:
The sporadic issues in the controller suite seems to be fixed now. But there seems to be still an issue in the registry and agent tests. Somehow it looks like the registry server can't be started as (maybe) the port can't be allocated. 🤔

hardikdr (approved):

PR #114, opened 2024-08-23
Add time triggered `Server` polling

afritzler:
# Proposed Changes

Currently we are missing e.g. PowerState changes on a `Server` if they are manually performed. In order to check periodically for the correct `Server` status we introduce a new flag `--server-resync-interval` (default: 30 sec) which will trigger a `Server` reconciliation according to this value.

Maybe in the future we can have an event source e.g. via BMC event subscriptions to remove the polling of resources. For now that is the easiest way to get potential state drifts resolved.

Additionally the `--requeue-interval` flag has been renamed to `--registry-resync-interval` as it better represents the actual usage of this flag.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -59,11 +59,11 @@ func main() {
 	var registryResyncInterval time.Duration
 	var webhookPort int
 	var enforceFirstBoot bool
-	var resyncInterval time.Duration
+	var serverResyncInterval time.Duration
 
 	flag.DurationVar(&registryResyncInterval, "registry-resync-interval", 10*time.Second,
 		"Defines the interval at which the registry is polled for new server information.")
-	flag.DurationVar(&resyncInterval, "resync-interval", 30*time.Second,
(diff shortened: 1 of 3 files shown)


defo89 (approved):

cmd/manager/main.go
```diff
@@ -56,11 +56,15 @@ func main() {
...
-	var requeueInterval time.Duration
+	var registryResyncInterval time.Duration
 	var webhookPort int
 	var enforceFirstBoot bool
+	var resyncInterval time.Duration
 
-	flag.DurationVar(&requeueInterval, "requeue-interval", 10*time.Second, "Reconciler requeue interval.")
+	flag.DurationVar(&registryResyncInterval, "registry-resync-interval", 10*time.Second,
+		"Defines the interval at which the registry is polled for new server information.")
+	flag.DurationVar(&resyncInterval, "resync-interval", 30*time.Second,
```
hardikdr, line 66: Optional: If it applies only to the `Server` CRs, `server-resync-interval` or `server-requeue-interval` could be helpful.

hardikdr (approved):

PR #120, opened 2024-09-06
Add `Tilt` for local development

defo89:
# Proposed Changes

- `make tilt-up` spins up local kind cluster and enables local development with Tilt
- it also adds redfish-mockup as a sidecar to the manager and posts server UUID to the registry to make Server `Available`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,9 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+GOARCH  := $(shell go env GOARCH)
+GOOS    := $(shell go env GOOS)
+
 # CONTAINER_TOOL defines the container tool to be used for building images.
 # Be aware that the target commands are only tested with Docker which is
 # scaffolded by default. However, you might want to replace it to use other
@@ -181,8 +184,12 @@ LOCALBIN ?= $(shell pwd)/bin
 $(LOCALBIN):
 	mkdir -p $(LOCALBIN)
 
+# curl retries
+CURL_RETRIES=3
+
 ## Tool Binaries
-KUBECTL ?= kubectl
+KUBECTL ?= $(LOCALBIN)/kubectl-$(ENVTEST_K8S_VERSION)
+KUBECTL_BIN ?= $(LOCALBIN)/kubectl
 KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
 ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_VERSION)
@@ -202,6 +209,13 @@ GEN_CRD_API_REFERENCE_DOCS_VERSION ?= v0.3.0
 kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
 $(KUSTOMIZE): $(LOCALBIN)
 	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v5,$(KUSTOMIZE_VERSION))
+
+.PHONY: kubectl
+kubectl: $(KUBECTL) ## Download kubectl locally if necessary.
+$(KUBECTL): $(LOCALBIN)
+	curl --retry $(CURL_RETRIES) -fsL https://dl.k8s.io/release/v$(ENVTEST_K8S_VERSION)/bin/$(GOOS)/$(GOARCH)/kubectl -o $(KUBECTL)
+	ln -sf "$(KUBECTL)" "$(KUBECTL_BIN)"
+	chmod +x "$(KUBECTL_BIN)" "$(KUBECTL)"
 
 .PHONY: controller-gen
 controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary.
@@ -241,3 +255,22 @@ GOBIN=$(LOCALBIN) go install $${package} ;\
 mv "$$(echo "$(1)" | sed "s/-$(3)$$//")" $(1) ;\
 }
 endef
+
+## --------------------------------------
+## Tilt / Kind
+## --------------------------------------
+
+KIND_CLUSTER_NAME ?= metal
+
+.PHONY: kind-create
+kind-create: $(ENVTEST) ## create metal kind cluster if needed
+	./scripts/kind-with-registry.sh
+
+.PHONY: kind-delete
+kind-delete: ## Destroys the "metal" kind cluster.
+	kind delete cluster --name=$(KIND_CLUSTER_NAME)
+	docker stop kind-registry && docker rm kind-registry
+
+.PHONY: tilt-up
+tilt-up: $(ENVTEST) $(KUSTOMIZE) kind-create ## start tilt and build kind cluster if needed
+	EXP_CLUSTER_RESOURCE_SET=true tilt up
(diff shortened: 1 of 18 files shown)


Makefile
```diff
@@ -241,3 +256,26 @@ GOBIN=$(LOCALBIN) go install $${package} ;\
...
+kind-delete: ## Destroys the "metal" kind cluster.
+	kind delete cluster --name=$(KIND_CLUSTER_NAME)
+	docker stop kind-registry && docker rm kind-registry
+
+.PHONY: tilt-up
+tilt-up: $(KUSTOMIZE) $(KUBECTL) kind-create ## start tilt and build kind cluster if needed
+	EXP_CLUSTER_RESOURCE_SET=true tilt up
+
+.PHONY: delete-cluster
+delete-cluster: delete-workload-cluster  ## Deletes the kind cluster "metal"
```
hardikdr, line 280: Do we need this?

config/dev/manager_patch.yaml
```diff
@@ -0,0 +1,35 @@
...
+    spec:
+      containers:
+      - name: manager
+        args:
+          - --health-probe-bind-address=:8081
+          - --metrics-bind-address=127.0.0.1:8080
+          - --leader-elect
+          - --mac-prefixes-file=/etc/macdb/macdb.yaml
+          - --probe-image=ghcr.io/ironcore-dev/metalprobe:latest
+          - --probe-os-image=ghcr.io/ironcore-dev/os-images/gardenlinux:1443
```
hardikdr, line 17: optional, as it's not effective. 
```
ghcr.io/ironcore-dev/os-images/gardenlinux:1443.3
```

bmc/redfish_kube.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get system: %w", err)
+	}
+	system.PowerState = redfish.OnPowerState
+	systemURI := fmt.Sprintf("/redfish/v1/Systems/%s", system.ID)
+	if err := system.Patch(systemURI, system); err != nil {
+		return fmt.Errorf("failed to set power state %s for system %s: %w", redfish.OnPowerState, systemUUID, err)
+	}
+	curlCmd := fmt.Sprintf(
+		`apk add curl && curl -H 'Content-Type: application/json' -d '{"SystemUUID":"%s"}' -X POST %s`,
```
hardikdr, line 75: Optional: A dummy network interface could also be posted here. 😎

hardikdr (approved):
Awesome local dev feature, thanks a bunch! 🎉

Makefile
defo89: good catch! `kind-delete` is already there and `delete-cluster` is not needed

config/dev/manager_patch.yaml
defo89: I will fix it regardless

internal/controller/server_controller.go
```diff
@@ -638,6 +641,10 @@ func (r *ServerReconciler) ensureServerPowerState(ctx context.Context, log logr.
 		server.Spec.Power == metalv1alpha1.PowerOff {
 		powerOp = powerOpOff
 	}
+	if server.Status.State == metalv1alpha1.ServerStateDiscovery &&
```
afritzler, line 644: Can we think of a way to keep local dev specific logic out of the main reconciliation loop?

internal/kube/kube.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"fmt"
+	"os"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// CreateKubeConfig creates a Kubernetes client configuration
+func CreateKubeConfig() (*rest.Config, error) {
```
afritzler, line 15: Can we construct here a `client.Client` instead of a REST config? That will make the usage up the chain easier as well as you only have to do `c.Get(ctx, ....)`.

afritzler (changes requested):

Makefile
```diff
@@ -203,6 +211,13 @@ kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
 $(KUSTOMIZE): $(LOCALBIN)
 	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v5,$(KUSTOMIZE_VERSION))
 
+.PHONY: kubectl
+kubectl: $(KUBECTL) ## Download kubectl locally if necessary.
+$(KUBECTL): $(LOCALBIN)
+	curl --retry $(CURL_RETRIES) -fsL https://dl.k8s.io/release/$(KUBECTL_VERSION)/bin/$(GOOS)/$(GOARCH)/kubectl -o $(KUBECTL)
```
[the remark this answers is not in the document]
afritzler: I like the self containing notion of this though :)

Makefile
```diff
@@ -192,6 +199,7 @@ GEN_CRD_API_REFERENCE_DOCS ?= $(LOCALBIN)/gen-crd-api-reference-docs-$(GEN_CRD_A
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.3.0
+KUBECTL_VERSION ?= v1.29.4
```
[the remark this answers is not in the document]
defo89: sure, will flip it

Makefile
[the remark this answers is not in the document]
defo89: yep, we do a local install to ./bin and don't mess with the host system, it's just that I could not find a way to `go install` the kubectl binary, so had to fall back to https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/

internal/kube/kube.go
defo89: true, I missed this one

internal/controller/server_controller.go
defo89: It's tricky because mockup server is always powered on real fast, and even `--enforce-first-boot` does not help. I have pushed a rework that is specific to this dev UUID. Works for now but we may want to revisit if we even want to spin up multiple mockups servers in kube.

defo89: actually found a simple way - just use `SetPXEBootOnce` instead of `PowerOn`. That does the trick

afritzler: Ok, it seems as this will take a while until we can use `go install` for `kubectl` https://github.com/kubernetes/kubectl/issues/1140

internal/controller/server_controller.go
```diff
@@ -35,6 +35,7 @@ const (
 	DefaultIgnitionSecretKeyName  = "ignition"
 	DefaultIgnitionFormatKey      = "format"
 	DefaultIgnitionFormatValue    = "fcos"
+	DefaultKubeNamespace          = "default"
```
afritzler, line 38: Can we move this default to `bmcutils`?

afritzler (changes requested):

afritzler:
@defo89 can you also add a section to https://github.com/ironcore-dev/metal-operator/blob/main/docs/dev_setup.md on how to use the Tilt setup?

defo89:
> @defo89 can you also add a section to https://github.com/ironcore-dev/metal-operator/blob/main/docs/dev_setup.md on how to use the Tilt setup?

Sure, I added some bits - pls have a [look](https://github.com/ironcore-dev/metal-operator/blob/15ef09a3b53ea740bc8275ba028d6247cbfbf397/docs/dev_setup.md#run-the-local-tilt-development-environment)

internal/controller/server_controller.go
defo89: default moved

afritzler (approved):

Issue #121, opened 2024-09-09
Server `.status.powerState` is not polled frequently


defo89:
**To Reproduce**
Create `ServerClaim` that binds to the server, wait for server to deploy, remove ServerClaim and observe that Servers `powerState` is not updated for a longer time

afritzler:
Maybe we should do a `wait.Until` before we patch the `PowerState` of the `Server`.

stefanhipfel:
that is odd, since in the server reconcile we check `server.Spec.ServerClaimRef == nil` and patch the state to `Available`. In the next loop the `updateServerStatus` is called which should check for the powerState.

In the `handleAvailableState` function we turn off the server. This function however never returns true to be re-queued.

PR #124, opened 2024-09-12
Do not cancel context before ticker ticks

defo89:
should fix https://github.com/ironcore-dev/metal-operator/issues/121

### DIFF
--- a/internal/controller/server_controller.go
+++ b/internal/controller/server_controller.go
@@ -12,6 +12,8 @@ import (
 	"sort"
 	"time"
 
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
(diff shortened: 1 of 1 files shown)


stefanhipfel:
looks good to me.

quick question, isn't the RegistryResyncInterval used in the wrong place here? it is used to re-queue to server reconciliation:

```
requeue, err := r.ensureServerStateTransition(ctx, log, server)
if requeue && err == nil {
    return ctrl.Result{Requeue: requeue, RequeueAfter: r.RegistryResyncInterval}, nil
}
```

defo89:
closing, superseded by https://github.com/ironcore-dev/metal-operator/pull/131

PR #125, opened 2024-09-12
Add handling of `Reset` ad-hoc operation to `Server` type

stefanhipfel:
very simple implementation, which does not check if the reset api call to the bmc was successful or not.

future improvements:
use Conditions and redfish Events to track the progress of a reset call

### DIFF
--- a/internal/controller/server_controller.go
+++ b/internal/controller/server_controller.go
@@ -832,12 +832,12 @@ func (r *ServerReconciler) handleAnnotionOperations(ctx context.Context, log log
 	if err := bmcClient.Reset(server.Spec.UUID, redfish.ResetType(operation)); err != nil {
 		return false, fmt.Errorf("failed to reset server: %w", err)
 	}
-	log.V(1).Info("Successfully executed operation", "Operation", operation)
+	log.V(1).Info("Operation completed", "Operation", operation)
 	serverBase := server.DeepCopy()
 	delete(annotations, metalv1alpha1.OperationAnnotation)
 	server.SetAnnotations(annotations)
 	if err := r.Patch(ctx, server, client.MergeFrom(serverBase)); err != nil {
-		return false, fmt.Errorf("failed to patch Server Annotations: %w", err)
+		return false, fmt.Errorf("failed to patch server annotations: %w", err)
 	}
 	return true, nil
 }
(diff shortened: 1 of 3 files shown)


internal/controller/server_controller.go
```diff
@@ -812,6 +817,31 @@ func (r *ServerReconciler) applyBiosSettings(ctx context.Context, log logr.Logge
 	return nil
 }
 
+func (r *ServerReconciler) handleAnnotionOperations(ctx context.Context, log logr.Logger, server *metalv1alpha1.Server) (bool, error) {
+	annotations := server.GetAnnotations()
+	operation, ok := annotations[metalv1alpha1.OperationAnnotation]
```
afritzler, line 822: Can we do this in one line?

internal/controller/server_controller.go
```diff
@@ -812,6 +817,31 @@ func (r *ServerReconciler) applyBiosSettings(ctx context.Context, log logr.Logge
...
+	bmcClient, err := GetBMCClientForServer(ctx, r.Client, server, r.Insecure)
+	if err != nil {
+		return false, fmt.Errorf("failed to create BMC client: %w", err)
+	}
+	defer bmcClient.Logout()
+	log.V(1).Info("Handling operation", "Operation", operation)
+	if err := bmcClient.Reset(server.Spec.UUID, redfish.ResetType(operation)); err != nil {
+		return false, fmt.Errorf("failed to reset server: %w", err)
+	}
+	log.V(1).Info("Successfully executed operation", "Operation", operation)
```
afritzler, line 835: Can we use here instead:
```
log.V(1).Info("Operation completed", "Operation", operation)
```

afritzler (changes requested):
Thanks @stefanhipfel. Just some minor things.

afritzler:
Do we also want to ad this ad-hoc handling to the `ServerClaim`? That way a user can powercycle the `Server` he claimed as he has typically no RBAC to modify the `Server` annotations directly.

internal/controller/server_controller.go
stefanhipfel: we could, but I am using the annotations variable in line 822/837/838

so I would need to call server.GetAnnotations() instead of the variable.

stefanhipfel:
> Do we also want to ad this ad-hoc handling to the `ServerClaim`? That way a user can powercycle the `Server` he claimed as he has typically no RBAC to modify the `Server` annotations directly.

I think so yes, but we should restrict the operations a user can do

afritzler:
Fair point. Lets keep this PR for `Server` operations only this feature for `ServerClaims` later.

afritzler (approved):

PR #128, opened 2024-09-13
Wait for `Server` power transition

### LINKED ISSUE #121: Server `.status.powerState` is not polled frequently

afritzler:
# Proposed Changes

After changing the powerstate of the `Server` do a periodic polling until the target powerstate of a `Server` is reached.

Fixes #121

### DIFF
--- a/internal/controller/server_controller.go
+++ b/internal/controller/server_controller.go
@@ -12,6 +12,9 @@ import (
 	"sort"
 	"time"
 
+	"github.com/ironcore-dev/metal-operator/bmc"
+	"k8s.io/apimachinery/pkg/util/wait"
+
 	"github.com/go-logr/logr"
 	"github.com/ironcore-dev/controller-utils/clientutils"
 	metalv1alpha1 "github.com/ironcore-dev/metal-operator/api/v1alpha1"
@@ -308,9 +311,6 @@ func (r *ServerReconciler) handleAvailableState(ctx context.Context, log logr.Lo
 	}
 	log.V(1).Info("Ensured initial boot configuration is deleted")
 
-	if err := r.ensureServerPowerState(ctx, log, server); err != nil {
-		return false, fmt.Errorf("failed to ensure server power state: %w", err)
-	}
 	if err := r.ensureIndicatorLED(ctx, log, server); err != nil {
 		return false, fmt.Errorf("failed to ensure server indicator led: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


afritzler:
We might want to make the polling interval + timeout configurable via flags.

defo89:
The logic in the code is right and it should fit well with baremetal. But this seems to break with redfish mockup server since it's always on and cannot be powered off. 
```
[manager] 2024-09-13T15:41:01Z	DEBUG	Read Server power state	{"controller": "server", "controllerGroup": "metal.ironcore.dev", "controllerKind": "Server", "Server": {"name":"endpoint-sample-system-0"}, "namespace": "", "name": "endpoint-sample-system-0", "reconcileID": "84d99d18-d6b7-456e-8317-15e03f4ee366", "PowerState": "On", "TargetPowerState": "Off"}
```

defo89:
not the best way but maybe calling `waitForServerPowerState` could be enabled via flag?

afritzler:
> not the best way but maybe calling `waitForServerPowerState` could be enabled via flag?

Or we do a POST against the BMC emulator and patch the `PowerState` in the test case.

defo89:
For Tilt I just tested - removing [this](https://github.com/ironcore-dev/metal-operator/blob/main/config/dev/manager_patch.yaml#L20) works and server reaches `Available` state. Would be ok for now

internal/controller/server_controller.go
```diff
@@ -665,16 +665,37 @@ func (r *ServerReconciler) ensureServerPowerState(ctx context.Context, log logr.
...
+			return fmt.Errorf("failed to wait for server power off server: %w", err)
+		}
 	}
 	log.V(1).Info("Ensured server power state", "PowerState", server.Spec.Power)
 
 	return nil
 }
 
+func waitForServerPowerState(ctx context.Context, log logr.Logger, bmcClient bmc.BMC, server *metalv1alpha1.Server, powerState redfish.PowerState) error {
+	if err := wait.PollUntilContextTimeout(ctx, 1*time.Second, 10*time.Second, true, func(ctx context.Context) (done bool, err error) {
```
defo89, line 685: should we bump the timeout to allow for baremetal to properly power on/off? Maybe 2 minutes?

afritzler: Addressed this via https://github.com/ironcore-dev/metal-operator/pull/128/commits/f35ecb925b0b8d7f89bc9fe1124ce49b60d1fda4

defo89 (approved):
lgtm, thanks!

PR #129, opened 2024-09-17
Retry with `ForcePowerOff` if graceful shutdown times out

defo89:
# Proposed Changes
- I observed that in case `Server` is stuck in a failed PXE Boot or in other non-responsive state, `redfish.GracefulShutdownResetType` would not power off the system. 
- So we can retry with `redfish.PushPowerButtonResetType` if above fails.
- Enabled via flag `--enforce-power-off`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot AS manager
+LABEL source_repository="https://github.com/ironcore-dev/metal-operator"
 WORKDIR /
 COPY --from=builder /workspace/manager .
 USER 65532:65532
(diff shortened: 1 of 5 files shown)


defo89:
hmm, need to revisit. For Dell it has worked, but on Lenovo I get:
```
2024-09-17T12:56:20.219583196Z 2024-09-17T12:56:20Z	ERROR	Reconciler error	{"controller": "server", "controllerGroup": "metal.ironcore.dev", "controllerKind": "Server", "Server": {"name":"node006-bb00-system-0"}, "namespace": "", "name": "node006-bb00-system-0", "reconcileID": "1e290486-c86f-4df4-90a5-5c55ab438ae1", "error": "failed to ensure server state transition: failed to ensure server power state: failed to power off server: failed to reset system to power on state: reset type 'PushPowerButton' is not supported by this service"}
```

afritzler (approved):
Lgtm! Maybe @stefanhipfel can also have a lot at this PR.

stefanhipfel:
lgtm as well. 
we could also use metal.ironcore.dev/operation annotation to force a power off. But maybe it is good that this is happening automatically.

defo89:
> lgtm as well. we could also use metal.ironcore.dev/operation annotation to force a power off. But maybe it is good that this is happening automatically.

good point. I'd see `metal.ironcore.dev/operation` annotation as a troubleshooting tool to perform the operation and also ability to force power off for users that will not use this `--enforce-power-off` option.

PR #131, opened 2024-09-18
Requeue for Power state polling and ensuring`ServerClaim` binding

defo89:
# Proposed Changes

- Requeue `Available` and `Reserved` servers to reconcile and poll power states
- To avoid setting PXE boot for all reserved `Servers` we only set it if it's powered off - when server transitions from `Available` (and is powered off) to `Reserved`
- add `ServerClaim` resync interval and requeue - esp to re-check available servers if no server was found for claim

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -61,6 +61,7 @@ func main() {
 	var enforceFirstBoot bool
 	var enforcePowerOff bool
 	var serverResyncInterval time.Duration
+	var claimResyncInterval time.Duration
 	var powerPollingInterval time.Duration
 	var powerPollingTimeout time.Duration
 
@@ -69,8 +70,10 @@ func main() {
 	flag.DurationVar(&powerPollingTimeout, "power-polling-timeout", 2*time.Minute, "Timeout for polling power state")
 	flag.DurationVar(&registryResyncInterval, "registry-resync-interval", 10*time.Second,
 		"Defines the interval at which the registry is polled for new server information.")
-	flag.DurationVar(&serverResyncInterval, "server-resync-interval", 30*time.Second,
+	flag.DurationVar(&serverResyncInterval, "server-resync-interval", 2*time.Minute,
 		"Defines the interval at which the server is polled.")
+	flag.DurationVar(&claimResyncInterval, "claim-resync-interval", 2*time.Minute,
+		"Defines the interval at which the server claim is polled.")
 	flag.StringVar(&registryURL, "registry-url", "", "The URL of the registry.")
 	flag.StringVar(&registryProtocol, "registry-protocol", "http", "The protocol to use for the registry.")
 	flag.IntVar(&registryPort, "registry-port", 10000, "The port to use for the registry.")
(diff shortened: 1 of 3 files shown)


internal/controller/serverclaim_controller.go
```diff
@@ -151,7 +153,7 @@ func (r *ServerClaimReconciler) reconcile(ctx context.Context, log logr.Logger,
 	}
 	if server == nil {
 		log.V(1).Info("No server found for claim")
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: r.ResyncInterval}, nil
```
afritzler, line 156: I am not sure we should do a polling here. My idea was to improve the watch setup and requeue all `ServerClaims` if e.g. `metadata.labels` on a `Server` changes.

internal/controller/serverclaim_controller.go
```diff
@@ -180,7 +182,7 @@ func (r *ServerClaimReconciler) reconcile(ctx context.Context, log logr.Logger,
 	log.V(1).Info("Ensured PowerState for Server", "Server", server.Name)
 
 	log.V(1).Info("Reconciled server claim")
-	return ctrl.Result{}, nil
+	return ctrl.Result{RequeueAfter: r.ResyncInterval}, nil
```
afritzler, line 185: Here as well. The desired power state is transported to the `Server` object. And there we already do the polling and ensuring the correct state. Not sure we should enforce a requeueing here on the `ServerClaim` level.

internal/controller/server_controller.go
```diff
@@ -328,11 +328,13 @@ func (r *ServerReconciler) handleReservedState(ctx context.Context, log logr.Log
 	}
 	log.V(1).Info("Server boot configuration is ready")
 
-	if err := r.pxeBootServer(ctx, log, server); err != nil {
-		return false, fmt.Errorf("failed to boot server: %w", err)
+	//TODO: handle working Reserved Server that was suddenly powered off but needs to boot from disk
+	if server.Status.PowerState == metalv1alpha1.ServerOffPowerState {
```
afritzler, line 332: I am not sure I understand this flow here. 🤔

afritzler (changes requested):

internal/controller/serverclaim_controller.go
defo89: that's a good point. We already watch on `Servers`, so we could just add a condition to reconcile unbound `ServerClaims` if any `Server` is labeled. I have added this bit and removed this requeue.

internal/controller/serverclaim_controller.go
defo89: Removed this requeue, this should then be covered by above comment.

internal/controller/server_controller.go
defo89: I would like to requeue `Reserved` server to poll for the power state but I don't want to set PXE boot again if it's already running and powered on. Given that we now poll for power more often, we ensure that for powered on `Server` boot order is not set. 

If server switches from `Available` (power state off) to `Reserved` then we set PXE boot. Maybe there is a better way to achieve this though.

cmd/manager/main.go
```diff
@@ -61,6 +61,7 @@ func main() {
 	var enforceFirstBoot bool
 	var enforcePowerOff bool
 	var serverResyncInterval time.Duration
+	var claimResyncInterval time.Duration
```
afritzler, line 64: Do we want to call this claim resync interal? (as it is only used for the registry sync interval now)

afritzler (changes requested):

internal/controller/server_controller.go
afritzler: Fair point.

cmd/manager/main.go
defo89: hm, we actually don't use it anymore. I will delete this and we can revisit if there is a need again.

afritzler (approved):

afritzler (approved):

PR #132, opened 2024-09-19
Add `Server` model to `.status`

defo89:
# Proposed Changes

- seems we already fetch server model, so this will just add it to `.status`

### DIFF
--- a/api/v1alpha1/bmc_types.go
+++ b/api/v1alpha1/bmc_types.go
@@ -161,7 +161,7 @@ const (
 // +kubebuilder:resource:scope=Cluster
 // +kubebuilder:printcolumn:name="MACAddress",type=string,JSONPath=`.status.macAddress`
 // +kubebuilder:printcolumn:name="MACAddress",type=string,JSONPath=`.status.ip`
-// +kubebuilder:printcolumn:name="Model",type=string,JSONPath=`.status.model`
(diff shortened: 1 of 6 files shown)


api/v1alpha1/bmc_types.go
```diff
@@ -160,8 +160,8 @@ const (
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:scope=Cluster
 // +kubebuilder:printcolumn:name="MACAddress",type=string,JSONPath=`.status.macAddress`
-// +kubebuilder:printcolumn:name="MACAddress",type=string,JSONPath=`.status.ip`
-// +kubebuilder:printcolumn:name="Model",type=string,JSONPath=`.status.model`
+// +kubebuilder:printcolumn:name="IP",type=string,JSONPath=`.status.ip`
+// +kubebuilder:printcolumn:name="Model",type=string,JSONPath=`.status.model`,priority=100
```
afritzler, line 164: Do we want to hide the model behind `-o wide`? I would say lets print it out by default.

afritzler (changes requested):

api/v1alpha1/bmc_types.go
defo89: sounds good, by default makes sense

afritzler (approved):

PR #136, opened 2024-09-25
Add timeout check in discovery state

stefanhipfel:


defo89 (approved):
thanks!

defo89:
lgtm from me, anything to add from your side @afritzler @hardikdr ?

cmd/manager/main.go
```diff
@@ -43,27 +43,31 @@ func init() {
...
+		webhookPort            int
+		enforceFirstBoot       bool
+		enforcePowerOff        bool
+		serverResyncInterval   time.Duration
+		powerPollingInterval   time.Duration
+		powerPollingTimeout    time.Duration
+		discoveryTimeout       time.Duration
+	)
 
+	flag.DurationVar(&discoveryTimeout, "discovery-timeout", 30*time.Minute, "Timeout for BMC discovery")
```
afritzler, line 70: As we are booting the `Server` into a discovery mode using the `metalprobe` agent I am not sure we should call the help hin there "BMC discovery". Maybe something like: "Timeout for discovery boot". would fit better.

internal/controller/server_controller_test.go
```diff
@@ -276,9 +276,15 @@ var _ = Describe("Server Controller", func() {
 		probeAgent := probe.NewAgent(server.Spec.UUID, registryURL, 100*time.Millisecond)
 		go func() {
 			defer GinkgoRecover()
+			time.Sleep(550 * time.Millisecond) // force discovery timeout
```
afritzler, line 279: Can we make this a little bit shorter?

afritzler (changes requested):

internal/controller/server_controller_test.go
stefanhipfel: it will become unpredictable, since it depends on the hardware how quick your reconcile loops are

stefanhipfel: I reduced it, lets see if it will fail on other hardware ;)

SchwarzM: might want to use a constant instead of hardcode ?

afritzler (approved):

PR #138, opened 2024-10-03
PoC: BIOS version & settings management

aobort:
# Proposed Changes

This PR contains PoC implementation for management of server's BIOS version and settings:
- API types
- Controller
- Service managing tasks (scan, apply settings, update version)

This approach fully separate concrete job implementation from the reconciliation flow.

# To discuss

### BootOrder

Do we need to move boot order field from server to serverBIOS CR? Semantically, yes. As it's one of BIOS settings.

### BIOS settings in `.status`

From my perspective, it reasonable to reflect in `.status.bios.settings` **only** those settings which are set in `.spec.bios.settings`. This will make comparison of `.spec` and `.status` much easier.

### Storing bios settings

Do we need custom type for bios setting: 

```go
type SettingsEntry struct {
    // +required
    Name string `json:"name"`
    
    // +required
    Value string `json:"value"`
        
    // +optional
    Unsupported bool `json:"unsupported,omitempty"`
}
```
to avoid attempting to apply settings which are not supported in the specified BIOS version? I.e.:

```yaml
kind: ServerBIOS
metadata:
  name: foo
spec:
  scanPeriodMinutes: 30
  serverRef:
    name: bar
  bios:
    # changing both version and 'unsupported' flag in the same time
    version: 1.0.0  --> 2.0.0
    settings:
    - name: legacyboot
      value: enabled
      unsupported: false  --> true
```

upgrading from version 1.0.0 to 2.0.0 will lead to the bios setting responsible for legacy boot is deprecated/unsupported, thus it is explicitly marked as unsupported and will not be considered during settings applying.


internal/controller/server_controller.go
```diff
@@ -806,58 +779,6 @@ func (r *ServerReconciler) applyBootOrder(ctx context.Context, log logr.Logger,
...
-		if bios.Version == version {
-			versionMatch = true
-			for key, value := range bios.Settings {
-				if res, ok := server.Status.BIOS.Settings[key]; !ok {
-					if !ok || res != value {
-						diff[key] = value
-					}
-				}
-			}
-			reset, err := bmcClient.SetBiosAttributes(server.Spec.UUID, diff)
```
stefanhipfel, line 838: How do we apply bios settings/ boot order now?

aobort: Boot order: there is a question to discuss in the PR's description, whether we need to keep boot order in `Server` CR and keep its processing in server-controller, or we'll move it to `ServerBIOS`. Semantically we should move it to `ServerBIOS`.

Settings: lines 231-245 in serverbios-controller:
```go
func (r *ServerBIOSReconciler) reconcileSettingsUpdate(
	ctx context.Context,
	log logr.Logger,
	serverBIOS *metalv1alpha1.ServerBIOS,
) (ctrl.Result, error) {
	log.V(1).Info("invoking settings update job")
	jobReference, err := r.createJob(ctx, log, serverBIOS, metalv1alpha1.ApplyBIOSSettingsJobType)
	if err != nil {
		return ctrl.Result{}, err
	}
	return r.patchJobReference(ctx, log, serverBIOS, metalv1alpha1.RunningJobRef{
		Type:   metalv1alpha1.ApplyBIOSSettingsJobType,
		JobRef: jobReference,
	})
}
```

The job implementation is in progress, hopefully will finish it today and push to PR.

internal/job/bios_bmc.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (e *BmcBIOSExecutor) Run(ctx context.Context, jobTypeString, serverBIOSRef string) error {
+	jobType := metalv1alpha1.JobType(jobTypeString)
+	switch jobType {
+	case metalv1alpha1.UpdateBIOSVersionJobType:
+		return e.UpdateBIOSVersion(ctx, serverBIOSRef)
+	case metalv1alpha1.ScanBIOSVersionJobType:
+		return e.ScanBIOSVersionSettings(ctx, serverBIOSRef)
+	case metalv1alpha1.ApplyBIOSSettingsJobType:
+		return e.ApplyBIOSSettings(ctx, serverBIOSRef)
```
stefanhipfel, line 33: Same as above.

internal/job/bios_bmc.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+func (e *BmcBIOSExecutor) Run(ctx context.Context, jobTypeString, serverBIOSRef string) error {
+	jobType := metalv1alpha1.JobType(jobTypeString)
+	switch jobType {
+	case metalv1alpha1.UpdateBIOSVersionJobType:
+		return e.UpdateBIOSVersion(ctx, serverBIOSRef)
+	case metalv1alpha1.ScanBIOSVersionJobType:
+		return e.ScanBIOSVersionSettings(ctx, serverBIOSRef)
```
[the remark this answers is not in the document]
aobort: Also, bios update will take time and require reboot. Some bios settings updates will also require reboot.
In separate job we can implement bmc/server health checker and ensure that the process was finished. Does not matter whether it was successful or not. And after that the target object in k8s will be updated and recociled.

It's better approach than artificially requeue object during recociliation if no updates were observed. Also we'll need to implement bmc-bios-whatnot checker and call it during reconciliation.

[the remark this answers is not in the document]
stefanhipfel: why the reconcile can not do those simple checks? Keep reconciling until the server rebooted, scan the settings again to verify if changes have been applied. Classic reconcile loop.
Can we make it optional? We have thousands of servers in ccloud. The operator would spawn thousand of jobs every time for a simple bios scan.
For Bios Updates I can see the need for external jobs, but that is not happening concurrently.
Additionally I prefer that only the operator patches the crd objects.

fmi/server_http.go
```diff
@@ -0,0 +1,130 @@
...
+	})
+
+	return g.Wait()
+}
+
+// registerRoutes registers the server's routes.
+func (s *ServerHTTP) registerRoutes() {
+	s.mux.HandleFunc("/scan", s.scanHandler)
+	s.mux.HandleFunc("/settings-apply", s.settingsApplyHandler)
+	s.mux.HandleFunc("/version-update", s.versionUpdateHandler)
```
stefanhipfel, line 63: idea: would be great to have an endpoint to get all current active tasks from the sync.Map

aobort: @stefanhipfel thanks for idea.
Totally agree. Also we might need to implement queues for requests, rate limiting, retrying on client side and tons of other stuff. Apart from that, I'd also like to generate http API from spec without hardcoding endpoints.

But for now, I think the main goal is to come to agreement - will we proceed further with this design or not.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	serverBIOS *metalv1alpha1.ServerBIOS,
+) (ctrl.Result, error) {
+	server, err := r.getReferredServer(ctx, log, serverBIOS.Spec.ServerRef.Name)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	// if referred server is not in Available state - stop reconciliation
+	if server.Status.State != metalv1alpha1.ServerStateAvailable {
+		return ctrl.Result{RequeueAfter: r.RequeueInterval}, nil
+	}
```
stefanhipfel, line 203: do we need to stop the server reconciliation in the meantime?

aobort: Yes. Bc both BIOS version and settings update will lead to server reboot. So for now we decided that we'll work only with servers which are in "Available" state.

stefanhipfel:
overall I think the POC looks ok

internal/controller/serverbios_controller.go
aobort: > ok but during a serverBios update, someone could still claim the server. Any available server can be claimed

@stefanhipfel that's a good point. However, I do not really like the idea that the serverBIOS controller would change server's state to exclude it from reconciliation. For number of reasons the server's "owner" might want to postpone the update. Especially in terms of version upgrade.
I could add the "Maintenance" state for server and update the controller, so it will check whether server's state is "Maintenance" instead of "Available".

stefanhipfel: @aobort yes i think so as well, that the serverBIOS should not change a server's state. 

Alternative idea: the server reconciler checks on the serverBIOS state.

stefanhipfel:
@aobort @afritzler what do u guys think about taking the bios controller out of the metal operator, same as the boot/ipam operator? We would be more flexible in the future and changes could be implemented independently of the metal operator

aobort:
> @defo89  @aobort @afritzler what do u guys think about taking the bios controller out of the metal operator, same as the boot/ipam operator? We would be more flexible in the future and changes could be implemented independently of the metal operator
> 
> 

I proposed it in the beginning of this "epic". Thus, I agree.

api/v1alpha1/serverbios_types.go
```diff
@@ -0,0 +1,69 @@
...
+	// Settings contains BIOS settings as map
+	// +optional
+	Settings map[string]string `json:"settings,omitempty"`
+}
+
+// ServerBIOSStatus defines the observed state of ServerBIOS
+type ServerBIOSStatus struct {
+	// BIOS contains a bios version and settings.
+	// +optional
+	BIOS BIOSSettings `json:"bios,omitempty"`
```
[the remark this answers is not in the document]
aobort: @stefanhipfel Taking into account:
- possible separating of metal-operator and bios/fw operator
- possible separation of the bios/fw operator and task runner

I'd suggest to proceed as is for now. Then we could add some endpoint to retrieve active tasks for concrete server or something like this. But it will require to store the references for running tasks etc. So I'd avoid adding it right now due to necessity to design this better.

PR #143, opened 2024-10-14
Extract storage details of a `Server` using Redfish

stefanhipfel:


api/v1alpha1/server_types.go
```diff
@@ -42,6 +42,32 @@ const (
...
+	Byte     Capacity = 1
+	KibiByte          = Byte * 1024
+	KiloByte          = Byte * 1000
+	MebiByte          = KibiByte * 1024
+	MegaByte          = KiloByte * 1000
+	GibiByte          = MebiByte * 1024
+	GigaByte          = MegaByte * 1000
+	TebiByte          = GibiByte * 1024
+	TeraByte          = GigaByte * 1000
+)
```
lukasfrank, line 59: What about using `resource.Quantity` from "k8s.io/apimachinery/pkg/api/resource"?

afritzler: This

stefanhipfel: What format do we prefer? DecimalSI or BinarySI?

lukasfrank: If the `Quantity` represents floating-point (like `CPU`) `DecimalSI` is being used. In case of `Storage`, `Memory` and so on you can safely use `BinarySI`.

stefanhipfel: redfish reports capacity in bytes.

so from our mockup redfish 8000000000000 bytes turn into `7812500000Ki`. Was hoping it would automatically turn into Ti.

lukasfrank: The exponent will be picked by the library in order to avoid floating numbers (which could lead to precision errors). 
If you like to println/log in a human readable way, you can still enforce the scale `ScaledValue(resource.Giga)`.

stefanhipfel: I tried that, but it does not affect the outcome of Marshalling. Anyway, the type is now `resource.Quantity`, so other services can easily parse it.

api/v1alpha1/server_types.go
```diff
@@ -192,6 +210,24 @@ type NetworkInterface struct {
...
+// Storage defines the details of one storage device
+type Storage struct {
+	// Name is the name of the storage interface.
+	Name string `json:"name,omitempty"`
+	// Rotational specifies whether the storage device is rotational.
+	Rotational bool `json:"rotational,omitempty"`
+	// Type specifies the type of the storage device.
+	Type string `json:"type,omitempty"`
+	// SizeBytes specifies the size of the storage device in bytes.
+	Capacity resource.Quantity `json:"capacity,omitempty"`
```
afritzler, line 222: Can we use a pointer field here? That way we don't need to deref later when assigning the value.

bmc/redfish.go
```diff
@@ -342,6 +342,56 @@ func (r *RedfishBMC) checkBiosAttributes(attrs map[string]string) (reset bool, e
...
+				Type:       d.DriveFormFactor,
+				SizeBytes:  d.CapacityBytes,
+				Vendor:     d.Manufacturer,
+				Model:      d.Model,
+				State:      d.Status.State,
+			})
+		}
+	}
+	if len(result) == 0 {
+		// if no storage is found, fall back to simpleStorage (outdated storage API)
```
stefanhipfel, line 373: redfish-mockup-server only supports simpleStorage and just returns an empty list for Storage.

internal/controller/server_controller.go
```diff
@@ -386,7 +387,23 @@ func (r *ServerReconciler) updateServerStatus(ctx context.Context, log logr.Logg
...
+		return fmt.Errorf("failed to get storages for Server: %w", err)
+	}
 	serverBase := server.DeepCopy()
+	server.Status.Storages = nil
+	for _, storage := range storages {
+		q := resource.NewQuantity(storage.SizeBytes, resource.BinarySI)
+		server.Status.Storages = append(server.Status.Storages, metalv1alpha1.Storage{
+			Name:     storage.Name,
+			Model:    storage.Model,
+			Capacity: *q,
```
afritzler, line 401: Assign the `resource.NewQuantity` directly here after changing the field value to a pointer (see comment above).

afritzler (changes requested):

internal/controller/server_controller.go
```diff
@@ -386,7 +387,23 @@ func (r *ServerReconciler) updateServerStatus(ctx context.Context, log logr.Logg
 		return fmt.Errorf("failed to get system info for Server: %w", err)
 	}
 
+	storages, err := bmcClient.GetStorages(server.Spec.UUID)
+	if err != nil {
+		return fmt.Errorf("failed to get storages for Server: %w", err)
+	}
 	serverBase := server.DeepCopy()
+	server.Status.Storages = nil
+	for _, storage := range storages {
```
stefanhipfel, line 396: do we need to keep checking storages on every `updateServerStatus`?

afritzler:
Can you run `make docs` to regenerate the API reference documentation to reflect the status change?

bmc/bmc.go
```diff
@@ -105,6 +107,22 @@ type Server struct {
...
+	Rotational bool `json:"rotational,omitempty"`
+	// Type specifies the type of the storage device.
+	Type redfish.FormFactor `json:"type,omitempty"`
+	// SizeBytes specifies the size of the storage device in bytes.
+	SizeBytes int64 `json:"sizeBytes,omitempty"`
+	// Vendor specifies the vendor of the storage device.
+	Vendor string `json:"vendor,omitempty"`
+	// Model specifies the model of the storage device.
+	Model string       `json:"model,omitempty"`
+	State common.State `json:"state,omitempty"`
```
afritzler, line 123: Can we add a comment to this field.

afritzler (changes requested):
Just some minor issue. Once the API reference docs are regenerated we can merge this PR. Looks good from my side.

afritzler (approved):

PR #159, opened 2024-10-31
Claim only `PoweredOff` Servers

defo89:
Also do not perform power changes in `Available` state if server is already powered off.

ServerClaim controller may patch the server in parallel to `spec.PowerState: On` which may result in powering on the server in Available state.

### DIFF
--- a/internal/controller/serverclaim_controller.go
+++ b/internal/controller/serverclaim_controller.go
@@ -351,6 +351,10 @@ func (r *ServerClaimReconciler) claimServerByReference(ctx context.Context, log 
 		log.V(1).Info("Server not in a claimable state", "Server", server.Name, "ServerState", server.Status.State)
 		return nil, nil
 	}
+	if server.Status.PowerState != metalv1alpha1.ServerOffPowerState {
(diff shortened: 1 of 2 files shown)


afritzler (approved):

internal/controller/serverclaim_controller.go
```diff
@@ -385,6 +385,10 @@ func (r *ServerClaimReconciler) claimServerBySelector(ctx context.Context, log l
 			log.V(1).Info("Server not in a claimable state", "Server", s.Name, "ServerState", s.Status.State)
 			continue
 		}
+		if s.Status.PowerState != metalv1alpha1.ServerOffPowerState {
```
afritzler, line 388: Shouldn't we handling this in `claimServerByReference` as well?

afritzler (changes requested):

internal/controller/serverclaim_controller.go
defo89: added it there, too

afritzler (approved):

PR #161, opened 2024-10-31
Use lower-case `Server` UUIDs

### LINKED ISSUE #160: Always use lower-case `Server` UUIDs
# Summary

Currently we rely on the the `BMC` to provide us the `System` (`Server`) UUID. This UUID can be in some cases upper-cased. To ensure a homogenous behavior also in regards to the way the `Ignition` endpoint is being created (lower cased) we should harmonize this by converting the `Server` UUID always to lower case.

afritzler:
# Proposed Changes

Use lower-case `Server` UUIDs.

Fixes #160

### DIFF
--- a/bmc/redfish.go
+++ b/bmc/redfish.go
@@ -399,7 +399,7 @@ func (r *RedfishBMC) getSystemByUUID(systemUUID string) (*redfish.ComputerSystem
 		return nil, err
 	}
 	for _, system := range systems {
-		if system.UUID == systemUUID {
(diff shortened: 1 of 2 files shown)


afritzler:
/cc @hardikdr 

Can we harmonize this:
https://github.com/ironcore-dev/boot-operator/commit/113b4f42a82cfdeb16b67fba71ce3b12b1aa8d9d

defo89 (approved):

PR #162, opened 2024-11-04
Add `metalctl` command line interface

### LINKED ISSUE #146: Add `metalctl` CLI with `move` Command to Migrate CRDs and CRs Between Clusters
## Description
We are introducing a new CLI tool called `metalctl` to interact with the `metal-operator` as part of our `metal-operator` ecosystem. The initial command to be implemented is the `move` command, which will facilitate the migration of ~~Custom Resource Definitions (CRDs) and~~ Custom Resources (CRs) related to `metal-operator` from one Kubernetes cluster to another.

## Goal
- Create a new CLI utility called `metalctl`.
- Integrate the `move` command to migrate all `metal-operator` related ~~CRDs and~~ CRs between two clusters.

### Key Objectives
- Ensure seamless migration of `metal-operator` ~~CRDs and~~ CRs.
- Ensure the migration process handles dependency order (CRDs must be ~~moved~~ present  before CRs).
- Handle failure scenarios gracefully with appropriate error messages and rollback functionality.

### Command: `move`
The `move` command will enable users to specify a source and target Kubernetes cluster for migrating ~~CRDs and~~ CRs associated with the `metal-operator`.

#### Command Options:
- `--source-kubeconfig` (required): Path to the kubeconfig of the source cluster.
- `--target-kubeconfig` (required): Path to the kubeconfig of the target cluster.
~~- `--crds-only` (optional): Flag to migrate only the CRDs without CRs.~~
~~- `--crs-only` (optional): Flag to migrate only the CRs without CRDs.~~
- `--namespace` (optional): Specify a namespace to filter CRDs and CRs to migrate. Defaults to all namespaces if not specified.
- `--dry-run` (optional): Perform a dry run to show what would be moved without executing the migration.
- `--verbose` (optional): Enable verbose logging for detailed output during migration.

### Example Usage:
```bash
metalctl move --source-kubeconfig=~/.kube/source-cluster.yaml \
              --target-kubeconfig=~/.kube/target-cluster.yaml \
              --namespace=metal-namespace
```

### Steps in the move Process:
1. Validate Input: Verify the presence of both `source-kubeconfig` and `target-kubeconfig`.
2. Fetch CRDs from Source Cluster: Use the source kubeconfig to list all CRDs related to the `metal-operator`.
3. Apply CRDs to Target Cluster: Use the target kubeconfig to apply the CRDs. Ensure idempotency in case they already exist.
4. Fetch CRs from Source Cluster: Fetch all CRs of the metal-operator from the source cluster.
5. Apply CRs to Target Cluster: Apply all CRs to the target cluster.
6. Verify Migration: Ensure the CRDs and CRs have been successfully created in the target cluster.
7. Error Handling and Rollback: If any step fails, provide detailed error messages and rollback any partial migrations where applicable.

### Definition of Done
* A new CLI tool `metalctl` is created with a functional move command.
* The move command successfully migrates `metal-operator` CRDs and CRs between clusters.
* Support for `dry-run`, ~~`crds-only`, `crs-only`~~, and `namespace` options.
* Detailed documentation for using the `move` command is included.
* Adequate test coverage for the `move` functionality, including unit tests and integration tests.

### Additional Considerations
* Future commands for `metalctl` can include operations like listing or deleting resources, which should be planned for in the CLI structure.
* `Console access`: Add a `console` command to access the serial console of a server. This will allow users to remotely view or interact with a server's console for debugging and management purposes.

SzymonSAP:
# Proposed Changes

This PR adds `metaclt move` command with it's flags described in a issue #146. 

# To discuss

- How to ensure the CRDs and CRs already existing in the target cluster are the same as in the source targets?
done
- How to ensure the CRDs and CRs have been successfully created in the target cluster?
done

Fixes #146


afritzler:
Thanks @SzymonSAP. Can you open the PR directly against the `main` branch?

SzymonSAP:
> Thanks @SzymonSAP. Can you open the PR directly against the `main` branch?

Sure, done it. Please let me know if the changes are on the right track and share any ideas what edge cases should also be covered.
I wonder about scenario in which there are already some CRDs or CRs in target cluster that are identical to source cluster. In such scenario `metalctl move` could be run successfully but it requires some code changes. Is it a real world use case?

cmd/metalctl/app/move.go
```diff
@@ -0,0 +1,226 @@
...
+	movedObjects := make([]client.Object, 0)
+
+	for _, sourceObj := range sourceObjs {
+		var err error
+		sourceObj.SetResourceVersion("")
+		if err = cl.Create(ctx, sourceObj); err == nil {
+			if crd, isCrd := sourceObj.(*apiextensionsv1.CustomResourceDefinition); isCrd &&
+				slices.IndexFunc(sourceObjs, func(obj client.Object) bool {
+					return obj.GetObjectKind().GroupVersionKind().Kind == crd.Spec.Names.Kind
+				}) != -1 {
```
Nuckal777, line 159: It's hard to read.
```
shouldWait := slices.IndexFunc(...) != -1
if crd, isCrd := ...; isCrd && shouldWait {}
```

cmd/metalctl/app/suite_test.go
```diff
@@ -0,0 +1,160 @@
...
+	Expect(metalv1alpha1.AddToScheme(k8sSchema.Scheme)).NotTo(HaveOccurred())
+
+	//+kubebuilder:scaffold:scheme
+
+	clients.source, err = client.New(cfg, client.Options{Scheme: k8sSchema.Scheme})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(clients.source).NotTo(BeNil())
+
+	// Target client without CRDs
+	testEnv = &envtest.Environment{
```
Nuckal777, line 86: While working, I would prefer 2 variables over shadowing. `sourceEnv` and `targetEnv` or similar.

cmd/metalctl/app/move.go
```diff
@@ -0,0 +1,226 @@
...
+			}
+		}
+	}
+
+	return metalObjects, nil
+}
+
+func clearObjectFields(obj client.Object) map[string]interface{} {
+	so, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	delete(so, "status")
```
Nuckal777, line 119: The status needs to be moved as well.

cmd/metalctl/app/move.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// getMetalObjects returns CRDs and CRs from metal group. CRDs in a returned list are before theirs CRs.
+func getMetalObjects(ctx context.Context, cl client.Client) ([]client.Object, error) {
+	crds := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := cl.List(ctx, crds); err != nil {
+		return nil, fmt.Errorf("couldn't list CRDs: %w", err)
+	}
+
+	metalObjects := make([]client.Object, 0)
```
Nuckal777, line 92: Maybe it is worth to store CRDs and CRs in 2 different slices? That way, type checks/branching while iterating them can be avoided later.

Nuckal777 (changes requested):
Thanks, for looking into this. 👍  Besides some small stuff, I think to larger topics are missing.

Firstly, the `status` of custom resources need to be copied over. This is especially important for `Servers`, where the status contains the lifecycle phase. Not copying the status here, would re-provision running servers (due to the reset into the discovery phase). One needs to use `cl.Status().Update()` to set the status after creation of a resource.

Secondly, the metal-operator uses owner references on many resources. They cannot be moved directly as they contain the UID of the owner, which is assigned on creation at random. To move these the owner needs to be looked up in the target cluster by name to patch the UID. E.g.

```go
var endpoint metalv1alphav1.Endpoint
if err := clients.target.Get(ctx, types.NamespacedName{Name: owner.Name}, &endpoint); err != nil {
	return err
}
cr.OwnerReferences[0].UID = endpoint.UID
```

As a side effect this requires ordering the creation of CRs properly. E.g. `Endpoints` need to be handled first, then `BMC` and `BMCSecrets` and so on. In that regard it may be worth thinking about an "integration" test, which creates an `Endpoint` against the redfish simulator in source cluster with the metal-operator attached and then try moving.

cmd/metalctl/app/move.go
SzymonSAP: done

cmd/metalctl/app/suite_test.go
SzymonSAP: done

cmd/metalctl/app/move.go
SzymonSAP: done

cmd/metalctl/app/move.go
```diff
@@ -0,0 +1,347 @@
...
+		}
+		for _, cr := range CRs.Items { // won't work with go version <1.22
+			crs = append(crs, &cr)
+		}
+	}
+
+	return crs, nil
+}
+
+func clearFields(obj client.Object) map[string]interface{} {
```
Nuckal777, line 119: `map[string]any` is more concise than `map[string]interface{}`. Given how often it appears a type alias may make sense, e.g. `type MapStrAny map[string]any`.

cmd/metalctl/app/move.go
```diff
@@ -0,0 +1,347 @@
...
+
+func crsOwnerReferenceTrees(crs []*unstructured.Unstructured) []*Node {
+	nodeMap := make(map[types.UID]*Node)
+
+	for _, cr := range crs {
+		nodeMap[cr.GetUID()] = &Node{Cr: cr}
+	}
+	roots := []*Node{}
+	for _, cr := range crs {
+		if len(cr.GetOwnerReferences()) == 0 || nodeMap[cr.GetOwnerReferences()[0].UID] == nil {
```
Nuckal777, line 194: Is the `nodeMap[cr.GetOwnerReferences()[0].UID] == nil` for the case that an owner exists on a CR, but is not a metal related resource?

cmd/metalctl/app/move.go
```diff
@@ -0,0 +1,347 @@
...
+		crsTree.Cr.SetResourceVersion("")
+		if err = cl.Create(ctx, crsTree.Cr); err != nil {
+			err = fmt.Errorf("CR %s couldn't be created in the target cluster: %w", crName(crsTree.Cr), err)
+			return
+		}
+		movedCrs = append(movedCrs, crsTree.Cr)
+	}
+
+	for _, crsTree := range crsTrees {
+		err = wait.PollUntilContextTimeout(ctx, time.Second, 30*time.Second, true, func(ctx context.Context) (bool, error) {
```
Nuckal777, line 259: What is the intent of the wait loop?

SzymonSAP: When creating children CRs they should have set owner reference. So optional argument `uid` is passed to `moveCrs` function. It is empty when creating `root` CRs (without owner or with owner not in metal CRs). It should be set when creating children CRs so owner CR must be created on the target cluster and its UID must be passed. This is why there is a wait loop, to retrive uid of an owner.

cmd/metalctl/app/move.go
```diff
@@ -0,0 +1,347 @@
...
+	cleanupErrs := make([]error, 0)
+	for _, obj := range objs {
+		if err := cl.Delete(ctx, obj); err != nil {
+			cleanupErrs = append(cleanupErrs, err)
+		}
+	}
+	return errors.Join(cleanupErrs...)
+}
+
+func moveCrds(ctx context.Context, cl client.Client, sourceCrds []*apiextensionsv1.CustomResourceDefinition) (movedCrds []*apiextensionsv1.CustomResourceDefinition, err error) {
```
afritzler, line 214: Heretic question @Nuckal777: Do we really need this? I would suggest we run a `make install` on the target cluster to install the CRDs into the target environment. That is also the way `clsuterctl` from the CAPI project is doing it. https://cluster-api.sigs.k8s.io/clusterctl/commands/move.html#pivot

afritzler (changes requested):

cmd/metalctl/app/move.go
```diff
@@ -0,0 +1,347 @@
...
+		if crd.Spec.Group == metalv1alphav1.GroupVersion.Group {
+			metalCrds = append(metalCrds, &crd)
+		}
+	}
+
+	return metalCrds, nil
+}
+
+func getMetalCrs(ctx context.Context, cl client.Client, crds []*apiextensionsv1.CustomResourceDefinition) ([]*unstructured.Unstructured, error) {
+	crs := make([]*unstructured.Unstructured, 0)
```
afritzler, line 103: Why not using a `[]client.Object` here and work with the concrete types? I would avoid using `unstructured` whenever possible. As `metalctl` is part of the `metal-operator` the type import should not be a problem.

afritzler (changes requested):

cmd/metalctl/app/move.go
SzymonSAP: Actually, type assertion compares exact type names:
```
interface conversion: interface {} is map[string]interface {}, not app.MapStrAny
```
More information: https://stackoverflow.com/questions/19577423/how-to-cast-to-a-type-alias-in-go

I changed `map[string]interface{}`with `map[string]any`.

cmd/metalctl/app/move.go
SzymonSAP: I used `unstructured` because I have a hard time implementing a loop over each CRD. If I want to do something like:
```
for _, CRList := range[]client.ObjectList{&metalv1alphav1.BMCList{}, &metalv1alphav1.BMCSecretList{}, ...} {
    client.List(CRList)
```
I can't collect its items as `client.ObjectList` is an interface (see https://github.com/kubernetes-sigs/controller-runtime/issues/2697).

So without some nasty casting I should list all CRs one by one and collect its items:
```
bmcList := &metalv1alphav1.BMCList{}
client.List(bmcList)
for _, cr := range bmcList.Items {
	crs = append(crs, &cr)
}

bmcSecretList := &metalv1alphav1.BMCSecretList{}
client.List(bmcSecretList)
for _, cr := range bmcSecretList.Items {
	crs = append(crs, &cr)
}
...
```

This is the solution you want me to use or there is a more clever way to do it?

cmd/metalctl/app/move.go
Nuckal777: It's what you [ordered in the ticket](https://github.com/ironcore-dev/metal-operator/issues/146), but I am fine with expecting the CRDs to present before the move command is issued.

cmd/metalctl/app/move.go
SzymonSAP: After a CR is created it takes some time to get it from api, so `Get` method might return an error at first.

cmd/metalctl/app/move.go
afritzler: @Nuckal777 dropping the CRD handling will make the codepath a lot easier (see comments about the usage of `unstructured`). Let me adjust the issue and I would say lets drop the CRD handling for now.

afritzler: Ok, I updated the issue. We can also drop the `--cr-only` flag as this should be the default (and only) copy operation.

afritzler:
Can you please rebase this PR branch and use https://github.com/ironcore-dev/metal-operator/pull/179 to add the license headers to all net-new go files?

SzymonSAP:
> Can you please rebase this PR branch and use #179 to add the license headers to all net-new go files?

I have pushed requested changes and rebased the PR. Please have a look ;)

Nuckal777:
I had a short alignment with @afritzler today. Here are the 2 main outcomes:
- We don't like to move CRDs with the `move` command. It is expected, that they are present in the target cluster beforehand, either via `make install` or `helm install`. You can remove the related code (Creation of CRs without the repsective CRD fails automatically).
- We keep the `unstructured.Unstructured` approach for moving CRs, because we would like to re-use it for the boot-operator. Could you encapsulate a `Move(ctx context.Context, clients Clients, crsToBeMoved apimachinery.GroupVersionKind)` (or similar) function into a dedicated package/directory, please? After merging here, we can move that package to ironcore-dev/controllerutils.

cmd/metalctl/app/move.go
```diff
@@ -0,0 +1,250 @@
...
+		if len(ownerReferences) == 1 && len(ownerUid) == 1 {
+			ownerReferences[0].UID = ownerUid[0]
+			crsTree.Cr.SetOwnerReferences(ownerReferences)
+		}
+
+		crsTree.Cr.SetResourceVersion("")
+		if err = cl.Create(ctx, crsTree.Cr); err != nil {
+			err = fmt.Errorf("CR %s couldn't be created in the target cluster: %w", crName(crsTree.Cr), err)
+			return
+		}
```
Nuckal777, line 187: I think the status subresource creation is missing.

SzymonSAP: I have added status update after getting CR however in tests `cl.Create` creates object with status so I'm not sure if it will work on real environment. It would be good to test using clusters.

cmd/metalctl/app/move.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func runMove(cmd *cobra.Command, args []string) error {
+	clients, err := makeClients()
+	if err != nil {
+		return err
+	}
+	ctx := cmd.Context()
+	crsSchema := []schema.GroupVersionKind{}
+	for _, crdKind := range []string{"BMC", "BMCSecret", "Endpoint", "Server", "ServerBootConfiguration", "ServerClaim"} {
```
Nuckal777, line 76: Out of interest: Is there a way to get this list automagically?

Nuckal777 (approved):
Thanks. 🥳 I will test this with my local 2 kind cluster setup.

cmdutils/move.go
```diff
@@ -0,0 +1,187 @@
...
+		crsTree.Cr.SetResourceVersion("")
+		if err = cl.Create(ctx, crsTree.Cr); err != nil {
+			err = fmt.Errorf("CR %s couldn't be created in the target cluster: %w", crName(crsTree.Cr), err)
+			return
+		}
+		movedCrs = append(movedCrs, crsTree.Cr)
+	}
+
+	for _, crsTree := range crsTrees {
+		err = wait.PollUntilContextTimeout(ctx, time.Second, 30*time.Second, true, func(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
SzymonSAP: Created const with 100ms

SzymonSAP:
> Sorry, the status of CRs is not copied over. The test pass with regard to the status, because it is never properly set and hence `nil == nil` equals `true`.

I pushed changes with fixed copying status. Thanks for checking that. Now it should work as expected.

Nuckal777 (approved):
Now, it works against my test setup. 👍 @afritzler feel free to merge or take another look.

afritzler (changes requested):
LGTM from my side. Can you please run `make fmt`. Also there some linting issue to be addressed.

afritzler:
Also it would make sense to add new documentation section `Usage/metalctl` and describe the use of the new CLI.

cmdutils/suite_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	clients.Target, err = client.New(targetCfg, client.Options{Scheme: k8sSchema.Scheme})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(clients.Target).NotTo(BeNil())
+
+	By("Starting the registry server")
+	var mgrCtx context.Context
+	mgrCtx, cancel := context.WithCancel(context.Background())
+	DeferCleanup(cancel)
+	registryServer := registry.NewServer("localhost:30000")
```
afritzler, line 108: I would leave out the registry server out of this test suite.

afritzler (changes requested):

cmdutils/suite_test.go
SzymonSAP: done ;)

afritzler:
@SzymonSAP can you add above this line https://github.com/ironcore-dev/metal-operator/blob/main/internal/controller/bmc_controller_test.go#L14

```
- Usage:
    - metalctl: usage/metalctl.md
 ```

And describe in the `metalctl.md` how to install and use the `metalctl` command line tool?

**Comment by SzymonSAP** [comment]
> Also it would make sense to add new documentation section `Usage/metalctl` and describe the use of the new CLI.

I added documentation strongly inspired by https://cluster-api.sigs.k8s.io/clusterctl/commands/move.html

**Review by afritzler — APPROVED** [review]

File: docs/usage/metalctl.md [hunk 14]
```diff
@@ -0,0 +1,53 @@
...
+Custom Resources defined in a single namespace, you can use the `--namespace` flag.
+
+Status and ownership of a metal Custom Resource is also moved. If a metal Custom Resource present on the source cluster
+exists on the target cluster with identical specification it won't be moved and no ownership of this object will be
+set. In case of any errors during the process there will be performed a cleanup and the target cluster will be restored
+to its previous state.
+
+> Warning!: 
+`metalctl move` has been designed and developed around the bootstrap use case described below, and currently this is
+the only use case verified by Cluster API E2E tests.
```
Nuckal777, line 28:
```suggestion
the only use case verified .
```

afritzler (approved):

afritzler (approved):

PR #169, opened 2024-11-08
Migrate to CRD validation rules

### LINKED ISSUE #168: Use CRD validation rules instead of validation webhooks
# Summary

Use [CRD validation rules](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules) (stable since kubernetes v1.29) instead of validation webhooks.

## Basic example

`controller-gen` marker:
```
+kubebuilder:validation:XValidation field
	marks a field as requiring a value for which a given
	[fieldPath=<string>]
	[message=<string>]
	[messageExpression=<string>]
	[reason=<string>]
	rule=<string>
```

## Motivation

Webhooks have a bit more maintenance + operational effort.

Nuckal777:
# Proposed Changes

- Use CRD validation rules instead of validation webhooks
- Make BMC `.spec.access` and `.spec.endpointRef` immutable

Fixes #168


afritzler (approved):
Thanks a lot @Nuckal777!

PR #173, opened 2024-11-08
adds TotalSystemMemory to server status

stefanhipfel:
adds TotalSystemMemory to the server status from bmc (redfish)

### DIFF
--- a/bmc/bmc.go
+++ b/bmc/bmc.go
@@ -188,16 +188,17 @@ type Processor struct {
 
 // SystemInfo represents basic information about the system.
 type SystemInfo struct {
-	Manufacturer      string
-	Model             string
-	Status            common.Status
-	PowerState        redfish.PowerState
-	NetworkInterfaces []NetworkInterface
-	Processors        []Processor
-	SystemUUID        string
-	SerialNumber      string
-	SKU               string
-	IndicatorLED      string
+	Manufacturer           string
+	Model                  string
+	Status                 common.Status
+	PowerState             redfish.PowerState
+	NetworkInterfaces      []NetworkInterface
+	Processors             []Processor
+	TotalSystemMemoryBytes int64
+	SystemUUID             string
+	SerialNumber           string
+	SKU                    string
+	IndicatorLED           string
 }
 
 // Manager represents the manager information.
--- a/bmc/redfish.go
+++ b/bmc/redfish.go
@@ -182,14 +182,15 @@ func (r *RedfishBMC) GetSystemInfo(systemUUID string) (SystemInfo, error) {
 	}
 
 	return SystemInfo{
-		SystemUUID:   system.UUID,
-		Manufacturer: system.Manufacturer,
-		Model:        system.Model,
-		Status:       system.Status,
-		PowerState:   system.PowerState,
-		SerialNumber: system.SerialNumber,
-		SKU:          system.SKU,
-		IndicatorLED: string(system.IndicatorLED),
+		SystemUUID:             system.UUID,
+		Manufacturer:           system.Manufacturer,
+		Model:                  system.Model,
+		Status:                 system.Status,
+		PowerState:             system.PowerState,
+		SerialNumber:           system.SerialNumber,
+		SKU:                    system.SKU,
+		IndicatorLED:           string(system.IndicatorLED),
+		TotalSystemMemoryBytes: int64(system.MemorySummary.TotalSystemMemoryGiB) * 1024 * 1024 * 1024,
 	}, nil
 }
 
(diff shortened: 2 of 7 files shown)


bmc/bmc.go
```diff
@@ -188,16 +188,17 @@ type Processor struct {
...
-	SerialNumber      string
-	SKU               string
-	IndicatorLED      string
+	Manufacturer           string
+	Model                  string
+	Status                 common.Status
+	PowerState             redfish.PowerState
+	NetworkInterfaces      []NetworkInterface
+	Processors             []Processor
+	TotalSystemMemoryBytes int64
```
afritzler, line 197: Can we use `resource.Quantity` here and cast the correct value when setting it?

bmc/redfish.go
```diff
@@ -182,14 +182,15 @@ func (r *RedfishBMC) GetSystemInfo(systemUUID string) (SystemInfo, error) {
...
-		IndicatorLED: string(system.IndicatorLED),
+		SystemUUID:             system.UUID,
+		Manufacturer:           system.Manufacturer,
+		Model:                  system.Model,
+		Status:                 system.Status,
+		PowerState:             system.PowerState,
+		SerialNumber:           system.SerialNumber,
+		SKU:                    system.SKU,
+		IndicatorLED:           string(system.IndicatorLED),
+		TotalSystemMemoryBytes: int64(system.MemorySummary.TotalSystemMemoryGiB) * 1024 * 1024 * 1024,
```
afritzler, line 193: To avoid the multiplication duding assignment we could switch to a `resource.Quanity` type as described above.

afritzler: We could do something like that:
```
	memoryString := fmt.Sprintf("%dGi", memoryGiB)
	quantity, err := resource.ParseQuantity(memoryString)
	if err != nil {
		return resource.Quantity{}, fmt.Errorf("failed to parse memory quantity: %w", err)
	}
```

internal/controller/server_controller.go
```diff
@@ -436,6 +436,7 @@ func (r *ServerReconciler) updateServerStatus(ctx context.Context, log logr.Logg
 	server.Status.Manufacturer = systemInfo.Manufacturer
 	server.Status.Model = systemInfo.Model
 	server.Status.IndicatorLED = metalv1alpha1.IndicatorLED(systemInfo.IndicatorLED)
+	server.Status.TotalSystemMemory = resource.NewQuantity(systemInfo.TotalSystemMemoryBytes, resource.BinarySI)
```
afritzler, line 439: The assignment here becomes pretty straight forward with the change above.

afritzler (changes requested):

bmc/redfish.go
Nuckal777:
```go
quantity := resource.NewQuantity(memoryGiB, resource.BinarySI)
```
would save the format/parse roundtrip + error checking

stefanhipfel: but adds converting float32 to int64

Nuckal777: Sure, but right now we do `float32 -> string -> int64`, where [int64 conversion lives in `ParseQuantity`](https://github.com/kubernetes/apimachinery/blob/v0.31.2/pkg/api/resource/quantity.go#L277). I agree that the `memoryGiB` would need scaling, so `quantity := resource.NewQuantity(int64(memoryGiB)*1024*1024*1024, resource.BinarySI)`. Alternativly, we care about the precision there is also `ParseDecimalQuantity`.

stefanhipfel: you are looking at an outdated code :)

internal/controller/server_controller.go
```diff
@@ -436,6 +436,7 @@ func (r *ServerReconciler) updateServerStatus(ctx context.Context, log logr.Logg
 	server.Status.Manufacturer = systemInfo.Manufacturer
 	server.Status.Model = systemInfo.Model
 	server.Status.IndicatorLED = metalv1alpha1.IndicatorLED(systemInfo.IndicatorLED)
+	server.Status.TotalSystemMemory = &systemInfo.TotalSystemMemory
```
afritzler, line 439: We should use `ptr.To()` when assigning pointer values to fields.

afritzler (changes requested):
Some minor issue. Apart from that the PR looks good.

internal/controller/server_controller.go
Nuckal777: Why? `ptr.To()` is literally `return &val`. `ptr.To` is useful to get a pointer to a literal in an argument for a method call, for instance. Here it's just more keypresses. 🤔

afritzler (approved):

PR #174, opened 2024-11-08
Add wait polling for retrieving redfish storages and systems

stefanhipfel:
- adds wait polling for retrieving redfish storages. Some vendors redfish resources are not immediately available after a powerOn.
- moves waitForPower function to bmc client


bmc/redfish.go
```diff
@@ -342,14 +344,29 @@ func (r *RedfishBMC) checkBiosAttributes(attrs map[string]string) (reset bool, e
...
 	system, err := r.getSystemByUUID(systemUUID)
 	if err != nil {
 		return nil, err
 	}
-	systemStorage, err := system.Storage()
+	var systemStorage []*redfish.Storage
+	err = wait.PollUntilContextTimeout(
+		ctx,
+		10*time.Second,
+		5*time.Minute,
```
defo89, line 356: should we make interval and timeout configurable like in `bmcClient.WaitForServerPowerState`?

stefanhipfel: I was thinking about it. We just have so many config options already :)

bmc/redfish.go
```diff
@@ -36,6 +46,8 @@ func NewRedfishBMCClient(
 	ctx context.Context,
 	endpoint, username, password string,
 	basicAuth bool,
+
+	options PollingOptions,
```
Nuckal777, line 50: Superfluos newline. Given the amount of optiones, it may be worth introducing a `BMCClientOptions` struct to hold all of them.

bmc/redfish.go
```diff
@@ -342,14 +371,29 @@ func (r *RedfishBMC) checkBiosAttributes(attrs map[string]string) (reset bool, e
...
+	var systemStorage []*redfish.Storage
+	err = wait.PollUntilContextTimeout(
+		ctx,
+		r.options.ResourcePollingInterval,
+		r.options.ResourcePollingTimeout,
+		true,
+		func(ctx context.Context) (bool, error) {
+			if ctx.Err() != nil {
+				return false, ctx.Err()
+			}
```
Nuckal777, line 388: Not necessary I think, as `PollUntilContextTimeout` checks for context cancellation.

bmc/redfish.go
```diff
@@ -416,11 +460,24 @@ func (r *RedfishBMC) GetStorages(systemUUID string) ([]Storage, error) {
...
+	var systems []*redfish.ComputerSystem
+	err := wait.PollUntilContextTimeout(
+		ctx,
+		r.options.ResourcePollingInterval,
+		r.options.ResourcePollingTimeout,
+		true,
+		func(ctx context.Context) (bool, error) {
+			if ctx.Err() != nil {
+				return false, ctx.Err()
+			}
```
Nuckal777, line 474: Not necessary I think, as `PollUntilContextTimeout` checks for context cancellation.

bmc/redfish.go
stefanhipfel: I had that already implemented in the las commit (https://github.com/ironcore-dev/metal-operator/pull/174/commits/c7bf55023862dfad777d8fc1f49019b6e6d30b59). 
Some of the settings are mandatory like user/pw, so I wanted to separate them

cmd/manager/main.go
```diff
@@ -44,30 +45,35 @@ func init() {
...
+		serverResyncInterval    time.Duration
+		powerPollingInterval    time.Duration
+		powerPollingTimeout     time.Duration
+		resourcePollingInterval time.Duration
+		resourcePollingTimeout  time.Duration
+		discoveryTimeout        time.Duration
 	)
 
 	flag.DurationVar(&discoveryTimeout, "discovery-timeout", 30*time.Minute, "Timeout for discovery boot")
+	flag.DurationVar(&resourcePollingInterval, "resource-polling-interval", 5*time.Second,
```
afritzler, line 74: I would set a higher default for the polling interval. 10 or even 20s. We don't want to DDoS the Redfish endpoint.

afritzler (changes requested):

bmc/redfish.go
afritzler: Yes, you don't necessarily need the `ctx.Err()` check inside the polling function because `wait.PollUntilContextTimeout` itself already handles context cancellation and timeout. Including it may be redundant.

afritzler (changes requested):

afritzler (approved):

PR #178, opened 2024-11-12
Add kustomization for managing remote clusters

Nuckal777:
# Proposed Changes

- Add kustomization for configuring the metal-operator against a remote cluster. The kubeconfig is intended to by supplied by the user via an additional kustomize overlay.


afritzler:
Shouldn't we rather put the remote case into a separate subfolder instead of putting everything into `config/default`. Something like `config/virtual` or so.

Nuckal777:
We cannot even do it easily with kustomize, because the rbac needs to be deployed into the remote-cluster. 😢

PR #210, opened 2024-12-19
Remove BIOS settings

aobort:
# Proposed Changes

As was discussed and agreed we'll proceed with separate firmware/BIOS management controller. Hence there is no need to store BIOS settings in `Server` spec/status and server-controller should not modify BIOS settings

- Replace `BIOS` field by `BIOSSettingsRef` field which has type `LocalObjectReference` and should reference the `ServerBIOS` resource
- Remove `applyBiosSettings` function from server-controller


afritzler:
Closing in favor of https://github.com/ironcore-dev/metal-operator/pull/267

PR #219, opened 2025-01-16
Add `metal` user and SSH Keypair during `Discovery` boot

### LINKED ISSUE #217: Generate and inject admin SSH key during `Discovery` phase
# Summary

## Problem definition

If a `Server` which has been memory booted into the `Discovery` phase can not report the state we only have the possibility to look at the serial console for potential problems.

## Proposed solution

A possible solution would be to create an SSH keypair and store it in a secret in the namespace of the `metal-operator`. The SSH public key will be used in the `Discovery` ignition. Once a server passes the `Discovery` phase we remove the secret again. This feature will allow an operator to debug issues which might happen with the `metalprobe` on the `Server`.

/cc @damyan

afritzler:
# Proposed Changes

This PR introduces a feature to create a `metal` user with a corresponding SSH keypair which is stored as a `Secret` alongside the discovery `ServerBootConfiguration`. The `Secret` is named
`<serverbootconfig.name-ssh>`. This allows an administrator to use the private key in this `Secret` to SSH into the `Server` in case something went wrong when running the `metalprobe` agent.

The `Secret` is being removed again once the `ServerBootConfiguration` for the discovery boot has been removed after a successful `Server` onboarding.

Fixes #217


internal/ignition/default.go
```diff
@@ -54,11 +55,16 @@ systemd:
         WantedBy=multi-user.target
 storage:
   files: []
-passwd: {}
+passwd:
+  users:
+    - name: metal
+      password_hash: "$6$nq7WpLYJhmsUn7Pa$SDFlU/34qMoEnscBK03Fqc26/RaV01QYB2dMvIgoRTFmNNYQVg7YYzTOG4phKAgMw5UwRU4ePwKndYea9X/iO0"
```
Nuckal777, line 61: Is a password required?

internal/controller/server_controller.go
```diff
@@ -516,10 +546,35 @@ func (r *ServerReconciler) applyDefaultIgnitionForServer(ctx context.Context, lo
...
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to generate private key: %w", err)
+	}
+
+	privateKeyBytes := x509.MarshalPKCS1PrivateKey(privateKey)
+	privateKeyPem := pem.EncodeToMemory(&pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: privateKeyBytes,
+	})
```
Nuckal777, line 559:
```go
privateKeyBlock, err := ssh.MarshalPrivateKey(privateKey, "")
if err != nil {
	return nil, nil, err
}
privateKeyPem := pem.EncodeToMemory(privateKeyBlock)
```
maybe?

internal/controller/server_controller.go
```diff
@@ -516,10 +546,35 @@ func (r *ServerReconciler) applyDefaultIgnitionForServer(ctx context.Context, lo
...
+
+	publicKey := &privateKey.PublicKey
+	publicKeyBytes, err := x509.MarshalPKIXPublicKey(publicKey)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to marshal public key: %w", err)
+	}
+	publicKeyPem := pem.EncodeToMemory(&pem.Block{
+		Type:  "PUBLIC KEY",
+		Bytes: publicKeyBytes,
+	})
```
Nuckal777, line 569: Does `ssh.MarshalAuthorizedKey(publicKey)` achieve the same?

internal/controller/server_controller.go
```diff
@@ -516,10 +546,35 @@ func (r *ServerReconciler) applyDefaultIgnitionForServer(ctx context.Context, lo
 	return nil
 }
 
-func (r *ServerReconciler) generateDefaultIgnitionDataForServer(flags string) ([]byte, error) {
-	ignitionData, err := ignition.GenerateDefaultIgnitionData(ignition.ContainerConfig{
-		Image: r.ProbeImage,
-		Flags: flags,
+func generateSSHKeyPairs() ([]byte, []byte, error) {
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
```
Nuckal777, line 550: Could we consider `ed25519` or 4096-bit keys?

Nuckal777 (changes requested):
Thanks. Should the secret be deleted when transitioning away from the `Discovery` phase?

internal/ignition/default.go
afritzler: Unfortunatelly yes. But I am not also generating a unique bycrypted passwd hash for each `Server`.

internal/controller/server_controller.go
afritzler: Good point. Addressed.

internal/controller/server_controller.go
afritzler: Already fixed with the last commit.

Nuckal777 (approved):

hardikdr (approved):
Nice feature, lgtm.

PR #222, opened 2025-01-23
Apply `BMC` labels to `Server` on creation

defo89:
If `BMC` is created by outside tooling and common labels are applied, it would be good if `Server` inherits them.

### DIFF
--- a/internal/controller/bmc_controller.go
+++ b/internal/controller/bmc_controller.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/ironcore-dev/controller-utils/clientutils"
+	"github.com/ironcore-dev/controller-utils/metautils"
 	metalv1alpha1 "github.com/ironcore-dev/metal-operator/api/v1alpha1"
 	"github.com/ironcore-dev/metal-operator/bmc"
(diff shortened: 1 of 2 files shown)


afritzler (approved):

internal/controller/bmc_controller.go
```diff
@@ -162,6 +162,7 @@ func (r *BMCReconciler) discoverServers(ctx context.Context, log logr.Logger, bm
 	for i, s := range servers {
 		server := &metalv1alpha1.Server{}
 		server.Name = bmcutils.GetServerNameFromBMCandIndex(i, bmcObj)
+		server.Labels = bmcObj.Labels
```
hardikdr, line 165: Should we append them instead?

afritzler, line 165: Also shouldn't we append the labels in the mutate function e.g. if the `Server` has already labels?

afritzler (changes requested):

internal/controller/bmc_controller.go
defo89: good point, let me change it

defo89:
@hardikdr @afritzler pls have another look. I changed it and added a test - but maybe there is a better way here.

afritzler (approved):

PR #261, opened 2025-03-11
Add per `BMC` protocol scheme configuration in macdb

afritzler:
# Proposed Changes

Allow the administrator to set in the macdb the protocol for a `BMC` individually. If set, it will have precedence over the globally configured `--insecure` operator flag.

### DIFF
--- a/api/v1alpha1/bmc_types.go
+++ b/api/v1alpha1/bmc_types.go
@@ -9,10 +9,15 @@ import (
 )
 
 const (
-	BMCType              = "bmc"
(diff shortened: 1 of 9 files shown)


Nuckal777:
Why is `ProtocolScheme` defined twice in different packages?

afritzler:
> Why is `ProtocolScheme` defined twice in different packages?

Well one is in the `macdb` package and the other one is the `api` type. I could simplify it by using the `api` version everywhere.

PR #208, opened 2024-12-17
Add `ServerMaintenance` CRD

afritzler:
# Proposed Changes

-
-
-

Fixes #

PR #262, opened 2025-03-12
Add `MaintenanceBootConfigurationRef` to `Server` type

afritzler:
# Proposed Changes

The `MaintenanceBootConfigurationRef` is used to boot a `Server` into a maintenance mode. This field is superseeding the `BootConfigurationRef` which is used for a regular/discovery boot. The reason to introduce a new field here is, that in case of a `ServerClaim` claiming a `Server` the `ServerBootConfigurationRef` field is set with a reference to a `ServerBootConfiguration` which is generated out of a `ServerClaim`.

/ref https://github.com/ironcore-dev/metal-operator/pull/208

/cc @stefanhipfel

### DIFF
--- a/api/v1alpha1/server_types.go
+++ b/api/v1alpha1/server_types.go
@@ -106,6 +106,11 @@ type ServerSpec struct {
 	// if no boot configuration is specified.
 	BootConfigurationRef *v1.ObjectReference `json:"bootConfigurationRef,omitempty"`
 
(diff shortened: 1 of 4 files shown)


stefanhipfel:
i have added an extra bootconfigRef myself, since I realised the same issue :)

Nuckal777 (approved):

Nuckal777 (approved):

hardikdr (approved):

afritzler:
@stefanhipfel should we merge this and you rebase your changes or is it better to close this PR?

PR #265, opened 2025-03-15
Add `MaintenanceReconciler` implementation 

stefanhipfel:
# Proposed Changes

- Add maintenance reconciler implementation
- Add suite tests to ensure the correct behaviour


internal/controller/servermaintenance_controller_test.go
```diff
@@ -0,0 +1,207 @@
...
+			HaveField("Status.State", metalv1alpha1.ServerMaintenanceStatePending),
+		))
+
+		By("Checking the Server is not in maintenance")
+		Eventually(Object(server)).Should(SatisfyAll(
+			HaveField("Status.State", metalv1alpha1.ServerStateReserved),
+		))
+
+		By("Approving the maintenance")
+		Expect(k8sClient.Get(ctx, types.NamespacedName{
```
afritzler, line 147: I don't think we need to do a `Get` here. Instead tely on the `Update` helper to just update the annotation. If the object is not present it will throw an error.

internal/controller/suite_test.go
```diff
@@ -57,6 +57,11 @@ func TestControllers(t *testing.T) {
 }
 
 func DeleteAllMetalResources(ctx context.Context, namespace string) {
+	//var serverMaintenance metalv1alpha1.ServerMaintenance
```
afritzler, line 60: Can we remove the commented out code?

internal/controller/servermaintenance_controller.go
```diff
@@ -0,0 +1,484 @@
...
+	if config.Status.State == metalv1alpha1.ServerBootConfigurationStateReady {
+		log.V(1).Info("Server maintenance boot configuration is ready", "Server", server.Name)
+		if err := r.setAndPatchServerPowerState(ctx, server, serverMaintenance); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *ServerMaintenanceReconciler) createServerBootConfiguration(ctx context.Context, maintenance *metalv1alpha1.ServerMaintenance, server *metalv1alpha1.Server) (*metalv1alpha1.ServerBootConfiguration, error) {
```
afritzler, line 197: Maybe we should call this method `applyServerBootConfiguration` as we are actually doing a `PATCH` apply here.

internal/controller/servermaintenance_controller.go
```diff
@@ -0,0 +1,484 @@
...
+	log := ctrl.LoggerFrom(ctx)
```
afritzler, line 198: Can we pass the log from above to this method?

internal/controller/servermaintenance_controller.go
```diff
@@ -0,0 +1,484 @@
...
+		Kind:       "ServerBootConfiguration",
+	}
+	if err := r.Patch(ctx, server, client.MergeFrom(serverBase)); err != nil {
+		return config, fmt.Errorf("failed to patch server maintenance boot configuration ref: %w", err)
+	}
+	return config, nil
+}
+
+func (r *ServerMaintenanceReconciler) setAndPatchServerPowerState(ctx context.Context, server *metalv1alpha1.Server, maintenance *metalv1alpha1.ServerMaintenance) error {
+	log := ctrl.LoggerFrom(ctx)
```
afritzler, line 234: Can we pass the log from the outside to this method.

internal/controller/servermaintenance_controller.go
```diff
@@ -0,0 +1,484 @@
...
+	if err := r.Patch(ctx, server, client.MergeFrom(serverBase)); err != nil {
+		return fmt.Errorf("failed to patch server power state: %w", err)
+	}
+	log.V(1).Info("Patched server power state", "Server", server.Name, "Power", server.Spec.Power)
+
+	return nil
+}
+
+func (r *ServerMaintenanceReconciler) patchServerRef(ctx context.Context, maintenance *metalv1alpha1.ServerMaintenance, server *metalv1alpha1.Server) error {
+	log := ctrl.LoggerFrom(ctx)
```
afritzler, line 246: Same here.

internal/controller/servermaintenance_controller.go
```diff
@@ -0,0 +1,484 @@
...
+		if !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to get server: %w", err)
+		}
+		return nil, fmt.Errorf("server not found")
+	}
+	return server, nil
+}
+
+func (r *ServerMaintenanceReconciler) cleanup(ctx context.Context, server *metalv1alpha1.Server) error {
+	log := ctrl.LoggerFrom(ctx)
```
afritzler, line 314: Same here.

internal/controller/servermaintenance_controller.go
stefanhipfel: yes

afritzler (approved):

PR #267, opened 2025-03-20
Add `BIOSSettings` type and controller implementation

nagadeesh-nagaraja:
# Proposed Changes

- removes bios settings from server controller
- add crd BiosSettings with mapping 1:1 to server crd
- add controller (BiosSettings Reconciler) to handle updating bios settings
- the controller workflow is as follow:

1. bios setting update is handled in in handleInMaintenanceState func.
2.  this state is reached with or without bios version upgrade, ie as long as bios version matches)
3. get the current bios setting, check the diff with the required setting and apply it
4. if reboot is needed, creates a maintenance task.
5. waits for maintenance to be granted, and then reboot the system
6. verify the settings are applied before moving the state to completed.


internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,580 @@
...
+	if err := r.updateServerBIOSStatus(ctx, log, serverBIOS, metalv1alpha1.BIOSMaintenanceStateInSettingUpdate); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{Requeue: true}, nil
+}
+
+func (r *ServerBIOSReconciler) handleSettingUpdateState(ctx context.Context, log logr.Logger, serverBIOS *metalv1alpha1.ServerBIOS, server *metalv1alpha1.Server) (ctrl.Result, error) {
+	// check if we had requested a maintenance and the system is rebooted
+	if err := r.waitOnMaintenanceRequest(ctx, log, serverBIOS, server); err != nil {
```
stefanhipfel, line 281: shouldnt we make sure that a metalv1alpha1.ServerMaintenance{} has been created before we check on it?

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,580 @@
...
+	}
+
+	if waitForMaintenance {
+		serverMaintenance, err := r.getReferredServerMaintenance(ctx, log, serverBIOS.Spec.ServerMaintenanceRef)
+		if err != nil {
+			return fmt.Errorf("unable to get the referred server maintenance")
+		}
+
+		if serverMaintenance.Status.State == metalv1alpha1.ServerMaintenanceStateInMaintenance {
+			if server.Spec.ServerMaintenanceRef == nil {
```
stefanhipfel, line 392: that case should not be possible. Your metalv1alpha1.ServerMaintenance{} object can only be in ServerMaintenanceStateInMaintenance, if it is attached to the ServerMaintenanceRef

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,580 @@
...
+				// server in maintenance for other tasks.
+				log.V(1).Info("Server does not have maintenance reference", "Server", server.Name, "Maintenance", server.Spec.ServerMaintenanceRef)
+				return fmt.Errorf("server missing maintenance reference")
+			} else if server.Spec.ServerMaintenanceRef.UID != serverBIOS.Spec.ServerMaintenanceRef.UID {
+				// server in maintenance for other tasks.
+				log.V(1).Info("Server is already in maintenance for other tasks", "Server", server.Name, "Maintenance", server.Spec.ServerMaintenanceRef.Name)
+				return fmt.Errorf("server is already in maintenance by other tasks")
+			} else {
+				if server.Spec.Power == metalv1alpha1.PowerOn {
```
stefanhipfel, line 401: Power Changes via ServerMaintenance will happen when its status is set to Completed

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: Yes, I agree. However, i do think having this check will make sure that case is not ignored and anyway does not case any harm having it?

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: the reboot is handled in the controller now.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: handled in `waitOnMaintenanceRequest `. name has been updated to reflect the correct behavior.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,580 @@
...
+		err := r.requestMaintenanceOnServer(ctx, log, serverBIOS, server)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// wait for maintenance request to be granted
+	if err := r.waitOnMaintenanceRequest(ctx, log, serverBIOS, server); err != nil {
+		log.V(1).Info("Waiting for maintenance to be granted before continuing with version upgrade", "reason", err)
+		return ctrl.Result{Requeue: true}, nil
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: fixed

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: ah.. i got it completely wrong. looks like in maintenance, we do not change the power state at all. May be I will add that check.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: Actually, based on the serverMaintenance Controller, this is a possible case. 
the serverMaintenanceController first moves the SM CR to InMaintenance, and then puts the reference of self to the server CR and then moves the server into maintenance. between this time, it can enter this state.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,827 @@
...
+	log.V(1).Info("Reconciling serverBIOS")
+
+	return r.reconciliationRequired(ctx, log, serverBIOS)
+}
+
+// Determine whether reconciliation is required. It's not required if:
+// - object is being deleted;
+// - object does not contain reference to server;
+// - object contains reference to server, but server references to another object with lower version;
+func (r *ServerBIOSReconciler) reconciliationRequired(
```
stefanhipfel, line 66: Maybe we should not call it required, since in many cases within this function reconcile is actually not required

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,827 @@
...
+			serverMaintenance = &serverMain
+		}
+	}
+
+	if createNew {
+		serverMaintenance = &metalv1alpha1.ServerMaintenance{}
+		serverMaintenance.Name = serverBIOS.Name + "-" + time.Now().Format("20060102150405")
+		serverMaintenance.Namespace = r.ManagerNamespace
+		opResult, err := controllerutil.CreateOrPatch(ctx, r.Client, serverMaintenance, func() error {
+			serverMaintenance.SetLabels(map[string]string{"creator": serverBIOS.Name})
```
stefanhipfel, line 568: Maybe we should use something like`firmware.ironcore.dev` as label key

internal/controller/serverbios_controller.go
stefanhipfel: just check for server state inMaintenance

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: ya, so this is something I wanted to discuss. actual intent here is to track the owner/creator of this resource, to be able to list it later. 
what is the right way to do it?  is it using the key "firmware.ironcore.dev/creator"?

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: renamed it to reconciliationIfRequired

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: done.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: updated the server_controller.go( handleMaintenanceState & reconcile

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,822 @@
...
+	log.V(1).Info("Reconciling serverBIOS")
+
+	return r.reconcileIfRequired(ctx, log, serverBIOS)
+}
+
+// Determine whether reconciliation is required. It's not required if:
+// - object is being deleted;
+// - object does not contain reference to server;
+// - object contains reference to server, but server references to another object with lower version;
+func (r *ServerBIOSReconciler) reconcileIfRequired(
```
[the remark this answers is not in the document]
stefanhipfel: the func is called required, even though in many cases reconcile is actually not required according to the func code :)

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,747 @@
...
+			// server maintenance ref is wrong in either server or serverBIOS
+			// wait for update on the server obj
+			log.V(1).Info("Server is already in maintenance for other tasks", "Server", server.Name, "serverMaintenanceRef", server.Spec.ServerMaintenanceRef)
+			return fmt.Errorf("server maintenance for other tasks %v", server.Name)
+		}
+	} else {
+		// we still need to wait for server to enter maintenance
+		// wait for update on the server obj
+		log.V(1).Info("Server not yet in maintenance", "Server", server.Name, "State", server.Status.State, "MaintenanceRef", server.Spec.ServerMaintenanceRef)
+		return fmt.Errorf("server is not yet in maintenance %v", server.Name)
```
stefanhipfel, line 528: maybe we should return a bool if maintenance is granted, since other states are not really an "error"

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,747 @@
...
+	}
+
+	return false, diff, nil
+}
+
+func (r *ServerBIOSReconciler) waitOnRequiredPowerStatus(server *metalv1alpha1.Server, powerState metalv1alpha1.ServerPowerState) bool {
+	return server.Status.PowerState == powerState
+}
+
+func (r *ServerBIOSReconciler) waitIfMaintenanceRequested(log logr.Logger, serverBIOS *metalv1alpha1.ServerBIOS, server *metalv1alpha1.Server) error {
```
stefanhipfel, line 510: maybe we should not use wait in the func name, since the function is not really waiting (blocking) for the maintenance to be granted.
maybe more like checkMaintenanceGranted?

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,747 @@
...
+
+	// todo: handle version check to detect and upgrade only higher version
+	if currentBiosVersion != serverBIOS.Spec.BIOS.Version {
+		return true, diff, nil
+	}
+
+	return false, diff, nil
+}
+
+func (r *ServerBIOSReconciler) waitOnRequiredPowerStatus(server *metalv1alpha1.Server, powerState metalv1alpha1.ServerPowerState) bool {
```
stefanhipfel, line 506: same as 510

internal/controller/serverbios_controller.go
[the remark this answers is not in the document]
nagadeesh-nagaraja: changed

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,747 @@
...
+	}
+
+	return serverMaintenance, nil
+}
+
+func (r *ServerBIOSReconciler) getReferredserverBIOS(
+	ctx context.Context,
+	log logr.Logger,
+	referredBIOSSetteingRef *corev1.LocalObjectReference,
+) (metalv1alpha1.ServerBIOS, error) {
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: done

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: agreed.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: done

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: done

nagadeesh-nagaraja:
fixed few more missed review comments

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+	server, err := r.getReferredServer(ctx, log, serverBIOS.Spec.ServerRef)
+	if err != nil {
+		log.V(1).Info("referred server object could not be fetched")
+		return ctrl.Result{}, err
+	}
+	// patch server with serverbios reference
+	if server.Spec.BIOSSettingsRef == nil {
+		if err := r.patchServerBIOSRefOnServer(ctx, log, server, &corev1.LocalObjectReference{Name: serverBIOS.Name}); err != nil {
+			return ctrl.Result{}, err
+		}
```
Nuckal777, line 97: Could we return early here with `requeue: true`? This would simplify a logic a bit, because the `else if` statement could be turned into a plain `if`.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+		// check if the current BIOS setting version is newer and update reference if it is newer
+		// todo : handle version checks correctly
+		if referredBIOSSetting.Spec.BIOS.Version < serverBIOS.Spec.BIOS.Version {
+			log.V(1).Info("Updating BIOSSetting reference to the latest BIOS version")
+			if err := r.patchServerBIOSRefOnServer(ctx, log, server, &corev1.LocalObjectReference{Name: serverBIOS.Name}); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
```
Nuckal777, line 111: Here we could also return early with `requeue: true`, I think.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+func (r *ServerBIOSReconciler) cleanupReferences(
+	ctx context.Context,
+	log logr.Logger,
+	serverBIOS *metalv1alpha1.ServerBIOS,
+) (err error) {
+	if err := r.cleanupServerMaintenanceReferences(ctx, log, serverBIOS); err != nil {
+		return err
+	}
+
+	if serverBIOS.Spec.ServerRef != nil {
```
Nuckal777, line 189: Please invert the condition and return early. This avoids one level of nesting.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+
+	return r.applySettingUpdateStateTransition(ctx, log, serverBIOS, server, &settingsDiff)
+}
+
+func (r *ServerBIOSReconciler) checkAndRequestMaintenance(
+	ctx context.Context,
+	log logr.Logger,
+	serverBIOS *metalv1alpha1.ServerBIOS,
+	server *metalv1alpha1.Server,
+	settingsDiff *map[string]string,
```
Nuckal777, line 347: Unnecesscary pointer.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+) (bool, error) {
+	// check if we need to request maintenance if we dont have it already
+	// note: having this check will reduce the call made to BMC.
+	if serverBIOS.Spec.ServerMaintenanceRef == nil {
```
Nuckal777, line 351: Invert condition and return early.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+	}
+	return false, nil
+}
+
+func (r *ServerBIOSReconciler) applySettingUpdateStateTransition(
+	ctx context.Context,
+	log logr.Logger,
+	serverBIOS *metalv1alpha1.ServerBIOS,
+	server *metalv1alpha1.Server,
+	settingsDiff *map[string]string,
```
Nuckal777, line 376: Unnecessary pointer.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+
+	if serverBIOS.Spec.ServerMaintenanceRef != nil {
+		serverMaintenance, err := r.getReferredServerMaintenance(ctx, log, server.Spec.ServerMaintenanceRef)
+		if err != nil {
+			log.V(1).Error(err, "unable to fetch serverMaintenance")
+			return ctrl.Result{}, err
+		}
+
+		serverMaintenanceBase := serverMaintenance.DeepCopy()
+		serverMaintenance.Status.State = metalv1alpha1.ServerMaintenanceStateCompleted
```
Nuckal777, line 493: Should the maintenance be set to the `Failed` state?

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+	log.V(1).Info("Failed to update bios setting", "ctx", ctx, "serverBIOS", serverBIOS, "server", server)
+	return ctrl.Result{}, nil
+}
+
+func (r *ServerBIOSReconciler) getBiosSettingDifference(
+	ctx context.Context,
+	log logr.Logger,
+	serverBIOS *metalv1alpha1.ServerBIOS,
+	server *metalv1alpha1.Server,
+) (updatedNeeded bool, diff map[string]string, err error) {
```
Nuckal777, line 520: I move out the check for the divergence of the bios version. It's only used in one place and a single line, currently.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+	bmcClient, err := bmcutils.GetBMCClientForServer(ctx, r.Client, server, r.Insecure, r.BMCOptions)
+	if err != nil {
+		return false, diff, fmt.Errorf("failed to create BMC client: %w", err)
+	}
+	defer bmcClient.Logout()
+
+	keys := make([]string, 0, len(serverBIOS.Spec.BIOS.Settings))
+	for k := range serverBIOS.Spec.BIOS.Settings {
+		keys = append(keys, k)
+	}
```
Nuckal777, line 531: Please use `maps.Keys()`.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+
+	return false, diff, nil
+}
+
+func (r *ServerBIOSReconciler) checkForRequiredPowerStatus(
+	server *metalv1alpha1.Server,
+	powerState metalv1alpha1.ServerPowerState,
+) bool {
+	return server.Status.PowerState == powerState
+}
```
Nuckal777, line 570: This function seems superfluous. The check can simply be done at the call site.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+	}
+
+	if server.Status.State == metalv1alpha1.ServerStateMaintenance {
+		if server.Spec.ServerMaintenanceRef == nil || server.Spec.ServerMaintenanceRef.UID != serverBIOS.Spec.ServerMaintenanceRef.UID {
+			// server in maintenance for other tasks. or
+			// server maintenance ref is wrong in either server or serverBIOS
+			// wait for update on the server obj
+			log.V(1).Info("Server is already in maintenance for other tasks", "Server", server.Name, "serverMaintenanceRef", server.Spec.ServerMaintenanceRef)
+			return false
+		}
```
Nuckal777, line 589: Returning early would avoid the `else` block.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+	serverBIOSReference *corev1.LocalObjectReference,
+) error {
+	if server.Spec.BIOSSettingsRef == serverBIOSReference {
+		return nil
+	}
+
+	var err error
+	serverBase := server.DeepCopy()
+	server.Spec.BIOSSettingsRef = serverBIOSReference
+	if err = r.Patch(ctx, server, client.MergeFrom(serverBase)); err != nil {
```
Nuckal777, line 697: The `var err error` could be removed by defining `err` as part of the `if err := ...` statement.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,838 @@
...
+	serverBIOS *metalv1alpha1.ServerBIOS,
+	serverMaintenance *metalv1alpha1.ServerMaintenance,
+) error {
+	serverBIOSBase := serverBIOS.DeepCopy()
+
+	if serverMaintenance == nil {
+		serverBIOS.Spec.ServerMaintenanceRef = nil
+	} else {
+		serverBIOS.Spec.ServerMaintenanceRef = &corev1.ObjectReference{
+			APIVersion: "metal.ironcore.dev/v1alpha1",
```
Nuckal777, line 716: `serverMaintenance.GroupVersionKind().GroupVersion().String()` should avoid hard-coding the apiVersion.

Nuckal777 (commented):
Thanks for looking into this. 👍 The logic in general seems plausible to me. I would like to suggest some small mostly stylistic improvements, though.

api/v1alpha1/serverbios_types.go
```diff
@@ -0,0 +1,103 @@
...
+
+	// ServerRef is a reference to a specific server to be claimed.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="serverRef is immutable"
+	ServerRef *corev1.LocalObjectReference `json:"serverRef,omitempty"`
+
+	// ServerMaintenancePolicy is maintenance policy to be enforced on the server.
+	ServerMaintenancePolicy ServerMaintenancePolicy `json:"ServerMaintenancePolicyTemplate,omitempty"`
+
+	// ServerMaintenanceRef is a reference to a ServerMaintenance object that that BIOS has requested for the referred server.
+	ServerMaintenanceRef *corev1.ObjectReference `json:"serverMaintenanceRef,omitempty"`
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: the corresponding CRD is called "serverMaintenance", Hence keeping the naming convention?

api/v1alpha1/serverbios_types.go
```diff
@@ -0,0 +1,103 @@
...
+
+const (
+	// SettingUpdateStateWaitOnServerReboot specifies that the bios setting state is waiting on server to turn off during Reboot.
+	BIOSSettingUpdateWaitOnServerRebootPowerOff BIOSSettingUpdateState = "WaitOnServerRebootPowerOff"
+	// BIOSSettingUpdateWaitOnServerRebootPowerOn specifies that the bios setting state is waiting on server to turn on during Reboot.
+	BIOSSettingUpdateWaitOnServerRebootPowerOn BIOSSettingUpdateState = "WaitOnServerRebootPowerOn"
+	// SettingUpdateStateIssued specifies that the bios new setting was posted to RedFish
+	BIOSSettingUpdateStateIssue BIOSSettingUpdateState = "IssueSettingUpdate"
+	// SettingUpdateStateCompleted specifies that the bios setting has been completed.
+	BIOSSettingUpdateStateVerification BIOSSettingUpdateState = "VerifySettingUpdate"
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: These actually corresponds to the current state (task). 
 state **IssueSettingUpdate** -> issues the setting update. ** VerifySettingUpdate** verifies the settings. 
 
 are these convention confusing or different from the norm?

bmc/bmc.go
```diff
@@ -67,6 +69,11 @@ type Bios struct {
 	Attributes map[string]string
 }
 
+type AllowedValues struct {
+	ValueDisplayName string
+	ValueName        string
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: these naming conventions are from the redfish API.  I would keep it as is.  

if not, I would have to add the json tag. let me know if your thought.

bmc/redfish.go
```diff
@@ -262,7 +265,7 @@ func (r *RedfishBMC) GetBiosAttributeValues(
 	systemUUID string,
 	attributes []string,
 ) (
-	result map[string]string,
+	result map[string]interface{},
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: I changed it to 
redfish.SettingAttributes --> as these are derived from that anyway.

bmc/redfish.go
```diff
@@ -280,42 +283,110 @@ func (r *RedfishBMC) GetBiosAttributeValues(
 	if err != nil {
 		return
 	}
-	result = make(map[string]string, len(attributes))
+	result = make(map[string]interface{}, len(attributes))
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: same as above.

bmc/redfish.go
```diff
@@ -280,42 +283,110 @@ func (r *RedfishBMC) GetBiosAttributeValues(
...
 ) {
-	reset = false
 	system, err := r.getSystemByUUID(ctx, systemUUID)
 	if err != nil {
-		return
+		return result, err
 	}
-	bios, err := system.Bios()
+
+	type tempSys redfish.ComputerSystem
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: fixed the unwanted types.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: I do not see much difference between the two.  let me know if there is?
I would keep it as is if no objection.

nagadeesh-nagaraja: Actually, I could just return as server is also watched by this controller and should requeue automatically anyway.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: dont think I understand this?

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: I feel its easier to read with the function :D I would prefer to leave it as is.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: I feel its easier to read with the function :D I would prefer to leave it as is.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: Plan was to use it for version upgrade which comes up next..

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: fixed.

nagadeesh-nagaraja: I still think requeueing is worse than having if else...let me know your thoughts...
I can revert it.

api/v1alpha1/serverbios_types.go
```diff
@@ -0,0 +1,103 @@
...
+	// Settings contains BIOS settings as map
+	// +optional
+	Settings map[string]string `json:"settings,omitempty"`
+}
+
+// BIOSMaintenanceState specifies the current state of the BIOS maintenance.
+type BIOSMaintenanceState string
+
+const (
+	// BIOSMaintenanceStateInVersionUpgrade specifies that the server bios is in version upgrade path.
```
afritzler, line 47: I think we should think about the state transitions and especially the naming here. I would assume there should only be things like `Pending`, `InProgress`, `Failed` and `Applied`. It don't think it is a good idea to do a fine grained state transitioning here. For that I would rather use transition conditions.

afritzler: /ref https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

api/v1alpha1/serverbios_types.go
```diff
@@ -0,0 +1,103 @@
...
+	BIOSMaintenanceStateSynced BIOSMaintenanceState = "SyncSettingsCompleted"
+	// BIOSMaintenanceStateFailed specifies that the server maintenance has failed.
+	BIOSMaintenanceStateFailed BIOSMaintenanceState = "Failed"
+)
+
+type BIOSSettingUpdateState string
+
+const (
+	// SettingUpdateStateWaitOnServerReboot specifies that the bios setting state is waiting on server to turn off during Reboot.
+	BIOSSettingUpdateWaitOnServerRebootPowerOff BIOSSettingUpdateState = "WaitOnServerRebootPowerOff"
```
afritzler, line 61: Same applies here as with the state transitions above.

api/v1alpha1/serverbios_types.go
```diff
@@ -0,0 +1,103 @@
...
+	BIOSSettingUpdateStateVerification BIOSSettingUpdateState = "VerifySettingUpdate"
+)
+
+// ServerBIOSStatus defines the observed state of ServerBIOS.
+type ServerBIOSStatus struct {
+
+	// State represents the current state of the bios configuration task.
+	State BIOSMaintenanceState `json:"state,omitempty"`
+	// UpdateSettingState represents the current state of the bios setting update task.
+	UpdateSettingState BIOSSettingUpdateState `json:"updateSettingState,omitempty"`
```
afritzler, line 76: I would add here a condition field slice in the status.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,849 @@
...
+}
+
+const serverBIOSFinalizer = "firmware.ironcore.dev/serverbios"
+
+const serverBIOSCreatorLabel = "firmware.ironcore.dev/created-by"
+
+// +kubebuilder:rbac:groups=metal.ironcore.dev,resources=serverbioses,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=metal.ironcore.dev,resources=serverbioses/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=metal.ironcore.dev,resources=serverbioses/finalizers,verbs=update
+//+kubebuilder:rbac:groups=metal.ironcore.dev,resources=servers,verbs=get;list;watch;update
```
afritzler, line 46: Can we use the same indentation here.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,849 @@
...
+//+kubebuilder:rbac:groups=metal.ironcore.dev,resources=servermaintenances,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=metal.ironcore.dev,resources=servermaintenances/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="batch",resources=jobs,verbs=get;list;watch;create;update;patch;delete
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+func (r *ServerBIOSReconciler) Reconcile(
```
afritzler, line 54: Cosmetic: For shorter method heads: can we put them into one line as this is true for all the methods in the project.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,849 @@
...
+	log.V(1).Info("Reconciling serverBIOS")
+
+	return r.reconcileExists(ctx, log, serverBIOS)
+}
+
+// Determine whether reconciliation is required. It's not required if:
+// - object is being deleted;
+// - object does not contain reference to server;
+// - object contains reference to server, but server references to another object with lower version;
+func (r *ServerBIOSReconciler) reconcileExists(
```
afritzler, line 72: Can we stick with the same definition of `reconcileExists` like we did here: https://github.com/ironcore-dev/metal-operator/blob/main/internal/controller/endpoint_controller.go#L54 That way every reconciler boilerplate part looks the same and it is easy for ppl to reverse engineer what is going on.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,849 @@
...
+		log.V(1).Info("referred server object could not be fetched")
+		return ctrl.Result{}, err
+	}
+	// patch server with serverbios reference
+	if server.Spec.BIOSSettingsRef == nil {
+		if err := r.patchServerBIOSRefOnServer(ctx, log, server, &corev1.LocalObjectReference{Name: serverBIOS.Name}); err != nil {
+			return ctrl.Result{}, err
+		}
+		// because we requeue server only after serverMaintenance is created. we need to manually requeue here.
+		return ctrl.Result{Requeue: true}, nil
```
afritzler, line 101: I would be careful with `Requeue: true`. The proper way to do that is to have in `SetupWithManager` a correct `Watch` setup for all dependant resources.

bmc/bmc.go
```diff
@@ -82,6 +89,7 @@ type RegistryEntryAttributes struct {
 	ResetRequired bool
 	Type          string
 	WriteOnly     bool
+	Value         []AllowedValues
```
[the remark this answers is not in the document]
Nuckal777: I would propose to rename the field from `Value` to `AllowedValues`.

internal/controller/serverbios_controller.go
```diff
@@ -0,0 +1,849 @@
...
+	serverBIOS *metalv1alpha1.ServerBIOS,
+	server *metalv1alpha1.Server,
+) (bool, error) {
+
+	// if Server maintenance ref is already given. no further action required.
+	if serverBIOS.Spec.ServerMaintenanceRef != nil {
+		return false, nil
+	}
+
+	serverMaintenance := &metalv1alpha1.ServerMaintenance{}
```
afritzler, line 624: Can we do the name and namespace assignment in the struct definition above?

internal/controller/serverbios_controller_test.go
```diff
@@ -0,0 +1,267 @@
...
+		Eventually(Object(server)).Should(SatisfyAll(
+			HaveField("Status.State", metalv1alpha1.ServerStateInitial),
+		))
+
+		By("update the server state to Available  state")
+		Eventually(UpdateStatus(server, func() {
+			server.Status.State = metalv1alpha1.ServerStateAvailable
+		})).Should(Succeed())
+
+		By("Ensuring that the server has the correct state")
```
afritzler, line 80: No need to check for that as we assume that a k8s status patch is working.

internal/controller/serverbios_controller_test.go
```diff
@@ -0,0 +1,267 @@
...
+				BIOS:                    metalv1alpha1.BIOSSettings{Version: "P79 v1.45 (12/06/2017)", Settings: BIOSSetting},
+				ServerRef:               &v1.LocalObjectReference{Name: server.Name},
+				ServerMaintenancePolicy: metalv1alpha1.ServerMaintenancePolicyEnforced,
+			},
+		}
+		Expect(k8sClient.Create(ctx, serverBIOS)).To(Succeed())
+
+		By("Ensuring that the Server has the correct server bios setting ref")
+		Eventually(Object(server)).Should(SatisfyAll(
+			HaveField("Spec.BIOSSettingsRef", Not(BeNil())),
```
afritzler, line 101: why not directly do a localobjectref compare here?

internal/controller/serverbios_controller_test.go
```diff
@@ -0,0 +1,267 @@
...
+				GenerateName: "test-",
+			},
+			Spec: metalv1alpha1.ServerBIOSSpec{
+				BIOS:                    metalv1alpha1.BIOSSettings{Version: "P79 v1.45 (12/06/2017)", Settings: BIOSSetting},
+				ServerRef:               &v1.LocalObjectReference{Name: server.Name},
+				ServerMaintenancePolicy: metalv1alpha1.ServerMaintenancePolicyEnforced,
+			},
+		}
+		Expect(k8sClient.Create(ctx, serverBIOS)).To(Succeed())
+
```
afritzler, line 98: You might want to check here that the `ServerMaintenance` object has been created correctly.

internal/controller/serverbios_controller_test.go
```diff
@@ -0,0 +1,267 @@
...
+		By("Ensuring that the Server has the correct server bios setting ref")
+		Eventually(Object(server)).Should(SatisfyAll(
+			HaveField("Spec.BIOSSettingsRef", Not(BeNil())),
+			HaveField("Spec.BIOSSettingsRef.Name", serverBIOS.Name),
+		))
+
+		Eventually(Object(serverBIOS)).Should(SatisfyAny(
+			HaveField("Status.State", metalv1alpha1.BIOSMaintenanceStateSynced),
+		))
```
afritzler, line 107: How about testing here also the deletion flow?

internal/controller/serverbios_controller_test.go
```diff
@@ -0,0 +1,267 @@
...
+			HaveField("Status.State", metalv1alpha1.ServerStateInitial),
+		))
+
+		By("update the server state to Available  state")
+		Eventually(UpdateStatus(server, func() {
+			server.Status.State = metalv1alpha1.ServerStateAvailable
+		})).Should(Succeed())
+
+		By("Ensuring that the server has the correct state")
+		Eventually(Object(server)).Should(SatisfyAll(
```
afritzler, line 124: Same as above. Can be removed.

internal/controller/serverbios_controller_test.go
```diff
@@ -0,0 +1,267 @@
...
+				BIOS:                    metalv1alpha1.BIOSSettings{Version: "P79 v1.45 (12/06/2017)", Settings: BIOSSetting},
+				ServerRef:               &v1.LocalObjectReference{Name: server.Name},
+				ServerMaintenancePolicy: metalv1alpha1.ServerMaintenancePolicyEnforced,
+			},
+		}
+		Expect(k8sClient.Create(ctx, serverBIOS)).To(Succeed())
+
+		By("Ensuring that the Server has the bios setting ref")
+		Eventually(Object(server)).Should(SatisfyAll(
+			HaveField("Spec.BIOSSettingsRef", Not(BeNil())),
```
afritzler, line 144: `LocalObjectRef` compare here.

internal/controller/serverbios_controller_test.go
```diff
@@ -0,0 +1,267 @@
...
+			HaveField("Status.State", metalv1alpha1.BIOSMaintenanceStateSynced),
+		))
+
+		By("Deleting the ServerBIOS")
+		Expect(k8sClient.Delete(ctx, serverBIOS)).To(Succeed())
+
+		By("Ensuring that the Server BIOS ref is empty")
+		Eventually(Object(server)).Should(SatisfyAll(
+			HaveField("Spec.BIOSSettingsRef", BeNil()),
+		))
```
afritzler, line 158: I would also ensure that the `ServerMaintenance` object is also gone.

internal/controller/serverbios_controller_test.go
```diff
@@ -0,0 +1,267 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: ns.Name,
+				Name:      serverBIOS.Spec.ServerMaintenanceRef.Name,
+			},
+		}
+		Eventually(Get(serverMaintenance)).Should(Succeed())
+
+		By("Ensuring that the server has accepted the Maintenance request")
+		Eventually(Object(server)).Should(SatisfyAll(
+			HaveField("Spec.ServerMaintenanceRef", Not(BeNil())),
```
afritzler, line 223: Direct ref comparsion here.

afritzler (changes requested):

afritzler:
A general question: we are calling the resource `ServerBIOS`. But this resource is only responsible to enforce a certain set of BIOS settings on a `Server` to be applied. Shouldn't this resource than rather be called `ServerBIOSSettings`? Or are we planning to also handling BIOS updates via this resource?

afritzler:
It would be great to capture the type + fields + flow in the docs already https://github.com/ironcore-dev/metal-operator/tree/main/docs/concepts That way we can maybe resolve conceptional issues faster than in the reconciler code.

nagadeesh-nagaraja:
> A general question: we are calling the resource `ServerBIOS`. But this resource is only responsible to enforce a certain set of BIOS settings on a `Server` to be applied. Shouldn't this resource than rather be called `ServerBIOSSettings`? Or are we planning to also handling BIOS updates via this resource?


@afritzler, Yes Idea is to handle BIOS version upgrade and Settings here. another option is to call it `biosMaintenance`?  let me know which sounds better.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: It was changed based on suggestion here: 
https://github.com/ironcore-dev/metal-operator/pull/267#discussion_r2035574965

server resource is watched and added to this controller queue. However, I do it after the serverMaintenance is created, to avoid too much unnecessary processing ( expect server resource is changed a lot). 

as this is only attaching the self ref to server, I prefer to not return here and continue. 
@Nuckal777  if you agree, I would remove this return after patching self ref to server resource.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: @afritzler I dont think I understood this. 
`func (r *EndpointReconciler) reconcileExists(ctx context.Context, log logr.Logger, endpoint *metalv1alpha1.Endpoint) (ctrl.Result, error) {
`

`func (r *ServerBIOSReconciler) reconcileExists(
	ctx context.Context,
	log logr.Logger,
	serverBIOS *metalv1alpha1.ServerBIOS,
) (ctrl.Result, error) {`

these look same to me? 

did you mean the definition of `func (r *ServerBIOSReconciler) reconcile(...` is keeping every controller methods same a requirement? 

reason to add extra parameter to to avoid extra call in subsequent methods to get server resource which I have already have.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: All the function parameters (for multiple line) where changed based on suggestion in https://github.com/ironcore-dev/metal-operator/pull/267#discussion_r2022999982

what is the convention, do we have any guidelines for formatting?

internal/controller/serverbios_controller_test.go
nagadeesh-nagaraja: deletions are tested in all the other 2 test cases. Hence, did not extra here. I do not think deletion needs to be tested multiple times in unit tests?

internal/controller/serverbios_controller.go
Nuckal777: I guess @afritzler proposes to still return, but without a explicit requeue and instead trigger a reconciliation whenever the `Server` objects change.

Nuckal777:
Regarding the "invert condition + early return" comments. I am huge fan of:
```go
if !precondition {
    return
}
// block
// of
// code
```
instead of
```go
if precondition {
    // block
    // of
    // code
}
```

nagadeesh-nagaraja:
> https://github.com/ironcore-dev/metal-operator/tree/main/docs/concepts
@Nuckal777 

Understood. I would argue that. 
```
if !precondition {
    return
}
// block
// of
// code
```
and 
```
if precondition {
    // block
    // of
    // code
}
return
```
are functionally same, just the coding style.  Correct me if I got that wrong


if not a huge objection. I would leave these two as is. 
https://github.com/ironcore-dev/metal-operator/pull/267#discussion_r2035592211
https://github.com/ironcore-dev/metal-operator/pull/267#discussion_r2035577882

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: have reverted based on https://github.com/ironcore-dev/metal-operator/pull/267#discussion_r2036620170

not returning gives us ability to not queue requests on servers until nessessary (after maintenance is created) and avoids extra cycles.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: same as above. please reopen if its absolutely required.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: closing based on https://github.com/ironcore-dev/metal-operator/pull/267#issuecomment-2791977586
please reopen if required.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: closing based on https://github.com/ironcore-dev/metal-operator/pull/267#issuecomment-2791977586
please reopen if required.

nagadeesh-nagaraja:
> It would be great to capture the type + fields + flow in the docs already https://github.com/ironcore-dev/metal-operator/tree/main/docs/concepts That way we can maybe resolve conceptional issues faster than in the reconciler code.

@afritzler Can I create a sub-issue and work on it in parallel? that way the PR will not be blocked?

internal/controller/serverbios_controller_test.go
nagadeesh-nagaraja: @afritzler this test does not create `ServerMaintenance` as the setting is not a mismatch. the `ServerMaintenance` is created only in the test "should move to through upgrade state if no bios version is not right".

bmc/bmc.go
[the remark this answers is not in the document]
nagadeesh-nagaraja: `Value` field is coming from the Redfish API
```json
 {
        "AttributeName": "IscsiDev1ConOrder",
        "CurrentValue": null,
        "DisplayName": "Connection Order",
        "DisplayOrder": 8904,
        "HelpText": "This field controls the order for which the iSCSI connections will be attempted.",
        "Hidden": false,
        "Immutable": false,
        "MenuPath": "./NetworkSettingsRef/IscsiDev1SettingsRef",
        "ReadOnly": false,
        "ResetRequired": true,
        "Type": "Enumeration",
        "Value": [
          {
            "ValueDisplayName": "Connection 1, Connection 2",
            "ValueName": "Con1Con2"
          },
          {
            "ValueDisplayName": "Connection 2, Connection 1",
            "ValueName": "Con2Con1"
          }
        ],
        "WarningText": null,
        "WriteOnly": false
      },
```

If you prefer to keep the that name. I could change the "`AllowedValues`" to "`Values`" if you prefer that?

api/v1alpha1/serverbios_types.go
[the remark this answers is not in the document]
nagadeesh-nagaraja: closing this, Please open if required.

api/v1alpha1/serverbios_types.go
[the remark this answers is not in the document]
nagadeesh-nagaraja: will need update based on: https://github.com/ironcore-dev/metal-operator/pull/267#discussion_r2036607559 
will close this and can be tracked at one place.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: please re open if required.

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: please reopen if required.

internal/controller/serverbios_controller_test.go
nagadeesh-nagaraja: closing as its done other tests

internal/controller/serverbios_controller_test.go
nagadeesh-nagaraja: done in other tests

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: will leave it as all other functions are multi line

api/v1alpha1/serverbios_types.go
nagadeesh-nagaraja: tracked in https://github.com/ironcore-dev/metal-operator/pull/267#discussion_r2036607559

api/v1alpha1/serverbios_types.go
nagadeesh-nagaraja: created issue, https://github.com/ironcore-dev/metal-operator/issues/295 will take up in subsequent PR

bmc/bmc.go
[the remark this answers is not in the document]
nagadeesh-nagaraja: keeping it as per redfish

internal/controller/serverbios_controller.go
nagadeesh-nagaraja: will leave it in pending state to be checked by the human user...

api/v1alpha1/biossettings_types.go
```diff
@@ -0,0 +1,101 @@
...
+	// +optional
+	SettingsMap map[string]string `json:"settings,omitempty"`
+}
+
+// BiosSettingsState specifies the current state of the BIOS maintenance.
+type BiosSettingsState string
+
+const (
+	// BiosSettingsStateInProgress specifies that the server bios is in setting update path.
+	BiosSettingsStateInProgress BiosSettingsState = "InProgress"
```
afritzler, line 48: Can we use here the following states: `Pending`, `InProgress`, `Applied`, `Successful` or `Ready`?

api/v1alpha1/biossettings_types.go
```diff
@@ -0,0 +1,101 @@
...
+	// State represents the current state of the bios configuration task.
+	State BiosSettingsState `json:"state,omitempty"`
+	// UpdateSettingState represents the current state of the bios setting update task.
+	UpdateSettingState BiosSettingUpdateState `json:"updateSettingState,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+
```
afritzler, line 80: Can we add here some sane printcolum fields e.g. like in the `Server` type:
```
//+kubebuilder:printcolumn:name="UUID",type=string,JSONPath=`.spec.uuid`
//+kubebuilder:printcolumn:name="Manufacturer",type=string,JSONPath=`.status.manufacturer`
//+kubebuilder:printcolumn:name="Model",type=string,JSONPath=`.status.model`
//+kubebuilder:printcolumn:name="SKU",type=string,JSONPath=`.status.sku`,priority=100
//+kubebuilder:printcolumn:name="SerialNumber",type=string,JSONPath=`.status.serialNumber`,priority=100
//+kubebuilder:printcolumn:name="PowerState",type=string,JSONPath=`.status.powerState`
//+kubebuilder:printcolumn:name="IndicatorLED",type=string,JSONPath=`.status.indicatorLED`,priority=100
//+kubebuilder:printcolumn:name="State",type=string,JSONPath=`.status.state`
//+kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
```

internal/controller/biossettings_controller.go
```diff
@@ -0,0 +1,776 @@
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to get referred serverMaintenance obj from biosSettings: %w", err)
+	}
+
+	// if we got the server ref. by and its not being deleted
+	if err == nil && serverMaintenance.DeletionTimestamp.IsZero() {
+		// if the serverMaintenace is not created by biosSettings Controller, dont delete.
+		labelsOnMaintenance := serverMaintenance.GetLabels()
+
+		// created by the controller
```
afritzler, line 128: We could leverage here `metav1.IsControlledBy()`. That way we don't need to set extra labels on the `Maintenance` object.

internal/controller/biossettings_controller.go
```diff
@@ -0,0 +1,776 @@
...
+func (r *BiosSettingsReconciler) handleFailedState(
+	ctx context.Context,
+	log logr.Logger,
+	biosSettings *metalv1alpha1.BiosSettings,
+	server *metalv1alpha1.Server,
+) (ctrl.Result, error) {
+	log.V(1).Info("Handle failed setting update with no maintenance reference")
+	// todo: revist this logic to either create maintenance if not present, put server in Error state on failed bios settings maintenance
+
+	// request maintenance if needed,
```
afritzler, line 449: Can we remove the commented out code here.

internal/controller/biossettings_controller.go
```diff
@@ -0,0 +1,776 @@
...
+	}
+
+	if biosSettings.Spec.ServerRef != nil {
+		server, err := r.getReferredServer(ctx, log, biosSettings.Spec.ServerRef)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return err
+		}
+		// if we can not find the server, nothing else to clean up
+		if apierrors.IsNotFound(err) {
+			return nil
```
afritzler, line 165: We could log here something like `Server is gone`.

internal/controller/biossettings_controller.go
```diff
@@ -0,0 +1,776 @@
...
+	}
+	return nil
+}
+
+func (r *BiosSettingsReconciler) cleanupReferences(
+	ctx context.Context,
+	log logr.Logger,
+	biosSettings *metalv1alpha1.BiosSettings,
+) (err error) {
+	if err := r.cleanupServerMaintenanceReferences(ctx, log, biosSettings); err != nil {
```
afritzler, line 154: Since we are setting ```
return controllerutil.SetControllerReference(biosSettings, serverMaintenance, r.Client.Scheme())
``` 
the `ServerMaintenance` object will be remove by kube once the `BIOSSetting` resource is also removed.

File: PROJECT [hunk 54]
```diff
@@ -71,4 +71,12 @@ resources:
   kind: ServerMaintenance
   path: github.com/ironcore-dev/metal-operator/api/v1alpha1
   version: v1alpha1
+- api:
+    crdVersion: v1
+  controller: true
+  domain: ironcore.dev
+  group: metal
+  kind: BiosSettings
```
afritzler, line 79: Can we name the resource `BIOSSettings`?

afritzler (changes requested):

afritzler:
The PR looks good from my side. Just some minor/cosmetic things.

api/v1alpha1/biossettings_types.go
nagadeesh-nagaraja: Changed to Pending, InProgress, Applied, Failed

afritzler (approved):

stefanhipfel (approved):

Nuckal777 (changes requested):
What about moving `.status.state` and `.status.updateState` into transition condition? Would we do that in a follow-up PR?

nagadeesh-nagaraja:
> What about moving `.status.state` and `.status.updateState` into transition condition? Would we do that in a follow-up PR?
Yes, I have subTask created to do that. 
https://github.com/ironcore-dev/metal-operator/issues/295

api/v1alpha1/biossettings_types.go
```diff
@@ -24,7 +24,7 @@ type BIOSSettingsSpec struct {
 	ServerRef *corev1.LocalObjectReference `json:"serverRef,omitempty"`
 
 	// ServerMaintenancePolicy is maintenance policy to be enforced on the server.
-	ServerMaintenancePolicy ServerMaintenancePolicy `json:"ServerMaintenancePolicyTemplate,omitempty"`
+	ServerMaintenancePolicy ServerMaintenancePolicy `json:"ServerMaintenancePolicy,omitempty"`
```
Nuckal777, line 27: lowercase, please

Nuckal777 (approved):

PR #286, opened 2025-04-08
Remove `Completed` state from `ServerMaintenance`

stefanhipfel:
# Proposed Changes

- No need to set Maintenance to completed once finished. Just delete the Maintenance CRD to remove server from maintenance.
- Set serverMaintenanceRef earlier, once maintenance got approved

### DIFF
--- a/internal/controller/servermaintenance_controller.go
+++ b/internal/controller/servermaintenance_controller.go
@@ -107,18 +107,12 @@ func (r *ServerMaintenanceReconciler) handlePendingState(ctx context.Context, lo
 			return ctrl.Result{}, nil
 		}
 	}
-	defer func() {
-		if serverMaintenance.Status.State == metalv1alpha1.ServerMaintenanceStateInMaintenance {
-			// put server in maintenance
-			if patchErr := r.patchServerRef(ctx, log, serverMaintenance, server); patchErr != nil {
-				err = patchErr
-				log.Error(err, "failed to patch server maintenance ref")
-				return
-			}
-		}
-	}()
 	if server.Spec.ServerClaimRef == nil {
 		log.V(1).Info("Server has no claim, move to maintenance right away", "Server", server.Name)
+		if err = r.updateServerRef(ctx, log, serverMaintenance, server); err != nil {
+			log.Error(err, "failed to patch server maintenance ref")
+			return ctrl.Result{}, err
+		}
 		if modified, err := r.patchMaintenanceState(ctx, serverMaintenance, metalv1alpha1.ServerMaintenanceStateInMaintenance); err != nil || modified {
 			return ctrl.Result{}, err
 		}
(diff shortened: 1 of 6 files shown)


internal/controller/servermaintenance_controller_test.go
```diff
@@ -187,14 +187,9 @@ var _ = Describe("ServerMaintenance Controller", func() {
 		Eventually(Object(serverMaintenance)).Should(SatisfyAll(
 			HaveField("Status.State", metalv1alpha1.ServerMaintenanceStateInMaintenance),
 		))
-		By("Setting ServerMaintenance to Completed")
-		Eventually(UpdateStatus(serverMaintenance, func() {
-			serverMaintenance.Status.State = metalv1alpha1.ServerMaintenanceStateCompleted
-		})).Should(Succeed())
+		By("Deleting ServerMaintenance to finish the maintennce on the server")
+		Eventually(k8sClient.Delete).WithArguments(ctx, serverMaintenance).Should(Succeed())
```
afritzler, line 191:
```suggestion
		Eventually(k8sClient.Delete, serverMaintenance).Should(Succeed())
```

internal/controller/servermaintenance_controller_test.go
```diff
@@ -266,20 +261,8 @@ var _ = Describe("ServerMaintenance Controller", func() {
...
-			HaveField("Status.State", metalv1alpha1.ServerMaintenanceStateCompleted),
-		))
-
-		Eventually(Object(server)).Should(SatisfyAll(
-			HaveField("Status.State", metalv1alpha1.ServerStateAvailable),
-			HaveField("Spec.ServerMaintenanceRef", BeNil()),
-			HaveField("Spec.MaintenanceBootConfigurationRef", BeNil()),
-		))
+		By("Deleting first ServerMaintenance to finish the maintennce on the server")
+		Eventually(k8sClient.Delete).WithArguments(ctx, serverMaintenance01).Should(Succeed())
```
afritzler, line 265: Same as above.

internal/controller/servermaintenance_controller.go
```diff
@@ -108,6 +107,16 @@ func (r *ServerMaintenanceReconciler) handlePendingState(ctx context.Context, lo
 			return ctrl.Result{}, nil
 		}
 	}
+	defer func() {
```
afritzler, line 110: Instead of using `defer` (I know other projects are using this pattern in their controller impl.) can we put this at the correct place in the reconciliation loop? We are not using `defer` based cleanups/updates in our code atm.

afritzler (changes requested):

internal/controller/servermaintenance_controller_test.go
stefanhipfel: lint was complaining about it and suggested it

afritzler: Uhm what? Ok.

internal/controller/servermaintenance_controller.go
stefanhipfel: ok I just wanted to reduce redundancy.
Removed it

internal/controller/servermaintenance_controller_test.go
```diff
@@ -253,9 +248,26 @@ var _ = Describe("ServerMaintenance Controller", func() {
...
+		By("Checking the ServerMaintenanceRef")
+		Eventually(Object(server)).Should(SatisfyAll(
+			HaveField("Spec.ServerMaintenanceRef", Not(BeNil())),
+			HaveField("Spec.MaintenanceBootConfigurationRef", Not(BeNil())),
+		))
 		Eventually(Object(serverMaintenance01)).Should(SatisfyAll(
 			HaveField("Status.State", metalv1alpha1.ServerMaintenanceStateInMaintenance),
 		))
+		bootConfig := &metalv1alpha1.ServerBootConfiguration{}
+		Eventually(k8sClient.Get).WithArguments(ctx, types.NamespacedName{
```
afritzler, line 260: Can we use `komega` `Get()` here.

afritzler (changes requested):

internal/controller/servermaintenance_controller.go
```diff
@@ -239,41 +246,29 @@ func (r *ServerMaintenanceReconciler) setAndPatchServerPowerState(ctx context.Co
...
+		Kind:       "ServerMaintenance",
+		Namespace:  maintenance.Namespace,
+		Name:       maintenance.Name,
+		UID:        maintenance.UID,
 	}
-	log.V(1).Info("Server maintenance completed", "Server", server.Name)
-	if err := r.cleanup(ctx, log, server); err != nil {
-		return ctrl.Result{}, err
+	// use update to not overwrite ServerMaintenanceRef if another maintenance was quicker
+	if err := r.Update(ctx, server); err != nil {
```
afritzler, line 262: Why not do a `PATCH` here?

afritzler (changes requested):

internal/controller/servermaintenance_controller.go
stefanhipfel: I want it to fail if in the meantime another ServerMaintenanceRef was set. (for whatever reason that may happen)

afritzler (approved):

PR #297, opened 2025-04-14
Add more unit tests for BiosSettings controller

nagadeesh-nagaraja:
- update redfish_local to add ability create unit tests which does redfish posts
- add more unit tests which tests updating the bios settings.
- fix few bugs identified during the unit testing.


nagadeesh-nagaraja:
rebase & merge changes with ibiosSettings branch

bmc/redfish.go
```diff
@@ -448,35 +447,60 @@ func (r *RedfishBMC) CheckBiosAttributes(attrs map[string]string) (reset bool, e
...
+					fmt.Errorf(
+						"attribute '%s's' value '%v' has wrong type. needed '%s' for '%v'",
+						name,
+						value,
+						entryAttribute.Type,
+						entryAttribute,
+					))
+				break
+			}
+			var invalidEnum bool
```
Nuckal777, line 486: Should this be called `validEnum`?

bmc/redfish_local.go
```diff
@@ -54,3 +63,181 @@ func (r RedfishLocalBMC) PowerOff(ctx context.Context, systemUUID string) error
...
+// mock SetBiosAttributesOnReset sets given bios attributes for unit testing.
+func (r *RedfishLocalBMC) SetBiosAttributesOnReset(
+	ctx context.Context,
+	systemUUID string,
+	attributes redfish.SettingsAttributes,
+) (err error) {
+	attrs := make(map[string]interface{}, len(attributes))
+	for name, value := range attributes {
+		attrs[name] = value
+	}
```
Nuckal777, line 76: `attrs` appears to be unused.

bmc/redfish_local.go
```diff
@@ -54,3 +63,181 @@ func (r RedfishLocalBMC) PowerOff(ctx context.Context, systemUUID string) error
...
+		return nil, nil
+	}
+
+	mockedAttributes := r.getMockedBIOSSettingData()
+
+	filteredAttr, err := r.getFilteredBiosRegistryAttributes(false, false)
+	if err != nil {
+		return nil, err
+	}
+	result = make(map[string]interface{}, len(attributes))
```
Nuckal777, line 119: `map[string]any`

bmc/redfish_local.go
```diff
@@ -5,16 +5,25 @@ package bmc
...
 
 	"github.com/stmcginnis/gofish/redfish"
 )
 
 var _ BMC = (*RedfishLocalBMC)(nil)
 
+var defaultMockedBIOSSetting = []map[string]any{
+	{"name": "abc", "type": "string", "reboot": false, "value": "blah"},
+	{"name": "fooreboot", "type": "integer", "reboot": true, "value": 123},
+}
```
Nuckal777, line 20: It seems that `defaultMockedBIOSSetting` is mostly indexed by the `name` property by iterating, which has linear time complexity. Would it make to use a `map[string]map[string]any`, where the `name` of a setting is the key, e.g. `{"abc": {"type": "string", "reboot": false, "value": "blah"}}`? If this is very inconvenient at API boundaries, we can keep for the mock.

bmc/redfish_local.go
```diff
@@ -54,3 +63,181 @@ func (r RedfishLocalBMC) PowerOff(ctx context.Context, systemUUID string) error
...
+		data.Immutable = immutable
+		data.ReadOnly = readOnly
+		data.Type = eachMock["type"].(string)
+		data.ResetRequired = eachMock["reboot"].(bool)
+		filtered[eachMock["name"].(string)] = data
+	}
+	return filtered, err
+}
+
+// check if the arrtibutes need to reboot when changed, and are correct type.
```
Nuckal777, line 157: typo

bmc/redfish_local.go
```diff
@@ -54,3 +63,181 @@ func (r RedfishLocalBMC) PowerOff(ctx context.Context, systemUUID string) error
...
+					fmt.Errorf(
+						"attribute '%s's' value '%v' has wrong type. needed '%s' for '%v'",
+						name,
+						value,
+						entryAttribute.Type,
+						entryAttribute,
+					))
+				break
+			}
+			var invalidEnum bool
```
Nuckal777, line 219: Should this variable be called `validEnum`?

bmc/redfish_local.go
```diff
@@ -54,3 +63,181 @@ func (r RedfishLocalBMC) PowerOff(ctx context.Context, systemUUID string) error
...
+			for _, attrValue := range entryAttribute.Value {
+				if attrValue.ValueName == value.(string) {
+					invalidEnum = true
+					break
+				}
+			}
+			if !invalidEnum {
+				err = errors.Join(err, fmt.Errorf("attribute %s value is unknown. needed %v", name, entryAttribute.Value))
+			}
+			continue
```
Nuckal777, line 229: Superfluous.

bmc/redfish.go
```diff
@@ -448,35 +447,60 @@ func (r *RedfishBMC) CheckBiosAttributes(attrs map[string]string) (reset bool, e
...
+				if attrValue.ValueName == value.(string) {
+					invalidEnum = true
 					break
 				}
 			}
-			if !ok {
+			if !invalidEnum {
 				errs = append(errs, fmt.Errorf("attribute %s value is unknown. needed %v", name, entryAttribute.Value))
 			}
 			continue
```
Nuckal777, line 496: Superfluous.

bmc/redfish_local.go
```diff
@@ -54,3 +63,181 @@ func (r RedfishLocalBMC) PowerOff(ctx context.Context, systemUUID string) error
...
+	if err != nil {
+		return reset, err
+	}
+
+	if len(filtered) == 0 {
+		return reset, err
+	}
+	var errs []error
+	//TODO: add more types like maps etc
+	for name, value := range attrs {
```
Nuckal777, line 172: This loop seems nearly identical to the same loop in `bmc/redfish.go`. Would it make sense to extract that into a common function?

internal/controller/biossettings_controller_test.go
```diff
@@ -145,123 +140,150 @@ var _ = Describe("BIOSSettings Controller", func() {
...
+
+		By("Ensuring that the Server has the bios setting ref")
+		Eventually(Object(server)).Should(SatisfyAll(
+			HaveField("Spec.BIOSSettingsRef", &v1.LocalObjectReference{Name: biosSettings.Name}),
+		))
+
+		By("Ensuring that the Maintenance resource has not been created")
+		var serverMaintenanceList metalv1alpha1.ServerMaintenanceList
+		Eventually(ObjectList(&serverMaintenanceList)).Should(HaveField("Items", BeEmpty()))
+		Eventually(Object(biosSettings)).Should(SatisfyAll(
```
Nuckal777, line 176: For testing the absence of something, I would opt for `Consistently` instead of `Eventually`.

Nuckal777 (commented):
Thanks for looking into additional tests. 👍 I have another stylistic remark though. The function definitions contain named return values, but doesn't utilise them, so I would simply remove the names.

bmc/redfish_local.go
nagadeesh-nagaraja: @Nuckal777 The problem is that

during unit tests, we want to run in redfish_local.go  
during actual run, we want to run redfish.go 

hence i need to mock all the function anyway.... the interfaces  wont switch between them. 

I do not think we can mix them up? I can sync up with you offline to understand this better.

bmc/redfish_local.go
nagadeesh-nagaraja: @Nuckal777 I dont think I understand this. Could you please elaborate?

nagadeesh-nagaraja: ah understood, I think you mean the `continue`  statement  :D yes!

internal/controller/biossettings_controller_test.go
nagadeesh-nagaraja: added eventually and then consistently to stabilize the test.

nagadeesh-nagaraja:
fix some more flaky tests

nagadeesh-nagaraja:
rebase to origin/main

internal/controller/biossettings_controller_test.go
```diff
@@ -88,180 +112,434 @@ var _ = Describe("BIOSSettings Controller", func() {
...
+		Eventually(Object(server)).Should(SatisfyAll(
+			HaveField("Spec.BIOSSettingsRef", BeNil()),
+		))
+	})
+
+	It("should update the setting without maintenance if setting requested needs no server reboot", func(ctx SpecContext) {
+		BiosSetting := make(map[string]string)
+		// settings which does not reboot. mocked at
+		// metal-operator/bmc/redfish_local.go defaultMockedBIOSSetting
+		BiosSetting["abc"] = "blahblah"
```
afritzler, line 176: Can we change this too `foo: bar` :)

afritzler (changes requested):

afritzler (approved):

PR #298, opened 2025-04-16
Check for pending tasks during `BIOSSettings` update

nagadeesh-nagaraja:
- checks if the pending tasks are present at the state Pending: move to failure if present. 
- checks if the pending tasks are present at the state at InProgress: move to failure if unknown settings update are present
-  checks if the pending tasks are present before and after issuing the bmc setting update: skip issue of bios update if the pending tasks are present. check if the pending tasks are as expected post issuing bmc settings update.


bmc/redfish_local.go
```diff
@@ -46,6 +48,18 @@ func (r RedfishLocalBMC) PowerOn(ctx context.Context, systemUUID string) error {
...
+	// mock the bmc update here
+	if len(pendingMockedBIOSSetting) > 0 {
+
+		for key, data := range pendingMockedBIOSSetting {
+			if _, ok := defaultMockedBIOSSetting[key]; ok {
+				defaultMockedBIOSSetting[key] = data
+			}
+		}
+		pendingMockedBIOSSetting = map[string]map[string]any{}
+		r.StoredBMCSettingData = defaultMockedBIOSSetting
```
nagadeesh-nagaraja, line 61: StoredBIOSSettingData

bmc/bmc.go
```diff
@@ -44,6 +44,8 @@ type BMC interface {
 
 	GetBiosAttributeValues(ctx context.Context, systemUUID string, attributes []string) (redfish.SettingsAttributes, error)
 
+	GetPendingAttributeValues(ctx context.Context, systemUUID string) (redfish.SettingsAttributes, error)
```
nagadeesh-nagaraja, line 47: GetBiosPendingAttributeValues

bmc/redfish_local.go
```diff
@@ -62,6 +76,28 @@ func (r RedfishLocalBMC) PowerOff(ctx context.Context, systemUUID string) error
...
 }
 
+func (r *RedfishLocalBMC) GetPendingAttributeValues(
+	ctx context.Context,
+	systemUUID string,
+) (
+	redfish.SettingsAttributes,
+	error,
+) {
+	// can not unit test this ATM.
```
nagadeesh-nagaraja, line 86: remove

api/v1alpha1/biossettings_types.go
```diff
@@ -8,34 +8,28 @@ import (
...
-
 // BIOSSettingsSpec defines the desired state of BIOSSettings.
 type BIOSSettingsSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-
-	// BIOSSettings specifies the BIOS settings for the selected serverRef or serverSelector.
-	BIOSSettings Settings `json:"biosSettings,omitempty"`
+	// BMCSettingsSpec specifies the BIOS settings for the selected serverRef.
+	BMCSettingsSpec Settings `json:"biosSettingsSpec,omitempty"`
```
Nuckal777, line 14: Why are there BMC Settings inside the `BIOSSettings` resource?

nagadeesh-nagaraja: ah, this is a copy paste typo! 
I must be blind, I was reading this as BIOSSettingsSpec and was confused what this comment meant :D

api/v1alpha1/biossettings_types.go
```diff
@@ -8,34 +8,28 @@ import (
...
-
 // BIOSSettingsSpec defines the desired state of BIOSSettings.
 type BIOSSettingsSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-
-	// BIOSSettings specifies the BIOS settings for the selected serverRef or serverSelector.
-	BIOSSettings Settings `json:"biosSettings,omitempty"`
+	// BIOSSettingsSpec specifies the BIOS settings for the selected serverRef.
+	BIOSSettingsSpec Settings `json:"biosSettingsSpec,omitempty"`
```
Nuckal777, line 14: `BIOSSettingsSpec` seems redundant, because it is usually used as `object.Spec.BIOSSettingsSpec`. There "Spec" appears twice.

api/v1alpha1/biossettings_types.go
```diff
@@ -8,34 +8,28 @@ import (
...
 
 	// ServerRef is a reference to a specific server to apply bios setting on.
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="serverRef is immutable"
 	ServerRef *corev1.LocalObjectReference `json:"serverRef,omitempty"`
 
 	// ServerMaintenancePolicy is maintenance policy to be enforced on the server.
-	ServerMaintenancePolicy ServerMaintenancePolicy `json:"serverMaintenancePolicy,omitempty"`
+	ServerMaintenancePolicyType ServerMaintenancePolicy `json:"serverMaintenancePolicyType,omitempty"`
```
Nuckal777, line 21: The previous name was fine for me.

bmc/redfish.go
```diff
@@ -317,49 +316,39 @@ func (r *RedfishBMC) GetPendingAttributeValues(
...
 
-	err = json.Unmarshal(biosSettingRawBody, &tBiosSetting)
+	err = json.Unmarshal(RespRawBody, &entity)
 	if err != nil {
-		return result, err
+		return err
 	}
-
-	return tBiosSetting.Attributes, nil
+	return nil
```
Nuckal777, line 351: We could save a few lines here by using `return json.Unmarshal(...)` directly.

internal/controller/biossettings_controller.go
```diff
@@ -454,20 +486,55 @@ func (r *BiosSettingsReconciler) handleFailedState(
 	return ctrl.Result{}, nil
 }
 
+func (r *BiosSettingsReconciler) checkforPendingSettingsOnBIOS(
```
Nuckal777, line 489: This function is hard to reason about. Would it make sense to split it into two functions? One for checking that are any changes and a second for calculating the diff.

api/v1alpha1/biossettings_types.go
nagadeesh-nagaraja: In one of the issue, `ServerMaintenancePolicyType` was suggested for this name. and I liked it instead of just `ServerMaintenancePolicy ` . 
I am ok with either as long as we keep it same for all resources which will refer to it.

api/v1alpha1/biossettings_types.go
nagadeesh-nagaraja: The  resource is called `BIOSSettings` so technically it will be `BIOSSettings.Spec.BIOSSettings` or `BIOSSettings.Spec.BIOSSettingsSpec` I like the second because that pattern can extent to other resource which might have more info. 

like example in "BIOSVersion" https://github.com/ironcore-dev/metal-operator/issues/305 
we need more information than just the version. Hence changing the name to be more generic.

internal/controller/biossettings_controller.go
nagadeesh-nagaraja: agree, It will make it more simpler to read as well. will change this.

api/v1alpha1/biossettings_types.go
nagadeesh-nagaraja: changed to 
```
kind: BIOSSettings
metadata:
  name: biossettings-sample
spec:
  serverRef:
    name: endpoint-sample-system-0
  version: 2.10.2
  settings:
    PxeDev1EnDis: Enabled
  serverMaintenancePolicy: OwnerApproval
```

cmd/manager/main.go
```diff
@@ -378,6 +378,13 @@ func main() { // nolint: gocyclo
 		setupLog.Error(err, "unable to create controller", "controller", "BIOSSettings")
 		os.Exit(1)
 	}
+	// nolint:goconst
+	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
```
nagadeesh-nagaraja, line 382: missed changes from PR #315

afritzler (approved):

PR #323, opened 2025-04-30
Introduce `metal.ironcore.dev/reboot-needed` key

afritzler:
# Proposed Changes

Introduce a `Server` reboot key constant which can be used as an annotation key on a `Server` object to indicate that a `Server` needs a reboot.

This key can be used upstream e.g. as `Node` label/annotation to indicate the workload to perform a `Server` reboot.

/cc @nagadeesh-nagaraja


Nuckal777:
How this the reboot different from a maintenance for the consumer? In both cases the workload needs to be evacuated.

PR #324, opened 2025-04-30
Prepare CRDs folder with kustomize

SzymonSAP:
# Proposed Changes

### DIFF
--- a/.github/workflows/publish-chart.yml
+++ b/.github/workflows/publish-chart.yml
@@ -41,8 +41,16 @@ jobs:
           fi
           echo "version=$CHART_VERSION" >> $GITHUB_OUTPUT
 
-      - name: Move crds folder out of templates
-        run: mv dist/chart/templates/crd dist/chart/crds
+      - name: Install Kustomize
+        run: |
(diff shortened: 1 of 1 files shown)


config/crd/patches/webhook_in_serverclaims.yaml
```diff
@@ -9,8 +9,8 @@ spec:
     webhook:
       clientConfig:
         service:
-          namespace: system
-          name: webhook-service
+          namespace: metal-operator-system
```
afritzler, line 12: The `namespace` should be stay `system` as it will be overwritten via `PrefixName` in a kustomization upstream.

afritzler (changes requested):

config/crd/patches/webhook_in_serverclaims.yaml
SzymonSAP: So I found a simpler solution to generate all the manifests from config/default kustomize and select only CRDs from output with `yq`. This way we don't have to worry about namespace and namePrefix

afritzler (approved):

PR #335, opened 2025-05-13
Use 9443 port for webhook

SzymonSAP:
# Proposed Changes
- This change is needed as a service generated by helm kubebuilder plugin hardcode 9443 port for webhooks
- I also removed list indentation in manager manifest and changed the order of some fields


defo89:
@afritzler can you recall why we changed this port to `9445` some time ago? https://github.com/ironcore-dev/metal-operator/pull/113

afritzler:
The reason was that we were colliding with some other processes using this port. 🦭 I guess we can roll this back and override this on our side via `kustomize`.

afritzler (approved):

defo89:
As a follow up, I opened an issue in kubernetes-sigs/kubebuilder for this.

PR #337, opened 2025-05-22
Use uri to fetch the ComputeSystem from redfish API instead of UUID

nagadeesh-nagaraja:
This change will reduce the number API calls to BMC when getting the ComputeSystem from the BMC. replaces getSystemByUUID with getSystemFromURI. 

getSystemByUUID calls `redfish/V1/Systems` and then all the systems in the members field and then filters the required systems using UUID. hence making redundant calls. 

getSystemFromURI calls the `redfish/V1/Systems/<system ID>` directly to reduce number of calls. the URI is saved when the BMC creates the server, or when the server is reconciled for the first time Hence can be reused.


internal/controller/suite_test.go
```diff
@@ -221,7 +221,7 @@ func SetupTest() *corev1.Namespace {
 				PowerPollingTimeout:  200 * time.Millisecond,
 				BasicAuth:            true,
 			},
-			DiscoveryTimeout: time.Second, // Force timeout to be quick for tests
+			DiscoveryTimeout: 2 * time.Second, // Force timeout to be quick for tests
```
nagadeesh-nagaraja, line 224: need to revert :D

stefanhipfel (approved):

defo89:
Could you please extend the tests with this field? So far I found these:
https://github.com/ironcore-dev/metal-operator/blob/main/internal/controller/bmc_controller_test.go#L82
https://github.com/ironcore-dev/metal-operator/blob/main/internal/controller/bmc_controller_test.go#L152
https://github.com/ironcore-dev/metal-operator/blob/main/internal/controller/server_controller_test.go#L82

There is also seems to be a conflict, so rebase is needed.

defo89 (approved):

PR #339, opened 2025-05-22
Move `Server.Status` popolation from `ServerReconciler` to `BMCReconciler`

stefanhipfel:
# Proposed Changes

BMC controller now loads Server Status Details (Manufacturer, Storages, CPUs) once during creation of a server CRD and when setting the anntotation `"metal.ironcore.dev/operation" : "update-server-details"`.

BMC controller reconcile loop now only gets called when server was deleted or ops annotation was set.

Fixes:
Redfish Processors are now called separately, only when needed.
In General: Less calls to BMC redfish api

Server Controller Test:
Before: 700 API calls
After: 305 API calls


internal/controller/bmc_controller.go
```diff
@@ -185,6 +190,79 @@ func (r *BMCReconciler) discoverServers(ctx context.Context, log logr.Logger, bm
 	return nil
 }
 
+func (r *BMCReconciler) writeServerDetails(ctx context.Context, log logr.Logger, server *metalv1alpha1.Server, bmcClient bmc.BMC) error {
```
afritzler, line 193: Wouldn't `updateServerDetails` be a better method name?

internal/controller/bmc_controller_test.go
```diff
@@ -81,6 +82,45 @@ var _ = Describe("BMC Controller", func() {
 			HaveField("Spec.UUID", "38947555-7742-3448-3784-823347823834"),
 			HaveField("Spec.SystemUUID", "38947555-7742-3448-3784-823347823834"),
 			HaveField("Spec.BMCRef.Name", endpoint.Name),
+			HaveField("Status.Manufacturer", "Contoso"),
+			HaveField("Status.SKU", "8675309"),
+			HaveField("Status.SerialNumber", "437XR1138R2"),
```
afritzler, line 87: We probably should also add the `processor` status check as well.

internal/controller/bmc_controller_test.go
```diff
@@ -81,6 +82,45 @@ var _ = Describe("BMC Controller", func() {
...
+		))
+
+		zeroCapacity := resource.NewQuantity(0, resource.DecimalSI)
+		// force calculation of zero capacity string
+		_ = zeroCapacity.String()
+		Eventually(Object(server)).Should(SatisfyAll(
```
afritzler, line 93: Can we merge those 2 `Eventuallys` together?

internal/controller/server_controller_test.go
```diff
@@ -425,49 +421,8 @@ var _ = Describe("Server Controller", func() {
 		// reconciliation as part of handling the available state.
 		Eventually(Object(server)).Should(HaveField("Status.State", metalv1alpha1.ServerStateAvailable))
 
-		zeroCapacity := resource.NewQuantity(0, resource.DecimalSI)
```
afritzler, line 428: We should move the processeros status at the top of the test file also to the BMC controller test.

afritzler (changes requested):

Nuckal777:
I am still wondering, if the `BMC` reconciler should be responsible for filling in the state the state of `Server` objects. 🤔  The main concern is PR wants to address is superfluous calls to the redfish API by introducing some checks on status fields already being present. This can also be achieved in the `Server` reconciler.

stefanhipfel:
> I am still wondering, if the `BMC` reconciler should be responsible for filling in the state the state of `Server` objects. 🤔 The main concern is PR wants to address is superfluous calls to the redfish API by introducing some checks on status fields already being present. This can also be achieved in the `Server` reconciler.

BMC reconciler already sets state fields of Server objects.

PR #355, opened 2025-06-05
Adopt Helm charts for remote cluster use case

SzymonSAP:
# Proposed Changes

-

### DIFF
--- a/.github/workflows/publish-chart.yml
+++ b/.github/workflows/publish-chart.yml
@@ -60,13 +60,17 @@ jobs:
           curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
           mv kustomize /usr/local/bin
 
+      - name: Package Helm chart with crds folder in template
+        run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-chart.yml
```diff
@@ -60,13 +60,17 @@ jobs:
...
           mv kustomize /usr/local/bin
 
+      - name: Package Helm chart with crds folder in template
+        run: |
+          helm package dist/chart --version ${{ steps.chart_version.outputs.version }}-crds
+
       - name: Prepare CRDs folder
         run: |
           mkdir -p dist/chart/crds
           kustomize build config/default | yq ea 'select(.kind == "CustomResourceDefinition")' > dist/chart/crds/crds.yaml
```
Nuckal777, line 70: Could the CRDs be removed entirely? The chart without CRDs is intended for the cluster with workers in the remote setup case. There the CRDs are only required in the remote cluster. In the standard single cluster setup the chart with crds can still be used. 🤔

Nuckal777 (approved):

PR #359, opened 2025-06-10
Switch docs to vitepress

afritzler:
# Proposed Changes

Use vitepress to generate and publish our project documentation. The github action has been changed to also use vitepress to publish the documentation content as GH pages.

The current dev setup has been kept in takt by running `make startdocs` locally.


afritzler:
The "Project tag line thingy" has been addressed in the latest version and has now some proper text 🙈

PR #370, opened 2025-06-16
Transition unclaimed `Servers` from `Maintenance` to `Initial` state

stefanhipfel:
# Fixes
If maintenance is removed from a server, the server state should go to state INITIAL instead of AVAILABLE (if there is no claim on this server). Otherwise we could miss the discovery step. It could also make sense to go back to initial in case hardware changed during maintenance.

Flow:

Server state: 
Initial -> Maintenance -> Initial
Available -> Maintenance -> Initial
Reserved -> Maintenance -> Reserved.

### DIFF
--- a/docs/concepts/servers.md
+++ b/docs/concepts/servers.md
@@ -76,17 +76,19 @@ stateDiagram-v2
     Initial --> Discovery : Server object created
     Discovery --> Available : Discovery complete
     Available --> Reserved : ServerClaim created
+    Reserved --> Maintenance : Maintenance initiated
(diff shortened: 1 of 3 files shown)


stefanhipfel:
> Thanks @stefanhipfel! Do you think it makes sense to add a unit test for this case?

added tests

Nuckal777:
Docs need an update: https://github.com/ironcore-dev/metal-operator/blob/main/docs/concepts/servers.md?plain=1#L82

afritzler (approved):

PR #322, opened 2025-04-30
Webhook bmc setting

nagadeesh-nagaraja:
Fixes #307

note: add missed SetupBIOSSettingsWebhookWithManager in main.go

PR #371, opened 2025-06-17
Handle deleting of server maintenance

SzymonSAP:
When BiosSettings cr is removed a ServerMaintenance cr should also be removed. Currently it is handled by owner reference but this solution doesn't work because BiosSettings is a cluster scoped resource and ServerMaintenance is namespaced resource. This commit introduces deleting ServerMaintenance when BiosSettings is deleted in a reconcile function. It also changes owning to watching of ServerMaintenance by BiosSettings controller.

# Proposed Changes

-
-
-

Fixes #

### DIFF
--- a/internal/controller/biossettings_controller.go
+++ b/internal/controller/biossettings_controller.go
@@ -99,6 +99,15 @@ func (r *BiosSettingsReconciler) delete(
 	}
 	log.V(1).Info("ensured references were cleaned up")
 
+	if err, isBeingDeleted := r.deleteServerMaintenance(ctx, log, biosSettings); err != nil {
+		log.Error(err, "failed to delete server maintenance")
+		return ctrl.Result{}, err
+	} else if isBeingDeleted {
+		log.V(1).Info("server maintenance is being deleted, requeueing")
+		return ctrl.Result{}, nil // reconcile will be triggered again when server maintenance is deleted
+	}
+	log.V(1).Info("ensured server maintenance is deleted")
+
 	log.V(1).Info("Ensuring that the finalizer is removed")
 	if modified, err := clientutils.PatchEnsureNoFinalizer(ctx, r.Client, biosSettings, biosSettingsFinalizer); err != nil || modified {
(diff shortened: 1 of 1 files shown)


nagadeesh-nagaraja:
The bios Settings should not be deleted while in progress. ( this will be handled using webhook and in deletion logic) 
PRs are in review. 
https://github.com/ironcore-dev/metal-operator/pull/322 (webhook to not delete inProgress biosettings)
https://github.com/ironcore-dev/metal-operator/pull/351/ ( not to delete BIOSSettings when in Progress)

SzymonSAP:
> The bios Settings should not be deleted while in progress. ( this will be handled using webhook and in deletion logic) PRs are in review. #322 (webhook to not delete inProgress biosettings) #351 ( not to delete (continue reconcile) BIOSSettings when in Progress)

I see, but does it impact this PR? Here I delete `ServerMaintenance` not `BiosSettings` and logic from #351 can be merged before deleting `ServerMaintenance`.

nagadeesh-nagaraja:
> > The bios Settings should not be deleted while in progress. ( this will be handled using webhook and in deletion logic) PRs are in review. #322 (webhook to not delete inProgress biosettings) #351 ( not to delete (continue reconcile) BIOSSettings when in Progress)
> 
> I see, but does it impact this PR? Here I delete `ServerMaintenance` not `BiosSettings` and logic from #351 can be merged before deleting `ServerMaintenance`.



This diff will not affect the other diffs, but I am afraid that this might introduce issues. 
Example. 
Consider the situation that the BIOSSettings has failed and the server is in messed up state. if you delete the BIOSSettings and then that deletes the serverMaintenance. The system will go back to Available/Reserved state even though it is totally messed up. 

current behavior is not like that. In Failed state, the maintenance is not removed, so if I delete the BiosSettings, the server is still in Maintenance and that will give a chance to fix or revert the changes before moving the server out of maintenance (or maybe create new BIOSSettings with same maintenance Ref to fix the server).

PR #376, opened 2025-06-19
Validate that the claim in `Server` is not switched directly

### LINKED ISSUE #374: Multiple ServerClaims for the Same Server
**Describe the bug**
When creating multiple serverclaims with a label-selector (in parallel), we end up with several serverclaims claiming the same server (the server itself only reports one serverclaim, of course)

**To Reproduce**
Create more than one serverclaim with label-selector, the more the better, and watch which server they claim.

**Expected behavior**
Each serverclaim ends up with a different server (or none).

**Screenshots**
I replaced some internal names BMC-system-0 is really the same system, and we have two pool.
```
shoot--ccloud--<workers-1>-z1-5d8c6-2hlhn    BMC-system-0   ...     Bound     32m
shoot--ccloud--<workers-1>-z1-5d8c6-dpdgs   BMC-system-0   ...     Bound     86m
shoot--ccloud--<workers-2>-z1-74c9f-z4g64    BMC-system-0   ...     Bound     30m
```

**Additional context**

Probably one step closer at the root cause to #356

fwiesel:
The ServerClaim has first to be released, which would remove the reference in the server, and only then one should claim it again.

This rule should ensure that, and hopefully prevent (#374).

# Proposed Changes

- Prohibit the change of a ServerClaimRef from one ServerClaim to the other before going through "null ".

Fixes #374

### DIFF
--- a/api/v1alpha1/server_types.go
+++ b/api/v1alpha1/server_types.go
@@ -83,6 +83,8 @@ type ServerSpec struct {
 
 	// ServerClaimRef is a reference to a ServerClaim object that claims this server.
 	// This field is optional and can be omitted if no claim is associated with this server.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == null || oldSelf == null || self == oldSelf",message="serverClaimRef cannot be switched directly"
 	ServerClaimRef *v1.ObjectReference `json:"serverClaimRef,omitempty"`
 
 	// ServerMaintenanceRef is a reference to a ServerMaintenance object that maintains this server.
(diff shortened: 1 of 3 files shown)


fwiesel:
I haven' tested it, and I am not 100% sure that this is a desired API change.
So, I first wanted to check with you, if that is the case, before implementing a test.

Nuckal777:
Run `make generate && make docs && make helm` please.

fwiesel:
Sorry, I was on PTO. I've regenerated the code, and pushed the changes.

Nuckal777 (approved):

PR #377, opened 2025-06-19
Update ServerClaimRef in Server instead of Patch

### LINKED ISSUE #374: Multiple ServerClaims for the Same Server

fwiesel:
Patch opts out of the optimistic locking functionality of the kubernetes api, and is therefore doesn't protect against concurrent modifications.

The lock in the controller only protects against concurrent execution of the same function in the same controller, but does nothing to stop concurrent modifications
with any other user of the object.

# Proposed Changes

- Use Update instead of Patch, as the first checks for concurrent modifications.

Fixes #374

### DIFF
--- a/internal/controller/serverclaim_controller.go
+++ b/internal/controller/serverclaim_controller.go
@@ -198,20 +198,17 @@ func (r *ServerClaimReconciler) ensureObjectRefForServer(ctx context.Context, lo
 		return false, nil
 	}
 
-	if server.Spec.ServerClaimRef == nil {
-		serverBase := server.DeepCopy()
-		server.Spec.ServerClaimRef = &v1.ObjectReference{
-			APIVersion: "metal.ironcore.dev/v1alpha1",
-			Kind:       "ServerClaim",
-			Namespace:  claim.Namespace,
-			Name:       claim.Name,
-			UID:        claim.UID,
-		}
-		if err := r.Patch(ctx, server, client.MergeFrom(serverBase)); err != nil {
-			return false, fmt.Errorf("failed to patch claim ref for server: %w", err)
-		}
-		log.V(1).Info("Patched ServerClaim reference on Server", "Server", server.Name, "ServerClaimRef", claim.Name)
+	server.Spec.ServerClaimRef = &v1.ObjectReference{
+		APIVersion: "metal.ironcore.dev/v1alpha1",
+		Kind:       "ServerClaim",
+		Namespace:  claim.Namespace,
+		Name:       claim.Name,
+		UID:        claim.UID,
 	}
+	if err := r.Update(ctx, server); err != nil {
+		return false, fmt.Errorf("failed to patch claim ref for server: %w", err)
+	}
+	log.V(1).Info("Updated ServerClaim reference on Server", "Server", server.Name, "ServerClaimRef", claim.Name)
 	return true, nil
 }
 


afritzler:
How about we switch here to server-side apply (via `client.Apply`) instead of using `Update`.

Nuckal777:
In my opinion we should try to identify the root cause of the issue before committing any solution/workaround. #376 might be helpful for that.

fwiesel:
I'd argue the other way around: Patch forgoes any concurrency control k8s offers with its `resourceVersion`.
I think, one needs a good reason to deviate from the standard concurrency pattern k8s proposes. See [RetryOnConflict](https://pkg.go.dev/k8s.io/client-go@v0.32.1/util/retry#RetryOnConflict).

The patch method drops the `resourceVersion`, so any conflict will be "solved" by last-write-wins. That is what we observe.

It is true, I cannot tell you why that is exactly happening, because there is some mutex there, that *should* prevent that. 
You might want to investigate that to identify "the root-cause", but I think *should* is here an important qualifier, as it only addresses a single golang process, but we are talking here modifying the `Spec`, which is a public interface. Any other client can modify these fields "at will" concurrently to this golang process, and no lock in this operator can address that.

> How about we switch here to server-side apply (via client.Apply) instead of using Update.

As I read it, it is part of the `client-go` `client.Client`, but not `controller-runtime` `client.Client`.
One of my major gripes with `controller-runtime` is a lack of [server-side apply](https://github.com/kubernetes-sigs/controller-runtime/issues/347)

SzymonSAP:
I'm not 100% sure but I think after merging XValidation for `ServerClaimRef` field, patch should return an error. In case you still want to introduce locking I think a compromise would be to use patching with optimistic locking like here:
```
r.Patch(ctx, server, client.MergeFromWithOptions(serverBase, client.MergeFromWithOptimisticLock{}))
```
Patching starves the network less than Updates this is why it is preferred to use partial patching object than full object updates.

fwiesel:
> I'm not 100% sure but I think after merging XValidation...

I'd agree. The validation should catch a concurrent claim. My concern only is that it doesn't really seem "k8s" like. 
For me, it is you just declare the intent in the spec, and that will happen eventually. But that's my 2¢. We can close this, if you're happy with the validation.

> I think a compromise would be to use patching with optimistic locking like here:

That's a nice find, I wasn't aware of `MergeFromWithOptimisticLock`.

Nuckal777:
Closing in favor #389.

PR #380, opened 2025-06-25
CleanUp BMCSetting when BMC is deleted

nagadeesh-nagaraja:
# Proposed Changes
clean-up BMCSettings when BMC is deleted. this will help recreating/re-applying the BMC setting when the BMC is recreated. 
Do not create serverMaintenances if user gives maintenances for partial set of servers controlled by BMC. 

Fixes #


internal/controller/bmc_controller.go
```diff
@@ -63,6 +63,19 @@ func (r *BMCReconciler) reconcileExists(ctx context.Context, log logr.Logger, bm
 
 func (r *BMCReconciler) delete(ctx context.Context, log logr.Logger, bmcObj *metalv1alpha1.BMC) (ctrl.Result, error) {
 	log.V(1).Info("Deleting BMC")
+
+	if bmcObj.Spec.BMCSettingRef != nil {
+		BMCSetting := &metalv1alpha1.BMCSettings{}
+		if err := r.Get(ctx, client.ObjectKey{Name: bmcObj.Spec.BMCSettingRef.Name}, BMCSetting); err != nil {
+			if !errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: I am not sure if I can achieve it in 1 line.... are you suggesting something like

I want to ignore not found error, but any other error need to be errored out.. if no error, delete the resource. 

any suggestion here?

internal/controller/bmc_controller.go
```diff
@@ -63,6 +63,19 @@ func (r *BMCReconciler) reconcileExists(ctx context.Context, log logr.Logger, bm
 
 func (r *BMCReconciler) delete(ctx context.Context, log logr.Logger, bmcObj *metalv1alpha1.BMC) (ctrl.Result, error) {
 	log.V(1).Info("Deleting BMC")
+
+	if bmcObj.Spec.BMCSettingRef != nil {
+		bmcSettings := &metalv1alpha1.BMCSettings{}
+		if err := r.Get(ctx, client.ObjectKey{Name: bmcObj.Spec.BMCSettingRef.Name}, bmcSettings); err != nil {
```
afritzler, line 69:
```suggestion
		if err := r.Get(ctx, client.ObjectKey{Name: bmcObj.Spec.BMCSettingRef.Name}, bmcSettings); !errors.IsNotFound(err) != nil {
```

SzymonSAP:
both `BMC` and `BMCSettings` are cluster scoped. Can't we use owner reference for garbage collection?

nagadeesh-nagaraja:
> both `BMC` and `BMCSettings` are cluster scoped. Can't we use owner reference for garbage collection?

are you suggesting to change the owner of the resource?
`BMCSettings` Could be created/owned by someone else. I do not like idea of changing the owner.

SzymonSAP:
> > both `BMC` and `BMCSettings` are cluster scoped. Can't we use owner reference for garbage collection?
> 
> are you suggesting to change the owner of the resource? `BMCSettings` Could be created/owned by someone else. I do not like idea of changing the owner.

owner reference is a list and there can be multiple owners of one resource, but only one owner with `controller` set to true (`SetControllerReference` vs `SetOwnerReference`). So even if some other controllers creates `BMCSettings` it can still set owner reference without controller flag.
This is more like open question than proposition to make changes. I try to better understand ;)

nagadeesh-nagaraja:
> > > both `BMC` and `BMCSettings` are cluster scoped. Can't we use owner reference for garbage collection?
> > 
> > 
> > are you suggesting to change the owner of the resource? `BMCSettings` Could be created/owned by someone else. I do not like idea of changing the owner.
> 
> owner reference is a list and there can be multiple owners of one resource, but only one owner with `controller` set to true (`SetControllerReference` vs `SetOwnerReference`). So even if some other controllers creates `BMCSettings` it can still set owner reference without controller flag. This is more like open question than proposition to make changes. I try to better understand ;)

as far as I understand, resource will be cleaned up only if owner ref list is empty. so dont think we can have multiple owners and deleting one owner resource will get this resource cleaned up. 

but I think @afritzler @Nuckal777 might have better understanding of this?

internal/controller/bmcsettings_controller.go
```diff
@@ -539,9 +539,26 @@ func (r *BMCSettingsReconciler) requestMaintenanceOnServers(
 		return false, nil
 	}
 
+	// if user gave some server with serverMaintenance but not all
+	// we want to request maintenance for the missing servers only.
+	// find the servers which has maintenance and do not create maintenance for them.
```
afritzler, line 544: Can we rephrase this comment a bit?

afritzler (changes requested):
Just some small issue with the comments - apart from that the PR looks good.

PR #382, opened 2025-06-26
Add an owner of bios settings to a server

### LINKED ISSUE #366: Remove `BIOSSettings` when `Server` is removed
**Describe the bug**
If I delete a bmc and rediscover it the applied bios setting are not touched / reset and reapplied

**To Reproduce**
Create a bmc apply bios settings and delete the bmc

**Expected behavior**
Bios settings have an owner reference and are deleted along with the bmc

SzymonSAP:
# Proposed Changes

- Add an owner of bios settings to a server

Fixes #366

### DIFF
--- a/internal/controller/biossettings_controller.go
+++ b/internal/controller/biossettings_controller.go
@@ -195,6 +195,12 @@ func (r *BiosSettingsReconciler) reconcile(ctx context.Context, log logr.Logger,
 		log.V(1).Info("referred server object could not be fetched")
 		return ctrl.Result{}, err
 	}
+
+	if err := r.setOwnerToServer(ctx, biosSettings, server); err != nil {
+		log.V(1).Error(err, "failed to set bios settings owner reference to server", "biosSettings", biosSettings.Name, "server", server.Name)
+		return ctrl.Result{}, fmt.Errorf("failed to set owner reference to server: %w", err)
+	}
+
 	// patch server with biossettings reference
 	if server.Spec.BIOSSettingsRef == nil {
 		if err := r.patchBiosSettingsRefOnServer(ctx, log, server, &corev1.LocalObjectReference{Name: biosSettings.Name}); err != nil {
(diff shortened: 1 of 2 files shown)


internal/controller/biossettings_controller.go
```diff
@@ -229,6 +235,21 @@ func (r *BiosSettingsReconciler) reconcile(ctx context.Context, log logr.Logger,
 	return r.ensureBIOSSettingsStateTransition(ctx, log, bmcClient, biosSettings, server)
 }
 
+func (r *BiosSettingsReconciler) setOwnerToServer(ctx context.Context, biosSettings *metalv1alpha1.BIOSSettings, server *metalv1alpha1.Server) error {
+	if controllerutil.HasControllerReference(biosSettings) {
```
nagadeesh-nagaraja, line 239: What is the expected behavior if the its been controlled by others? 

The controller of this resource needs to delete it if the server is deleted?

internal/controller/biossettings_controller.go
```diff
@@ -229,6 +235,21 @@ func (r *BiosSettingsReconciler) reconcile(ctx context.Context, log logr.Logger,
...
+		return nil
+	}
+
+	biosSettingsBase := biosSettings.DeepCopy()
+	if err := controllerutil.SetControllerReference(server, biosSettings, r.Scheme); err != nil {
```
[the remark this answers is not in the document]
SzymonSAP: Deleting with ownership or finalizers are basically two working solution. Both are common in current code. I agree that deleting it with finalizers is easier to understand and has less edge cases. So I can go with deleting it as you suggest

internal/controller/biossettings_controller.go
nagadeesh-nagaraja: today, we do not have anything. but in future we might have multiple of them creating and controlling it... 

true, server might not be the right place for it. May be we can do it inside `BIOSSettings` itself? 
I am not entirely sure of what is the best practice. 

letting the `server` delete ( or even `BIOSSettings` delete itself if server is not found) might be a common place and easier implementation than asking each controller owning `BIOSSettings` to handle the deletion individually, if its not against kubernetes principles.

consider this example: we will definitely have `BIOSSettingSet` in the future. it creates the `BIOSSettings` for each (selected) server. 

when server is deleted, it also deletes the referred BIOSSettings. The `BIOSSettingSet` then re-creates `BIOSSettings` again when the server is added again. of-course we could also add deletion logic here ( but that will not handle the cases where `BIOSSettings` was not created by it) 


may be we can have a discussion next meeting and see what is the best approach?

nagadeesh-nagaraja (approved):

PR #384, opened 2025-06-30
Add `BIOSVersionSet` resource and controller

nagadeesh-nagaraja:
# Proposed Changes

Fixes #


api/v1alpha1/biosversionset_types.go
```diff
@@ -0,0 +1,81 @@
...
+	Completed int32 `json:"completed,omitempty"`
+	// Failed is the total number of failed server in the set.
+	Failed int32 `json:"failed,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:printcolumn:name="BIOSVersion",type=string,JSONPath=`.spec.version`
+// +kubebuilder:printcolumn:name="serverLabelSelector",type=string,JSONPath=`.spec.serverSelector`
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: any suggestion if I will be able to print this? even if like a map or something?

internal/controller/biosversionset_controller.go
```diff
@@ -0,0 +1,350 @@
...
+		err = r.updateStatus(ctx, log, currentStatus, biosVersionSet)
+		if err != nil {
+			log.Error(err, "failed to update current Status")
+			return ctrl.Result{}, err
+		}
+		log.Info("Waiting on the created BIOSVersion to reach terminal status")
+		return ctrl.Result{}, nil
+	}
+
+	if err := r.deleteBIOSVersions(ctx, log, &metalv1alpha1.ServerList{}, ownedBiosVersions); err != nil {
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: I dont think I understand correctly. do you mean to say that we remove finalizer instead of deleteing the BIOSSettings here?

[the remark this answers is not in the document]
SzymonSAP: That said we don't have to delete `BIOSVersion`, it should be handled by k8s garbage collector based on ownership.

internal/controller/biosversionset_controller.go
```diff
@@ -0,0 +1,350 @@
...
+
+	serverMap := make(map[string]bool)
+	for _, server := range serverList.Items {
+		serverMap[server.Name] = true
+	}
+
+	var errs []error
+	for _, biosVersion := range biosVersionList.Items {
+		if !serverMap[biosVersion.Spec.ServerRef.Name] {
+			if biosVersion.Status.State == metalv1alpha1.BIOSVersionStateInProgress {
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: we do both webhook and the controller check during deletion: ( as long as the finalizers are present)
```
1. block deleting, so return an error from webhook, 
2. allow setting deletion time stamp in BIOSVersion even if in progress and remove its finalizer only after the state has changed.
```

I think its good point not to continue deletion of biosVersionSet when all biosVersion is not deleted. I can add that checks!

internal/controller/biosversionset_controller.go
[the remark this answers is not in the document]
nagadeesh-nagaraja: ah, so basically, this step is redundant? 

but I think its clearer to make sure that you have cleaned up biosVersion before removing your finalizer? what do you think?

internal/controller/biosversionset_controller.go
```diff
@@ -0,0 +1,355 @@
...
+					return []ctrl.Request{{
+						NamespacedName: client.ObjectKey{
+							Name:      biosVersionSet.Name,
+							Namespace: biosVersionSet.Namespace,
+						}}}
+				}
+			}
+		}
+	}
+	return nil
```
SzymonSAP, line 345:
```suggestion
	reqs := make([]ctrl.Request, 0)
	for _, biosVersionSet := range biosVersionSetList.Items {
		selector, err := metav1.LabelSelectorAsSelector(&biosVersionSet.Spec.ServerSelector)
		if err != nil {
			log.Error(err, "failed to convert label selector")
			return nil
		}
		if selector.Matches(labels.Set(host.GetLabels())) {
			reqs = append(reqs, ctrl.Request{
				NamespacedName: client.ObjectKey{
					Name:      biosVersionSet.Name,
					Namespace: biosVersionSet.Namespace,
				},
			})
		}
	}
	return reqs
}
```
When calling `getServersBySelector` you list Servers but later you simply filter only host from them. It is much easier to simply check if host matches label selector. I also don't quite get why you added `if len(ownedBiosVersion.Items) != len(serverList.Items)`. I assume to prevent reqonciling without changes but it shouldn't be done like this. One more reconcile without change doesn't make a different but to save some extra work we can add a pridicate to watch. Please check my next comment.

internal/controller/biosversionset_controller.go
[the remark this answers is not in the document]
nagadeesh-nagaraja: added check to wait until deleting all the biosVersion before deleting self. this gives us a overview in case any biosVersion is unable to delete....

api/v1alpha1/biosversionset_types.go
```diff
@@ -0,0 +1,81 @@
...
+	// details regarding the image to use to upgrade to given BIOS version
+	// +required
+	Image ImageSpec `json:"image"`
+
+	// ServerSelector specifies a label selector to identify the servers that are to be selected.
+	// +required
+	ServerSelector metav1.LabelSelector `json:"serverSelector"`
+
+	// ServerMaintenancePolicy is maintenance policy to be enforced on the server.
+	ServerMaintenancePolicy ServerMaintenancePolicy `json:"serverMaintenancePolicy,omitempty"`
```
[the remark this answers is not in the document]
SzymonSAP: Why didn't you add `ServerMaintenancePolicy` to common spec?

internal/controller/biosversionset_controller.go
```diff
@@ -0,0 +1,346 @@
...
+			newBiosVersionName := fmt.Sprintf("%s-%s", biosVersionSet.Name, server.Name)
+			if len(newBiosVersionName) > utilvalidation.DNS1123SubdomainMaxLength {
+				log.V(1).Info("BiosVersion name is too long, it will be shortened using randan", "name", newBiosVersionName)
+				key := make([]byte, 10)
+				size, err := rand.Read(key)
+				if err != nil || size != len(key) {
+					return fmt.Errorf("failed to generate random bytes for BIOSVersion name: %w. size returned %d", err, size)
+				}
+				newBiosVersionName = fmt.Sprintf("%s-%s", biosVersionSet.Name, string(key))
+			}
```
SzymonSAP, line 189: It still might fail if `biosVersionSet.Name` is equal `utilvalidation.DNS1123SubdomainMaxLength`. It is impossible to avoid completely name collisions and too long name at the same time but this should be very rare case. Why don't you want to simply trim name to max length? If you still prefer to add this random string at the end lets go with well tested `GenerateName` from `ObjectMeta`.

internal/controller/biosversionset_controller.go
```diff
@@ -0,0 +1,346 @@
...
+
+			newBiosVersion := &metalv1alpha1.BIOSVersion{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: newBiosVersionName,
+				}}
+
+			opResult, err := controllerutil.CreateOrPatch(ctx, r.Client, newBiosVersion, func() error {
+				newBiosVersion.Spec.VersionUpdateSpec = *biosVersionSet.Spec.VersionUpdateSpec.DeepCopy()
+				newBiosVersion.Spec.ServerRef = &corev1.LocalObjectReference{Name: server.Name}
+				return controllerutil.SetControllerReference(biosVersionSet, newBiosVersion, r.Client.Scheme())
```
SzymonSAP, line 199: Previously you were also coping `ServerMaintenancePolicy`. It is absent in `VersionUpdateSpec` so it won't be copied. Is it not needed?

nagadeesh-nagaraja: this was an oversight mistake. Will fix it.

internal/controller/biosversionset_controller.go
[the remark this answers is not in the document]
nagadeesh-nagaraja: have deleted the logic to remove biosVersion in deletion logic

internal/controller/biosversionset_controller.go
```diff
@@ -0,0 +1,335 @@
...
+	for _, biosVersion := range biosVersionList.Items {
+		switch biosVersion.Status.State {
+		case metalv1alpha1.BIOSVersionStateCompleted:
+			currentStatus.Completed += 1
+		case metalv1alpha1.BIOSVersionStateFailed:
+			currentStatus.Failed += 1
+		case metalv1alpha1.BIOSVersionStateInProgress:
+			currentStatus.InProgress += 1
+		case metalv1alpha1.BIOSVersionStatePending, "":
+			currentStatus.InProgress += 1
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: Set does not have Pending...I have merged it with InProgress. at set level, I dont think it make much difference.

api/v1alpha1/biosversion_types.go
```diff
@@ -29,19 +29,24 @@ const (
...
+type VersionUpdateSpec struct {
 	// Version contains a BIOS version to upgrade to
 	// +required
 	Version string `json:"version"`
 
-	// An indication of whether the server's upgrade service should bypass vendor update policies
+	// Policy An indication of whether the server's upgrade service should bypass vendor update policies
 	// +optional
-	UpdatePolicy *UpdatePolicy `json:"updatePolicy,omitempty"`
+	Policy *UpdatePolicy `json:"updatePolicy,omitempty"`
```
afritzler, line 39: Isn't `Policy` to generic?

api/v1alpha1/biosversionset_types.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// BIOSVersionSetSpec defines the desired state of BIOSVersionSet.
+type BIOSVersionSetSpec struct {
+
```
afritzler, line 15: Please remove newline.

api/v1alpha1/biosversionset_types.go
```diff
@@ -0,0 +1,75 @@
+// SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and IronCore contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
afritzler, line 10: Please remove those comments.

api/v1alpha1/biosversionset_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// ServerSelector specifies a label selector to identify the servers that are to be selected.
+	// +required
+	ServerSelector metav1.LabelSelector `json:"serverSelector"`
+
+	// BiosVersionTemplate defines the template for the BIOSversion Resource to be applied to the servers.
+	BiosVersionTemplate VersionUpdateSpec `json:"biosVersionTemplate,omitempty"`
+
+	// ServerMaintenancePolicy is a maintenance policy to be enforced on the server.
+	// +optional
+	ServerMaintenancePolicy ServerMaintenancePolicy `json:"serverMaintenancePolicy,omitempty"`
```
afritzler, line 25: Do we need that field here? We have a `serverMaintenancePolicy` field in the `BIOSVersionSpec`. I would remove that from the `Set` here.

api/v1alpha1/biosversionset_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// TotalVersionResource is the number of Settings current created by the set.
+	TotalVersionResource int32 `json:"totalVersionResource,omitempty"`
+	// Pending is the total number of pending server in the set.
+	Pending int32 `json:"pending,omitempty"`
+	// InMaintenance is the total number of server in the set that are currently in InProgress.
+	InProgress int32 `json:"inProgress,omitempty"`
+	// Completed is the total number of completed server in the set.
+	Completed int32 `json:"completed,omitempty"`
+	// Failed is the total number of failed server in the set.
+	Failed int32 `json:"failed,omitempty"`
```
afritzler, line 41: Can we change those status fields and have something similar like the `ReplicaSets` in k8s? https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#replicasetstatus-v1-apps

internal/controller/biosversionset_controller.go
```diff
@@ -0,0 +1,351 @@
...
+			opResult, err := controllerutil.CreateOrPatch(ctx, r.Client, newBiosVersion, func() error {
+				newBiosVersion.Spec.VersionUpdateSpec = *biosVersionSet.Spec.BiosVersionTemplate.DeepCopy()
+				newBiosVersion.Spec.ServerMaintenancePolicy = biosVersionSet.Spec.ServerMaintenancePolicy
+				newBiosVersion.Spec.ServerRef = &corev1.LocalObjectReference{Name: server.Name}
+				return controllerutil.SetControllerReference(biosVersionSet, newBiosVersion, r.Client.Scheme())
+			})
+			if err != nil {
+				errs = append(errs, err)
+			}
+			log.V(1).Info("Created biosVersion", "BIOSVersion", newBiosVersion.Name, "server ref", server.Name, "Operation", opResult)
```
afritzler, line 193:
```suggestion
			log.V(1).Info("Created BIOSVersion", "BIOSVersion", newBiosVersion.Name, "server ref", server.Name, "Operation", opResult)
```

internal/controller/bmcsettings_controller.go
```diff
@@ -576,11 +577,22 @@ func (r *BMCSettingsReconciler) requestMaintenanceOnServers(
...
+					Namespace:    r.ManagerNamespace,
+					GenerateName: newServerMaintenanceName[:utilvalidation.DNS1123SubdomainMaxLength-10] + "-",
+				}}
+		} else {
+			serverMaintenance = &metalv1alpha1.ServerMaintenance{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: r.ManagerNamespace,
+					Name:      newServerMaintenanceName,
+				}}
+		}
```
afritzler, line 595: This is IMO a valid improvement. However I would put this into a separate PR.

afritzler:
Thanks for the PR @nagadeesh-nagaraja. Can we also include the docs under `docs/conecpts/biosversionset.md`?

api/v1alpha1/biosversionset_types.go
nagadeesh-nagaraja: moved it to `VersionUpdateSpec`

api/v1alpha1/biosversionset_types.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// BIOSVersionSetSpec defines the desired state of BIOSVersionSet.
+type BIOSVersionSetSpec struct {
+	// ServerSelector specifies a label selector to identify the servers that are to be selected.
+	// +required
+	ServerSelector metav1.LabelSelector `json:"serverSelector"`
+
+	// BiosVersionTemplate defines the template for the BIOSversion Resource to be applied to the servers.
+	BiosVersionTemplate VersionUpdateSpec `json:"biosVersionTemplate,omitempty"`
```
afritzler, line 17: Why not using here the `BIOSVersionSpec` type instead of splitting it in the `BIOSVersion`?

afritzler (changes requested):

api/v1alpha1/biosversionset_types.go
nagadeesh-nagaraja: BIOSVersionSpec has `ServerRef *corev1.LocalObjectReference `json:"serverRef,omitempty"`` which we do not needed as a spec in BIOSVersionSet....

afritzler (approved):

PR #386, opened 2025-07-01
Add Conditions to`BIOSSettings` status

### LINKED ISSUE #295: Add transition `Conditions` to `BIOSSettings` resource
https://github.com/ironcore-dev/metal-operator/pull/267

use Status to transition state. use Condition instead. 
 
             I think we should think about the state transitions and especially the naming here. I would assume there should only be things like `Pending`, `InProgress`, `Failed` and `Applied`. It don't think it is a good idea to do a fine grained state transitioning here. For that I would rather use transition conditions.

_Originally posted by @afritzler in https://github.com/ironcore-dev/metal-operator/pull/267#discussion_r2036607559_

nagadeesh-nagaraja:
Fixes #295

also fixes following: 

reset BMC before starting, to reduce flakiness 
Add lastAppliedTime status to provide details on when setting was applied. 
fix, bmc version json path in kubebuilder print. 
add timeout to error out when setting update is stalled for more than 2 hours.


internal/controller/biossettings_controller.go
```diff
@@ -757,45 +1093,43 @@ func (r *BiosSettingsReconciler) updateBiosSettingsStatus(
...
-	biosSettingsBase := biosSettings.DeepCopy()
-	biosSettings.Status.UpdateSettingState = state
-
 	if err := r.Status().Patch(ctx, biosSettings, client.MergeFrom(biosSettingsBase)); err != nil {
-		return fmt.Errorf("failed to patch BIOSSettings UpdateSetting status: %w", err)
+		return fmt.Errorf("failed to patch BIOSSettings status: %w", err)
 	}
 
-	log.V(1).Info("Updated biosSettings UpdateSetting state ", "new state", state)
+	log.V(1).Info("Updated biosSettings state ", "new state", state, "new conditions", biosSettings.Status.Conditions)
```
afritzler, line 1132: We should use here something like `"State", state` in the info key value map. That way it is easier to query for that in the logging backend.

internal/controller/biossettings_controller.go
```diff
@@ -38,9 +39,26 @@ type BiosSettingsReconciler struct {
...
 	BMCOptions       bmc.Options
 	ResyncInterval   time.Duration
+	TimeoutExpiry    time.Duration
 }
 
-const biosSettingsFinalizer = "metal.ironcore.dev/biossettings"
+const (
+	biosSettingsFinalizer = "metal.ironcore.dev/biossettings"
+
+	serverMaintenanceCreatedCondition = "serverMaintenanceCreated"
```
afritzler, line 48: Can we use a camel case naming for the condition types? Something like what you would see in the `Pod` status: 
```kind: Pod
status:
  conditions:
    - lastTransitionTime: "2019-10-22T16:29:24Z"
      status: "True"
      type: PodScheduled
    - lastTransitionTime: "2019-10-22T16:29:24Z"
      status: "True"
      type: Initialized
    - lastTransitionTime: "2019-10-22T16:29:31Z"
      status: "True"
      type: ContainersReady
    - lastTransitionTime: "2019-10-22T16:29:31Z"
      status: "True"
      type: Ready
```

api/v1alpha1/biossettings_types.go
```diff
@@ -61,17 +61,24 @@ type BIOSSettingsStatus struct {
 
 	// State represents the current state of the bios configuration task.
 	State BIOSSettingsState `json:"state,omitempty"`
-	// UpdateSettingState represents the current state of the bios setting update task.
-	UpdateSettingState BIOSSettingUpdateState `json:"updateSettingState,omitempty"`
+
+	AppliedStateTimeStamp *metav1.Time `json:"appliedTimeStamp,omitempty"`
```
afritzler, line 65: What is the purpose of this status field? "Last successful applied"?

internal/controller/biossettings_controller.go
```diff
@@ -122,12 +140,26 @@ func (r *BiosSettingsReconciler) cleanupServerMaintenanceReferences(
...
 		if metav1.IsControlledBy(serverMaintenance, biosSettings) {
 			// if the biosSettings is not being deleted, update the
 			log.V(1).Info("Deleting server maintenance", "serverMaintenance Name", serverMaintenance.Name, "state", serverMaintenance.Status.State)
+			acc := conditionutils.NewAccessor(conditionutils.AccessorOptions{})
+			condition, err = r.getCondition(acc, biosSettings.Status.Conditions, serverMaintenanceDeletedCondition)
+			if err != nil {
+				return fmt.Errorf("failed to get the delete condition while clean up maintenance %v", err)
+			}
+			if err := acc.Update(
+				condition,
```
afritzler, line 156: I would probably also make sense to add the transition timestamp here?

afritzler (changes requested):

api/v1alpha1/biossettings_types.go
nagadeesh-nagaraja: To print when was the settings applied when we do kubectl get ...... 
this will show us when was the settings applied to the server.

internal/controller/biossettings_controller.go
nagadeesh-nagaraja: I dont think i understand. sorry for stupid question....

but I checked and they seems to be all camelCased?

```
	serverMaintenanceCreatedCondition = "serverMaintenanceCreated"
	serverMaintenanceDeletedCondition = "serverMaintenanceDeleted"
	pendingVersionUpdateCondition     = "pendingBIOSVersionUpdate"
	pendingSettingCheckCondition      = "pendingSettingStateCheck"
	timeoutStartCondition             = "settingUpdateStartTime"
	timedOutCondition                 = "timedOutDuringSettingUpdate"
	turnServerOnCondition             = "turnServerOnCondition"
	issueSettingsUpdateCondition      = "issueSettingsUpdate"
	unknownPendingSettingCondition    = "unknownPendingSettingStateCheck"
	skipRebootCondition               = "SkipServerRebootPostSettingUpdate"
	rebootPowerOffCondition           = "rebootPowerOff"
	rebootPowerOnCondition            = "rebootPowerOn"
	verifySettingCondition            = "verifySettingsPostUpdate"
```

internal/controller/biossettings_controller.go
afritzler: Ah ok

api/v1alpha1/biossettings_types.go
afritzler: Got it. Can we come up with a better name for this status field like `lastSuccessfulUpdate`?

stefanhipfel (approved):

PR #387, opened 2025-07-08
Add `Server` validation webhook

nagadeesh-nagaraja:
Prohibit the deletion of `Server` object when `Server` is in `Maintenance` state.


internal/controller/server_controller.go
```diff
@@ -120,6 +120,14 @@ func (r *ServerReconciler) reconcileExists(ctx context.Context, log logr.Logger,
 }
 
 func (r *ServerReconciler) delete(ctx context.Context, log logr.Logger, server *metalv1alpha1.Server) (ctrl.Result, error) {
+	if !controllerutil.ContainsFinalizer(server, ServerFinalizer) {
+		return ctrl.Result{}, nil
+	}
+
+	if server.Status.State == metalv1alpha1.ServerStateMaintenance {
+		log.V(1).Info("postponing delete as server is in Maintenance state")
```
afritzler, line 128: Skipping instead of postponing. We typically start the `Info()` logs with an upper case.

internal/controller/bmcsettings_controller.go
```diff
@@ -91,6 +91,12 @@ func (r *BMCSettingsReconciler) delete(
 	if !controllerutil.ContainsFinalizer(bmcSetting, BMCSettingFinalizer) {
 		return ctrl.Result{}, nil
 	}
+
+	if bmcSetting.Status.State == metalv1alpha1.BMCSettingsStateInProgress {
+		log.V(1).Info("postponing delete as Settings update is in progress")
```
afritzler, line 96: Same as below.

afritzler (changes requested):

internal/webhook/v1alpha1/biossettings_webhook.go
```diff
@@ -115,3 +135,23 @@ func checkForDuplicateBiosSettingsRefToServer(
...
 	return nil, nil
 }
+
+func checkBiosSettingsUpdateDuringInProgress(
+	biosSettingsList *metalv1alpha1.BIOSSettingsList,
+	biosSettings *metalv1alpha1.BIOSSettings,
+) error {
+	for _, bs := range biosSettingsList.Items {
+		if biosSettings.Name == bs.Name {
+			if bs.Status.State == metalv1alpha1.BIOSSettingsStateInProgress && controllerutil.ContainsFinalizer(&bs, controller.BIOSSettingsFinalizer) {
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: ah! its redundant.. 
intention was to let force update if user wanted to. I think I need to better solution here.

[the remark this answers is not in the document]
nagadeesh-nagaraja: Option was between 
```
if bs.Status.State == metalv1alpha1.BIOSSettingsStateInProgress && controllerutil.ContainsFinalizer(biosSettings, controller.BIOSSettingsFinalizer)
```
or 
```
if bs.Status.State == metalv1alpha1.BIOSSettingsStateInProgress && controllerutil.ContainsFinalizer(biosSettings, <new finalizer to indicate force resource update>)
```

I have choosen second option to keep it clean.

SzymonSAP:
I don't quite get `ForceUpdateResource` finalizer idea. You added it in this PR but you only check it with `controllerutil.ContainsFinalizer`, never set to any resource. Could you describe what do you want to achieve with this and use it in code?

nagadeesh-nagaraja:
> I don't quite get `ForceUpdateResource` finalizer idea. You added it in this PR but you only check it with `controllerutil.ContainsFinalizer`, never set to any resource. Could you describe what do you want to achieve with this and use it in code?


there is the generic Idea. 

When the resources are InProgress, we should not let them be updated. However, I would like to provide a backdoor to update when the resource InProgress. if they set the finalizer with `ForceUpdateResource`, the webhook will allow updating the resource when in  InProgress state.

internal/webhook/v1alpha1/biossettings_webhook.go
```diff
@@ -86,7 +100,16 @@ func (v *BIOSSettingsCustomValidator) ValidateDelete(ctx context.Context, obj ru
...
+	bs := &metalv1alpha1.BIOSSettings{}
+	err := v.Client.Get(ctx, client.ObjectKey{
+		Name:      biossettings.GetName(),
+		Namespace: biossettings.GetNamespace(),
+	}, bs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get BMCSettings: %w", err)
+	}
+
+	if bs.Status.State == metalv1alpha1.BIOSSettingsStateInProgress {
```
SzymonSAP: Why do you have to get `BIOSSettings` and not to use `obj`?

internal/webhook/v1alpha1/server_webhook.go
```diff
@@ -0,0 +1,75 @@
...
+// ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type Server.
+func (v *ServerCustomValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	server, ok := obj.(*metalv1alpha1.Server)
+	if !ok {
+		return nil, fmt.Errorf("expected a Server object but got %T", obj)
+	}
+	serverlog.Info("Validation for Server upon deletion", "name", server.GetName())
+
+	s := &metalv1alpha1.Server{}
+	err := v.Client.Get(ctx, client.ObjectKey{
```
SzymonSAP, line 62: the same as above

internal/webhook/v1alpha1/server_webhook.go
```diff
@@ -0,0 +1,75 @@
...
+		Name:      server.GetName(),
+		Namespace: server.GetNamespace(),
+	}, s)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get BMCSettings: %w", err)
+	}
+
+	if s.Status.State == metalv1alpha1.ServerStateMaintenance && controllerutil.ContainsFinalizer(s, controller.ServerFinalizer) {
```
SzymonSAP, line 70:
```suggestion
	if s.Status.State == metalv1alpha1.ServerStateMaintenance && ShouldAllowForceDeleteInProgress(s, controller.ServerFinalizer) {
```

SzymonSAP (commented):
I got a few minor comments and I also want to ask you to do some environment tests before merging to make sure it works as expected

internal/webhook/v1alpha1/helper.go
```diff
@@ -0,0 +1,18 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ShouldAllowForceUpdateInProgress checks if the object should force allow update.
+func ShouldAllowForceUpdateInProgress(obj client.Object) bool {
+	val, found := obj.GetAnnotations()[metalv1alpha1.ForceUpdateAnnotation]
+	if !found {
+		return false
+	}
+	return val == metalv1alpha1.OperationAnnotationForceUpdateInProgress
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: obj.GetAnnotations() wont return a uninitialized map. 

I think this logic need not be further optimized, let me know your thought.

internal/webhook/v1alpha1/biossettings_webhook.go
nagadeesh-nagaraja: checked the code, looks like obj contains the copy of what is present in the API.

internal/controller/server_controller.go
```diff
@@ -113,21 +113,33 @@ func (r *ServerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 }
 
 func (r *ServerReconciler) reconcileExists(ctx context.Context, log logr.Logger, server *metalv1alpha1.Server) (ctrl.Result, error) {
-	if !server.DeletionTimestamp.IsZero() {
+	if delete := r.shouldDelete(log, server); delete {
```
SzymonSAP, line 116: the same as above

SzymonSAP (approved):
I got one more suggestion, except that LGTM :) Good job!

Nuckal777 (approved):

PR #410, opened 2025-07-29
Use `GeneratedNames` when creating `ServerMaintenance` for `BIOSSettings`

### LINKED ISSUE #409: `BMCSettings` fails to create `ServerMaintenances` when names are longer than 253
when server name and BMCSettings name add up to more than 253, the BMCSettings controller fails to create ServerMaintenances. 

solution: fix by cropping and generating the names, when the name is more than 253.

nagadeesh-nagaraja:
Fixes #409

### DIFF
--- a/internal/controller/biosversionset_controller_test.go
+++ b/internal/controller/biosversionset_controller_test.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/utils/ptr"
 
 	metalv1alpha1 "github.com/ironcore-dev/metal-operator/api/v1alpha1"
+	"github.com/ironcore-dev/metal-operator/bmc"
 )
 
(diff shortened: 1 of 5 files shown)


afritzler:
Q: Why are we not just using a `GeneratedName` approach and store the name in the `Status` of the settings object?

nagadeesh-nagaraja:
> Btw. why do we want to encode the `ServerName` into the `ServerMaintenance` object in the first place? Would it be good enough to just create the `ServerMaintenance` object with the same name as the `BIOSSettings`?

we will have multiple servers under BMC, hence we will need to differentiate. 

I think using the GenerateName is a good option.

afritzler (approved):

PR #412, opened 2025-08-01
Add BMCVersionSet resource and controller

### LINKED ISSUE #420: Add BMCVersionSet type and controller
Create a Set for BMCVersionSet to handle multiple BMC to update BMC version at a time.

Example CRD Spec 

```yaml
kind: BMCVersionSet
metadata:
  name: bmc-version-set-sample
spec:
  BMCSelector:
    matchLabels:
       Manufacturer: a1
       Platform: foo
  bmcVersionTemplate:
    version: 2.10.3 # version to update to
    ServerMaintenancePolicy: OwnerApproval  # Policy an defined in ServerMaintenance resource
    updatePolicy: force
    image: 
        uri: https://bar-2.10.3.flash.img
        transferProtocol: https
status:
    FullyLabeledBMCs: 10  # is the number of BMC selected in the set.
    AvailableBMCVersion: 10 # number of BMCVersion we current have created
    PendingBMCVersion: 5 # is the number of BMCVersion update is Pending
    InProgressBMCVersion: 5 # is the number of BMCVersion update is InProcess
    FailedBMCVersion: 2 # is the number BMCVersion update has FAILED
    CompletedBMCVersion: 3 # is the number of BMCVersion update has Completed
```

workflow: 
- Given the selector field, all BMC matching that label will be selected to apply the specified version update. 
- Will create BMCVersion resource for each selected BMC. 
- The controller will monitor the changes to BMC (delete/create/ updates to label) and respond to create/delete the BMCVersion resources. 
- The Controller will list the BMC selected. BMCVersion created, and their current state as (Pending/InProgress/Failed/Completed)

nagadeesh-nagaraja:
Fixes #420

### DIFF
--- a/internal/controller/bmcversionset_controller.go
+++ b/internal/controller/bmcversionset_controller.go
@@ -169,7 +169,7 @@ func (r *BMCVersionSetReconciler) createMissingBMCVersions(
 		if _, ok := BMCWithBMCVersion[bmc.Name]; !ok {
 			newBMCVersion := &metalv1alpha1.BMCVersion{
 				ObjectMeta: metav1.ObjectMeta{
-					GenerateName: "bmc-version-update-through-set-",
+					GenerateName: "bmc-version-set-",
 				}}
 
 			opResult, err := controllerutil.CreateOrPatch(ctx, r.Client, newBMCVersion, func() error {
--- a/api/v1alpha1/bmcversion_types.go
+++ b/api/v1alpha1/bmcversion_types.go
@@ -26,7 +26,7 @@ type BMCVersionTemplate struct {
 	// +required
 	Version string `json:"version"`
 
-	// UpdatePolicy An indication of whether the server's upgrade service should bypass vendor update policies
+	// UpdatePolicy is an indication of whether the server's upgrade service should bypass vendor update policies
 	// +optional
 	UpdatePolicy *UpdatePolicy `json:"updatePolicy,omitempty"`
 
@@ -45,7 +45,7 @@ type BMCVersionTemplate struct {
 
 // BMCVersionSpec defines the desired state of BMCVersion.
 type BMCVersionSpec struct {
-	// defines the template for Version to be applied on the server's BMC.
+	// BMCVersionTemplate defines the template for BMC version to be applied on the server's BMC.
 	BMCVersionTemplate `json:",inline"`
 
 	// BMCRef is a reference to a specific BMC to apply BMC upgrade on.
--- a/config/crd/bases/metal.ironcore.dev_bmcversions.yaml
+++ b/config/crd/bases/metal.ironcore.dev_bmcversions.yaml
@@ -172,8 +172,8 @@ spec:
                   type: object
                 type: array
               updatePolicy:
-                description: UpdatePolicy An indication of whether the server's upgrade
-                  service should bypass vendor update policies
+                description: UpdatePolicy is an indication of whether the server's
+                  upgrade service should bypass vendor update policies
                 type: string
               version:
                 description: Version contains a BMC version to upgrade to
--- a/config/crd/bases/metal.ironcore.dev_bmcversionsets.yaml
+++ b/config/crd/bases/metal.ironcore.dev_bmcversionsets.yaml
@@ -205,7 +205,7 @@ spec:
                       type: object
                     type: array
                   updatePolicy:
-                    description: UpdatePolicy An indication of whether the server's
+                    description: UpdatePolicy is an indication of whether the server's
                       upgrade service should bypass vendor update policies
                     type: string
                   version:
--- a/dist/chart/templates/crd/metal.ironcore.dev_bmcversions.yaml
+++ b/dist/chart/templates/crd/metal.ironcore.dev_bmcversions.yaml
@@ -178,8 +178,8 @@ spec:
                   type: object
                 type: array
               updatePolicy:
-                description: UpdatePolicy An indication of whether the server's upgrade
-                  service should bypass vendor update policies
+                description: UpdatePolicy is an indication of whether the server's
+                  upgrade service should bypass vendor update policies
                 type: string
               version:
                 description: Version contains a BMC version to upgrade to
--- a/dist/chart/templates/crd/metal.ironcore.dev_bmcversionsets.yaml
+++ b/dist/chart/templates/crd/metal.ironcore.dev_bmcversionsets.yaml
@@ -211,7 +211,7 @@ spec:
                       type: object
                     type: array
                   updatePolicy:
-                    description: UpdatePolicy An indication of whether the server's
+                    description: UpdatePolicy is an indication of whether the server's
                       upgrade service should bypass vendor update policies
                     type: string
                   version:
--- a/docs/api-reference/api.md
+++ b/docs/api-reference/api.md
@@ -1794,7 +1794,7 @@ BMCVersionTemplate
 <p>
 (Members of <code>BMCVersionTemplate</code> are embedded into this type.)
 </p>
-<p>defines the template for Version to be applied on the server&rsquo;s BMC.</p>
+<p>BMCVersionTemplate defines the template for BMC version to be applied on the server&rsquo;s BMC.</p>
 </td>
 </tr>
 <tr>
@@ -2067,7 +2067,7 @@ BMCVersionTemplate
 <p>
 (Members of <code>BMCVersionTemplate</code> are embedded into this type.)
 </p>
-<p>defines the template for Version to be applied on the server&rsquo;s BMC.</p>
+<p>BMCVersionTemplate defines the template for BMC version to be applied on the server&rsquo;s BMC.</p>
 </td>
 </tr>
 <tr>
@@ -2208,7 +2208,7 @@ UpdatePolicy
 </td>
 <td>
 <em>(Optional)</em>
-<p>UpdatePolicy An indication of whether the server&rsquo;s upgrade service should bypass vendor update policies</p>
+<p>UpdatePolicy is an indication of whether the server&rsquo;s upgrade service should bypass vendor update policies</p>
 </td>
 </tr>
 <tr>


api/v1alpha1/bmcversionset_types.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// BMCVersionSetSpec defines the desired state of BMCVersionSet.
+type BMCVersionSetSpec struct {
+	// BMCSelector specifies a label selector to identify the servers that are to be selected.
+	// +required
+	BMCSelector metav1.LabelSelector `json:"serverSelector"`
```
Nuckal777, line 14: Use `bmcSelector` in the json tag, please.

api/v1alpha1/bmcversionset_types.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// BMCVersionTemplate defines the template for the BMCversion Resource to be applied to the servers.
+	BMCVersionTemplate BMCVersionTemplate `json:"bmcVersionTemplate,omitempty"`
+}
+
+// BMCVersionSetStatus defines the observed state of BMCVersionSet.
+type BMCVersionSetStatus struct {
+	// fullyLabeledServers is the number of server in the set.
+	FullyLabeledServers int32 `json:"fullyLabeledServers,omitempty"`
```
Nuckal777, line 23: How this field different to `AvailableBMCVersion`? Also the suffix needs be to be changed to `BMC`.

internal/controller/bmcversionset_controller.go
```diff
@@ -0,0 +1,339 @@
...
+	metalv1alpha1 "github.com/ironcore-dev/metal-operator/api/v1alpha1"
+)
+
+// BMCVersionSetReconciler reconciles a BMCVersionSet object
+type BMCVersionSetReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+const BMCVersionSetFinalizer = "metal.ironcore.dev/BMCVersionSet"
```
Nuckal777, line 33: Should the finalizer be lowercase analogous to the `bmcversion` finalizer? (Technically, the `biosversionset` would also need to be tweaked.)

internal/controller/bmcversionset_controller.go
```diff
@@ -0,0 +1,339 @@
...
+	if shouldIgnoreReconciliation(bmcVersionSet) {
+		log.V(1).Info("Skipped BMCVersionSet reconciliation")
+		return ctrl.Result{}, nil
+	}
+
+	if modified, err := clientutils.PatchEnsureFinalizer(ctx, r.Client, bmcVersionSet, BMCVersionSetFinalizer); err != nil || modified {
+		return ctrl.Result{}, err
+	}
+
+	serverList, err := r.getBMCBySelector(ctx, bmcVersionSet)
```
Nuckal777, line 115: `bmcList` instead of `serverList`.

internal/controller/bmcversionset_controller.go
```diff
@@ -0,0 +1,339 @@
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	ownedBMCVersions, err := r.getOwnedBMCVersions(ctx, bmcVersionSet)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	log.V(1).Info("Summary of servers and BMCVersions", "Server count", len(serverList.Items),
```
Nuckal777, line 125: `BMC count` instead of `server count`.

internal/controller/bmcversionset_controller.go
```diff
@@ -0,0 +1,339 @@
...
+
+func (r *BMCVersionSetReconciler) createMissingBMCVersions(
+	ctx context.Context,
+	log logr.Logger,
+	bmcList *metalv1alpha1.BMCList,
+	bmcVersionList *metalv1alpha1.BMCVersionList,
+	bmcVersionSet *metalv1alpha1.BMCVersionSet,
+) error {
+
+	BMCWithBMCVersion := make(map[string]bool)
```
Nuckal777, line 161: Could be a `map[string]struct{}`.

internal/controller/bmcversionset_controller.go
```diff
@@ -0,0 +1,339 @@
...
+
+func (r *BMCVersionSetReconciler) deleteOrphanBMCVersions(
+	ctx context.Context,
+	log logr.Logger,
+	bmcList *metalv1alpha1.BMCList,
+	bmcVersionList *metalv1alpha1.BMCVersionList,
+) ([]string, error) {
+
+	bmcMap := make(map[string]bool)
+	for _, server := range bmcList.Items {
```
Nuckal777, line 196: `bmc` instead of `server`.

internal/controller/bmcversionset_controller.go
```diff
@@ -0,0 +1,339 @@
...
+
+	var errs []error
+	var warnings []string
+	for _, bmcVersion := range bmcVersionList.Items {
+		if !bmcMap[bmcVersion.Spec.BMCRef.Name] {
+			if bmcVersion.Status.State == metalv1alpha1.BMCVersionStateInProgress {
+				log.V(1).Info("waiting for BMCVersion to move out of InProgress state", "BMCVersion", bmcVersion.Name, "status", bmcVersion.Status)
+				warnings = append(warnings, fmt.Sprintf("BMCVersion %s is still in progress, skipping deletion", bmcVersion.Name))
+				continue
+			}
```
Nuckal777, line 208: I wonder how relevant this check is. Could we just issue the delete here and let the `BMCVersion` controller decide how to handle the situation of a deleted `BMC` object?

internal/controller/bmcversionset_controller.go
```diff
@@ -0,0 +1,339 @@
...
+		// if the host label matches the selector, enqueue the request
+		if selector.Matches(labels.Set(host.GetLabels())) {
+			reqs = append(reqs, ctrl.Request{
+				NamespacedName: client.ObjectKey{
+					Name:      bmcVersionSet.Name,
+					Namespace: bmcVersionSet.Namespace,
+				},
+			})
+		} else { // if the label has been removed
+			ownedBMCVersions, err := r.getOwnedBMCVersions(ctx, &bmcVersionSet)
```
Nuckal777, line 311: Could this list call be done once outside the loop?

internal/controller/bmcversionset_controller_test.go
```diff
@@ -0,0 +1,402 @@
...
+				HaveField("Status.CompletedBMCVersion", BeNumerically("==", 2)),
+				HaveField("Status.InProgressBMCVersion", BeNumerically("==", 0)),
+				HaveField("Status.FailedBMCVersion", BeNumerically("==", 0)),
+			))
+
+			By("Deleting the resource")
+			Expect(k8sClient.Delete(ctx, bmcVersionSet)).To(Succeed())
+		})
+
+		It("should successfully reconcile the resource when server are deleted/created", func(ctx SpecContext) {
```
Nuckal777, line 217: `bmc` instead of `server`.

config/samples/metal_v1alpha1_bmcversionset.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: metal.ironcore.dev/v1alpha1
+kind: BMCVersionSet
+metadata:
+  labels:
+    app.kubernetes.io/name: metal-operator
+    app.kubernetes.io/managed-by: kustomize
+  name: bmcversionset-sample
+spec:
+  # TODO(user): Add fields here
```
Nuckal777, line 9: The example should be completed.

Nuckal777 (commented):
Thanks. 👍 The implementation looks sound to me in general. I got some questions and found a few copy-paste errors.

api/v1alpha1/bmcversionset_types.go
nagadeesh-nagaraja: This shows the count of BMC filtered. AvailableBMCVersion shows the count of BMCVersion created. 
they converge to same number. but during transitions (BMC deletion/addition/label change), there can be differences

internal/controller/bmcversionset_controller.go
nagadeesh-nagaraja: we need `bmcVersionSet` Hence better to do it here? 

taking a extreme example, if no labels are deleted on the BMC. we dont make this call?
if we move it outside of loop,  we will have to do it for all bmcVersionSet in the List.

internal/controller/bmcversionset_controller.go
nagadeesh-nagaraja: when webhook is enabled, `r.Delete(ctx, &bmcVersion);` will throw error when BIOSVersion is InProgress
its better not to call it when we know it will error anyway.

internal/controller/bmcversionset_controller.go
Nuckal777: I missed that `r.getOwnedBMCVersions(ctx, &bmcVersionSet)` depends on the currently iterated `bmcVersionSet`, so this does not apply.

internal/controller/bmcversionset_controller.go
```diff
@@ -0,0 +1,340 @@
...
+	for _, bmcVersion := range bmcVersionList.Items {
+		BMCWithBMCVersion[bmcVersion.Spec.BMCRef.Name] = struct{}{}
+	}
+
+	var errs []error
+	for _, bmc := range bmcList.Items {
+		if _, ok := BMCWithBMCVersion[bmc.Name]; !ok {
+			newBMCVersion := &metalv1alpha1.BMCVersion{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "bmc-version-update-through-set-",
```
defo89, line 172: Are we ok with this long resource names or should we maybe shorten it?

PR #413, opened 2025-08-01
Add pure go Redfish mock server implementation

afritzler:
# Proposed Changes

Add a pure go Redfish mock server implementation as a replacement for the containerized local Redfish setup. The `Makefile` targets for `make startbmc` and `make stopbmc` will stay around and will be removed in the future. The server can be run locally via:

```
 go run ./bmc/mock/main.go  
```

The `data` dir contains the same configuration as provided by the official mock server (https://github.com/DMTF/Redfish-Mockup-Server/tree/main/public-rackmount1).

## Other improvments
- Fixed typos in tests
- Fixed `context` handling in `registry` server

### DIFF
--- a/bmc/bmc.go
+++ b/bmc/bmc.go
@@ -104,13 +104,11 @@ type BMC interface {
 	// GetStorages retrieves storage information for the system.
 	GetStorages(ctx context.Context, systemURI string) ([]Storage, error)
 
+	// GetProcessors retrieves processor information for the system.
 	GetProcessors(ctx context.Context, systemURI string) ([]Processor, error)
 
-	UpgradeBiosVersion(
-		ctx context.Context,
-		manufacturer string,
-		parameters *redfish.SimpleUpdateParameters,
-	) (string, bool, error)
+	// UpgradeBiosVersion upgrades the BIOS version for the system.
+	UpgradeBiosVersion(ctx context.Context, manufacturer string, parameters *redfish.SimpleUpdateParameters) (string, bool, error)
 
 	// GetBiosUpgradeTask retrieves the task for the BIOS upgrade.
 	GetBiosUpgradeTask(ctx context.Context, manufacturer string, taskURI string) (*redfish.Task, error)
--- a/bmc/mock/server/server.go
+++ b/bmc/mock/server/server.go
@@ -21,22 +21,23 @@ import (
 
 var (
 	//go:embed data/**
-	dataFS    embed.FS
-	overrides sync.Map
+	dataFS embed.FS
 )
 
 type MockServer struct {
-	log     logr.Logger
-	addr    string
-	handler http.Handler
-	mu      sync.Mutex
+	log       logr.Logger
+	addr      string
+	handler   http.Handler
+	mu        sync.RWMutex
+	overrides map[string]any
 }
 
 func NewMockServer(log logr.Logger, addr string) *MockServer {
 	mux := http.NewServeMux()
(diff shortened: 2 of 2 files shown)


afritzler:
@nagadeesh-nagaraja we might want to harmonize the `BIOS/BMCSettings` and `Version` mock setup into this mock server in the future.

bmc/redfish_local.go
```diff
@@ -21,79 +22,83 @@ type RedfishLocalBMC struct {
...
 			return
 		}
+
 		system.PowerState = redfish.OffPowerState
 		system.RawData = nil
-		systemURI := fmt.Sprintf("/redfish/v1/Systems/%s", system.ID)
-		if err := system.Patch(systemURI, system); err != nil {
-			log := ctrl.LoggerFrom(ctx)
-			log.V(1).Error(err, "failed to Patch system to power off", "systemURI", systemURI)
+		fullURI := fmt.Sprintf("/redfish/v1/Systems/%s", system.ID)
```
afritzler, line 91: Do we need here the full uri or is the `systemUIR` which is passed down enought?

nagadeesh-nagaraja: technically, the systemURI should be same as `fmt.Sprintf("/redfish/v1/Systems/%s", system.ID)`. so we could directly use that.

bmc/redfish_local.go
```diff
@@ -21,79 +22,83 @@ type RedfishLocalBMC struct {
...
+	}(ctx)
 	return nil
 }
 
-func (r RedfishLocalBMC) PowerOff(ctx context.Context, systemURI string) error {
+func (r *RedfishLocalBMC) PowerOff(ctx context.Context, systemURI string) error {
+	go func(ctx context.Context) {
+		select {
+		case <-ctx.Done():
+			return
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: as we have our own mock for the redfish, we should be able to move all these delay logic to simulate the server turn-off/on/reset into the mock itself.

nagadeesh-nagaraja (commented):
This will help us in a lot of ways!
can get rid of all modification to redfish_local.go :) 
and have stable tests!

Thank you so much for this :)

bmc/mock/main.go
```diff
@@ -0,0 +1,41 @@
...
+	srv := server.NewMockServer(log, ":8000")
+
+	if err := srv.Start(ctx); err != nil {
+		log.Error(err, "Failed to start mock server")
+		return
+	}
+
+	<-ctx.Done()
+
+	time.Sleep(1 * time.Second)
```
Nuckal777, line 39: Context handling seems a bit iffy to me. Something like the following across `main.go` and `server.go` should suffice:
```go
ctx, cancel := signal.NotifyContext()
defer cancel()

go server.ListenAndServe()

<-ctx.Done()
timeout, cancel := context.WithTimeout(context.Background(), time.Second)
defer cancel()
server.Shutdown(timeout)
```

afritzler: Wall my idea was to have a cancelable `Start()` method which I can use in the suite test setup and stop the server via `defer` cancelling the context. With the suggested solution I need to replicate what has been done in `main.go` in every suite test setup or?

bmc/redfish_local.go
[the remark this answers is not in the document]
afritzler: Rolled back those changes.

Issue #332, opened 2025-05-12
Extend `Server.status.networkInterfaces`


defo89:
# Summary

Include all physical interfaces to `Server.status.networkInterfaces` (device name, model, ip, mac, speed).
This will help to identify correct device parameters which could be useful for creating ignition config.

Follow-up:

- remove any reference to `ClientIP` in boot-operator CR specs (separate issue tbd)

PR #415, opened 2025-08-05
Extend `Server.status.networkInterfaces`

videlov:
# Proposed Changes

- Adds more physical interface details to `Server.status.networkInterfaces`: pci address, model and speed.

Fixes https://github.com/ironcore-dev/metal-operator/issues/332


videlov:
> Do you think we can use https://github.com/shirou/gopsutil or https://github.com/jaypipes/ghw to extract the network device specific information? This module is already multi OS enabled.

Like the `gopsutil` much better but unfortunately it do not cover what we need here. Will use `ghw`.

videlov:
image

PR #441, opened 2025-08-20
Support fuzzy UUID match for registry

hardikdr:
# Proposed Changes

-
-
-

Fixes #


.github/workflows/publish-docker.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       with:
         context: .
         platforms: linux/amd64,linux/arm64
-        push: ${{ github.event_name != 'pull_request' }}
+        push: ${{ github.event_name == 'pull_request' }}
```
damyan, line 84: Please use `ok-to-image` if you want a PR docker build and revert that change

PR #462, opened 2025-09-04
Handle Ignore operations in sets and ServerMaintenance

nagadeesh-nagaraja:
# Proposed Changes

- add ignore operation annotation handling in serverMaintenance Controller
- Propagates the ignore operation annotation from BIOS/BMC Settings/Version Sets to the child objects. 
- Does not Claim server which is ongoing Maintenance or has ongoing Maintenance queue.

### DIFF
--- a/internal/controller/serverclaim_controller.go
+++ b/internal/controller/serverclaim_controller.go
@@ -306,17 +306,19 @@ func (r *ServerClaimReconciler) claimServer(ctx context.Context, log logr.Logger
 		return nil, err
 	}
 
-	var selectedServer *metalv1alpha1.Server
-	var prevSelectedServer *metalv1alpha1.Server
+	// fetch previously claimed server if its present
+	// keeping this separate for reason of clarity
 	for _, server := range serverList.Items {
 		// find previously claimed server
 		if ref := server.Spec.ServerClaimRef; ref != nil {
 			if ref.UID == claim.UID && ref.Name == claim.Name && ref.Namespace == claim.Namespace {
-				prevSelectedServer = &server
-				break
+				return &server, nil
 			}
 		}
+	}
 
+	var selectedServer *metalv1alpha1.Server
+	for _, server := range serverList.Items {
 		selector, err := metav1.LabelSelectorAsSelector(claim.Spec.ServerSelector)
 		if err != nil {
 			return nil, err
@@ -346,12 +348,8 @@ func (r *ServerClaimReconciler) claimServer(ctx context.Context, log logr.Logger
 		// we continue through the loop to find if previously claimed server is also present
 		if selectedServer == nil && r.isServerClaimable(ctx, log, &server, claim) {
 			selectedServer = &server
+			break
 		}
-	}
-
-	// use previously claimed server if its found
-	if prevSelectedServer != nil {
-		selectedServer = prevSelectedServer
 	}
 	if selectedServer == nil {
 		return nil, nil


Nuckal777 (commented):
Thanks. 👍 I have a general question regarding the annotations. I always understood `metal.ironcore.dev/operation=ignore` as in stop reconciling this specific object and don't do anything based on it's metadata/spec, which would include annotating child objects. Should a `metal.ironcore.dev/operation=ignore-childs` or similar be added, which would mean reconcile this object, but add the ignore annotation to all child resources? 🤔

internal/controller/serverclaim_controller.go
```diff
@@ -364,6 +385,12 @@ func (r *ServerClaimReconciler) claimServerByReference(ctx context.Context, log
 		log.V(1).Info("Server is not powered off", "Server", server.Name, "PowerState", server.Status.PowerState)
 		return nil, nil
 	}
+	isUnderMaintenance, err := r.isUnderMaintenanceQueue(ctx, log, server)
+	// is undergoing maintenance and not in Reserved State, we should not claim this server
+	if server.Status.State != metalv1alpha1.ServerStateReserved && (err != nil || isUnderMaintenance) {
+		log.V(1).Info("Server is undergoing Maintenances", "Server", server.Name, "error", err)
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
Nuckal777: That's not what I meant. I am proposing to filter out all invalid servers from the server list in `claimServer()` before it is passed to the selection functions.

api/v1alpha1/constants.go
```diff
@@ -6,8 +6,14 @@ package v1alpha1
...
 	OperationAnnotation = "metal.ironcore.dev/operation"
-	// OperationAnnotationIgnore skips the reconciliation of a resource if set to true.
+	// PropagatedOperationAnnotation indicates which operation should be performed outside the current spec definition flow.
+	PropagatedOperationAnnotation = "metal.ironcore.dev/operation-propogated"
+	// OperationAnnotationIgnore skips the reconciliation of a resource if OperationAnnotation is set to this.
 	OperationAnnotationIgnore = "ignore"
+	// OperationAnnotationIgnoreChild skips the reconciliation of a resource's Child if OperationAnnotation is set to this.
+	OperationAnnotationIgnoreChild = "ignore-child"
+	// PropagatedOperationAnnotationIgnored skips the reconciliation of a resource if PropagatedOperationAnnotation is set to this.
+	PropagatedOperationAnnotationIgnored = "propagated-ignore"
```
Nuckal777, line 16: Usage of this value in full reads `metal.ironcore.dev/operation-propogated: propagated-ignore`. Maybe we can simply reuse `ignore` here?

internal/controller/biossettingsset_controller.go
```diff
@@ -268,6 +305,14 @@ func (r *BIOSSettingsSetReconciler) patchBIOSSettingsfromTemplate(
 				biosSettings.Spec.BIOSSettingsTemplate = *biosSettingsTemplate.DeepCopy()
 				biosSettings.Spec.ServerMaintenanceRef = serverMaintenanceRef
 			}
+
+			if isChildIgnoredThroughSets(&biosSettings) {
```
Nuckal777, line 309: Is a negation missing here by chance?

internal/controller/helper.go
```diff
@@ -25,6 +25,30 @@ func shouldIgnoreReconciliation(obj client.Object) bool {
 	return val == metalv1alpha1.OperationAnnotationIgnore
 }
 
+// shouldChildIgnoreReconciliation checks if the object Child should ignore reconciliation.
+// if Parent has OperationAnnotation set to ignore-child, Child should also ignore reconciliation.
+func shouldChildIgnoreReconciliation(ParentObj client.Object) bool {
```
Nuckal777, line 30: `parentObj` should start lowercase.

internal/controller/serverclaim_controller.go
```diff
@@ -305,131 +305,110 @@ func (r *ServerClaimReconciler) claimServer(ctx context.Context, log logr.Logger
...
+	var selectedServer *metalv1alpha1.Server
+	var prevSelectedServer *metalv1alpha1.Server
+	for _, server := range serverList.Items {
+		// find previously claimed server
+		if ref := server.Spec.ServerClaimRef; ref != nil {
+			if ref.UID == claim.UID && ref.Name == claim.Name && ref.Namespace == claim.Namespace {
+				prevSelectedServer = &server
+				break
+			}
+		}
```
Nuckal777, line 318: I am open for this refactoring. I'd like to keep the shortcut for already claimed servers from the previous code though:
```go
if server := checkForPrevUsedServer(log, serverList.Items, claim); server != nil {
		return server, nil
}
```

internal/controller/serverclaim_controller.go
```diff
@@ -305,131 +305,110 @@ func (r *ServerClaimReconciler) claimServer(ctx context.Context, log logr.Logger
...
 	if server.Status.PowerState != metalv1alpha1.ServerOffPowerState {
 		log.V(1).Info("Server is not powered off", "Server", server.Name, "PowerState", server.Status.PowerState)
-		return nil, nil
+		return false
 	}
-	if claim.Spec.ServerSelector == nil {
-		return server, nil
+	isUnderMaintenance, err := r.isUnderMaintenanceQueue(ctx, log, server)
+	// is undergoing maintenance and not in Reserved State, we should not claim this server
+	if server.Status.State != metalv1alpha1.ServerStateReserved && (err != nil || isUnderMaintenance) {
```
Nuckal777, line 407: `server.Status.State` is always `Available` here, due to the check `server.Status.State != metalv1alpha1.ServerStateAvailable` above.

Nuckal777 (changes requested):
Thanks for taking another look. :+1:

If I am not mistaken some set-controllers also don't remove the `ignore` annotation from childs, if the `ignore-childs` annotation is removed.

Now there is dedicated `ignore-childs` operation, which allows the annotation handling to be moved into the regular reconciliation flow (create and patch methods) of all set-controllers. That would be quite a simple if statement in these methods. What do you think?

internal/controller/serverclaim_controller.go
nagadeesh-nagaraja: just to understand, what is the advantage? is it readability? 

However, having the separate method means you will loop twice. 
here we achieve the same with single loop of resource list.

internal/controller/biossettingsset_controller.go
nagadeesh-nagaraja: No, at this point in code, if the Child has Ignore annotation added through Set, we would like to delete it ( because the Set object does not have the annotation anymore)

Issue #419, opened 2025-08-06
Add BMCSettingSet type and controller


nagadeesh-nagaraja:
Create a Set for BMCSettings to handle multiple BMC to set bmc settings at a time.

Example CRD Spec 

```yaml
apiVersion: metal.ironcore.dev/v1alpha1
kind: BMCSettingsSet
metadata:
  labels:
  name: bmc-setting-set-sample
spec:
  BMCSelector:
    matchLabels:
       Manufacturer: a1
       Platform: foo
  bmcSettingsTemplate:
    version: 2.10.3 # version to apply settings to
    settings: 
          foo: bar 
          foo2: bar2
    ServerMaintenancePolicy: OwnerApproval  # Policy an defined in ServerMaintenance resource
status:
    FullyLabeledBMCs: 10  # is the number of BMC selected in the set.
    AvailableBMCSettings: 10 # number of BMCSettings we current have created
    PendingBMCSettings: 5 # is the number of BMCSettings update is Pending
    InProgressBMCSettings: 5 # is the number of BMCSettings update is InProgress
    FailedBMCSettings: 2 # is the number BMCSettings update has FAILED
    CompletedBMCSettings: 3 # is the number of BMCSetting update has Completed
```

workflow: 
- Given the selector field, all BMC matching that label will be selected to apply the specified list of settings. 
- Will create BMCSettings resource for each selected BMC. 
- The controller will monitor the changes to BMC (delete/create/ updates to label) and respond to create/delete the BMCSettings resources. 
- The Controller will list the BMC selected. BMCSettings created, and their current state as (Pending/InProgress/Failed/Completed)

davidgrun:
added new logic for MaintenanceRef handling

PR #471, opened 2025-09-16
feat: Implement BMCSettingsSet controller with reconciliation logic

davidgrun:
Added BMCSettingsSetReconciler to manage BMCSettingsSet resources.

Issue: [419](https://github.com/ironcore-dev/metal-operator/issues/419)


internal/controller/bmcsettingsset_controller.go
```diff
@@ -0,0 +1,420 @@
...
+		}
+		opResult, err := controllerutil.CreateOrPatch(ctx, r.Client, &bmcSettings, func() error {
+
+			if bmcSettingsTemplate.ServerMaintenanceRefs != nil {
+				bmcSettings.Spec.BMCSettingsTemplate = *bmcSettingsTemplate.DeepCopy()
+			} else {
+				serverMaintenanceRefs := bmcSettings.Spec.ServerMaintenanceRefs
+				bmcSettings.Spec.BMCSettingsTemplate = *bmcSettingsTemplate.DeepCopy()
+				bmcSettings.Spec.ServerMaintenanceRefs = serverMaintenanceRefs
+			}
```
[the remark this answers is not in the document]
davidgrun: good point will generate a function for this

api/v1alpha1/bmcsettings_types.go
```diff
@@ -35,6 +30,17 @@ type BMCSettingsSpec struct {
 	ServerMaintenanceRefs []ServerMaintenanceRefItem `json:"serverMaintenanceRefs,omitempty"`
 }
 
+// BMCSettingsSpec defines the desired state of BMCSettings.
+
+type BMCSettingsSpec struct {
+	BMCSettingsTemplate `json:",inline"`
+
+	// BMCRef is a reference to a specific BMC to apply setting to.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="serverRef is immutable"
```
nagadeesh-nagaraja: can we change this to message="BMCRef is immutable"  
also make it required please

internal/controller/bmcsettingsset_controller.go
```diff
@@ -0,0 +1,563 @@
...
+	if err != nil {
+		log.Error(err, "Failed to list owned BMCSettings")
+		return ctrl.Result{}, fmt.Errorf("failed to get owned BMCSettings resources %w", err)
+	}
+	delTableBMCSettings := map[string]struct{}{}
+	for _, bmcSettings := range ownedBMCSettings.Items {
+
+		if bmcSettings.Status.State != metalv1alpha1.BMCSettingsStateInProgress {
+			delTableBMCSettings[bmcSettings.Name] = struct{}{}
+		} else if len(bmcSettings.Spec.ServerMaintenanceRefs) == 0 {
```
nagadeesh-nagaraja: improvement for next PR (and not all other resources which have similar logic) we should also consider the cases where we are just waiting for serverMaintenance to be approved. these can also be deleted.

internal/controller/biossettings_controller_test.go
```diff
@@ -21,6 +21,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// added const field for go linting
+const (
+	testBarValue = "bar"
+)
+
```
Nuckal777: Probably, we should not run the `goconst` (?) linter on the test files, instead of introducing an inconsistently used constant.

internal/controller/bmcsettingsset_controller.go
```diff
@@ -0,0 +1,562 @@
...
+			// If no ServerMaintenanceRefs is set, the BMCSettings is not actively being processed
+			delTableBMCSettings[bmcSettings.Name] = struct{}{}
+		}
```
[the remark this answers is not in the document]
davidgrun: good point , adjusted

PR #450, opened 2025-08-26
Add BMC reset functionality and new annotation for forced reset

stefanhipfel:
# Proposed Changes

- Introduce new force BMC reset operation annotation: `metal.ironcore.dev/operation: ForceReset`.
- Introduce new flag `--bmc-failure-reset-delay`. If set, the force reset on a BMC will be triggered.

PR #475, opened 2025-09-26
Bmc redfish reset

stefanhipfel:
due to workerless shoot, the operator cannot create jobs for long running tasks like ssh resets. Therefor only redfish reset is possible.


Nuckal777 (commented):
Thanks. 👍  Just some chore left. (I already reviewed the code in #450).

PR #477, opened 2025-09-29
Normalize BMCOptions in BMC controller

fwiesel:
All other controllers use BasicAuth, and call the options just BMCOptions, so let's normalise on that.
This reduces the variance in behaviour between the controllers and the requirements on potential mocking servers.

# Proposed Changes

-
-
-

Fixes #


fwiesel:
I wanted to try to get the metal-controller to run against [Virtual Redfish BMC](https://docs.openstack.org/sushy-tools/latest/user/dynamic-emulator.html), and was running into that issue, because it doesn't implement the session protocol.

It would be possible of course to implement sessions there, but I am wondering if it isn't of advantage for the metal controller as well to use just one kind of authentication method.

What do you say?

fwiesel:
I've rebased it, but the test is failing "randomly" in other controllers:
- `BIOSVersionSet`
- `BIOSSettings Sequence`
- `Endpoints`

Noticeably, not the `BMCReconciler` I've changed. "Randomly", as I see similar but not the same errors locally.

PR #480, opened 2025-10-01
Handle multiple mocked server in unit testing

nagadeesh-nagaraja:
# Proposed Changes
 handle multiple mocked server in unit testing. by creating mock data for each mocked server
the distinction is based on bmc user for the mocked server in unit test

### DIFF
--- a/bmc/mockup.go
+++ b/bmc/mockup.go
@@ -13,41 +13,51 @@ import (
 type RedfishMockUps struct {
 	mu sync.RWMutex
 
-	BIOSSettingAttr       map[string]map[string]any
-	PendingBIOSSetting    map[string]map[string]any
-	BIOSVersion           string
-	BIOSUpgradingVersion  string
-	BIOSUpgradeTaskIndex  int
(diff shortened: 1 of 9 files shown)


afritzler (approved):

bmc/mockup.go
```diff
@@ -13,41 +13,51 @@ import (
 type RedfishMockUps struct {
 	mu sync.RWMutex
 
-	BIOSSettingAttr       map[string]map[string]any
-	PendingBIOSSetting    map[string]map[string]any
-	BIOSVersion           string
-	BIOSUpgradingVersion  string
-	BIOSUpgradeTaskIndex  int
+	BIOSSettingAttr       map[string]map[string]map[string]any
+	PendingBIOSSetting    map[string]map[string]map[string]any
```
Nuckal777, line 17: Could we do something about the typing here? E.g.
```go
type RedfishMockUps {
    mu sync.RWMutex
    servers map[string]ServerMockUp
}
type ServerMockUp {
    BIOSSettingAttr map[string]map[string]any
    ...
}
```

PR #483, opened 2025-10-06
read SEL from bmc to set server health condition

stefanhipfel:
This change reads the critical SEL (system event logs) from a BMC to update the server health condition.

open points:
- what if there are multiple critical logs. Condition can only be set to false once
- should the logs be cleared when adding a new server to metal-api?


afritzler:
We should close this PR in favor of https://github.com/ironcore-dev/metal-operator/pull/494

Issue #469, opened 2025-09-09
Find a generic Solution to Maintenance tasks causing server (node) in and out of "Ready" state due to reboots


nagadeesh-nagaraja:
# Summary of Issue

when the servers are rebooted during Maintenance, the system loads from Disk and kubelet will get the system into "Ready state. when the maintenance involves several reboot or longer maintenance cycle, this can cause the server to be added to the cluster even during the Maintenance. 

## Motivation

Given that multiple kind of `serverMaintenances` can lead to `server` reboots. We would need a generic solution to discard the server which are under Maintenance to be out of cluster irrespective of if its rebooted or not. 

Example case:
A server being rebooting due to power supply/Motherboard/memory issue etc.  Simply transferring it to Maintenance should let the system not be added back to cluster (or READY state) when it reboots again. 

Example case:
The Claimed `Server` is approved by Owner for Maintenance. However, It might be a good approach to make sure that the server is not getting back into Ready state during Maintenance State, even if whatever operations might lead to a reboot. instead of expecting every Maintenance operator to take care of it individually.

## Proposed Solution
Solution 1) 
When `Server` is transferred to Maintenance state, we somehow make the kubelet not start. may be having a hook which will call before starting the startup script. 
Cons: this means, we will need to update the OS and start-up script provided by the upstream projects. 

Solution 2) 
`ServerMaintanence` resource will Change the BootOrder (or disable Server's Network, or both) to stop the server booting from disk and kubelet to report itself to be ready. 
it will then patch the previously existing bootOrder (enable network, or both) to enable the server  reboot into disk and kubelet communication. 

Pros: This will give a generic way to solve the issue of server in Maintenance being reported being in "READY" state to upstream projects

atd9876:
Detailed solution proposal:
https://github.com/ironcore-dev/metal-operator/blob/feature/maintenance-boot-override/docs/proposals/maintenance-boot-override.md

PR #488, opened 2025-10-08
Reboot skip disk

### LINKED ISSUE #472: Server Resource Controller setBoot Order is not working as expected
**Describe the bug**
the BootOrder spec for Server Resource is creating a job to update the boot order for server (at least on dell) and is not actually setting the boot order requested. In the next reconcile the Server Controller fails because the job is still pending and the Redfish API Call to set the boot order will fail. 

**To Reproduce**
create server with BootOrder Spec details

**Expected behavior**
set bootOrder correctly. without just scheduling the job for it.

nagadeesh-nagaraja:
Fixes #472 #469


internal/controller/biossettings_controller.go
```diff
@@ -1143,30 +1149,42 @@ func (r *BiosSettingsReconciler) checkForRequiredPowerStatus(
 }
 
 func (r *BiosSettingsReconciler) checkIfMaintenanceGranted(
+	ctx context.Context,
 	log logr.Logger,
 	biosSettings *metalv1alpha1.BIOSSettings,
 	server *metalv1alpha1.Server,
-) bool {
+) (bool, metalv1alpha1.ServerMaintenanceState) {
```
Nuckal777: Could `bool` return value be removed and the call site simply checks for `state == InMaintenance`?

internal/controller/servermaintenance_controller.go
```diff
@@ -150,59 +167,88 @@ func (r *ServerMaintenanceReconciler) handlePendingState(ctx context.Context, lo
...
+		// note: if the server is already off, this will be a no-op
+		if server.Spec.Power != metalv1alpha1.PowerOff && server.Status.PowerState != metalv1alpha1.ServerOffPowerState {
+			if err := r.setAndPatchServerPowerState(ctx, log, server, metalv1alpha1.PowerOff); err != nil {
+				return ctrl.Result{}, err
+			}
+			// requeue to give time to the server to power off
+			return ctrl.Result{RequeueAfter: r.ResyncInterval}, nil
+		}
+		// if no config found or if it has changed, we need to create or path it
+		config, err = r.applyServerBootConfiguration(ctx, log, serverMaintenance, server)
```
Nuckal777: Does this PR cover cover the use case to change bios settings, which don't need a reboot? In that case the boot config dance isn't strictly needed.

internal/controller/servermaintenance_controller.go
```diff
@@ -285,9 +335,27 @@ func (r *ServerMaintenanceReconciler) delete(ctx context.Context, log logr.Logge
...
+	if requeue, err := r.revertMaintenanceBootConfig(ctx, log, server, serverMaintenance); err != nil {
+		return ctrl.Result{}, err
+	} else if requeue {
+		log.V(1).Info("Wait for boot order to be reverted on server", "Server", server.Name)
+		return ctrl.Result{RequeueAfter: r.ResyncInterval}, nil
+	}
+	log.V(1).Info("Boot order reverted on server", "Server", server.Name)
+	// make sure the server is powered on before removing maintenance
+	// this will ensure the server is booted into default boot order
+	if server.Spec.Power != metalv1alpha1.PowerOn && server.Status.PowerState != metalv1alpha1.ServerOnPowerState {
```
Nuckal777: Should the metal-operator really boot once into the original boot order? What if a user deliberately powered off a `Server` via a `ServerClaim`?

internal/controller/servermaintenance_controller.go
```diff
@@ -348,22 +422,295 @@ func (r *ServerMaintenanceReconciler) cleanup(ctx context.Context, log logr.Logg
...
+			return false
+		}
+
+		var isPxeBootOption = func(bootOption *redfish.BootOption) bool {
+			return strings.Contains(strings.ToLower(bootOption.DisplayName), "pxe")
+		}
+
+		bootOrderChanged := rearrangeBootOrder(bootOrder, bootOptions, isPxeBootOption, isDiskBootOption)
+
+		if reflect.DeepEqual(bootOrder, bootOrderChanged) {
```
Nuckal777: Consider `slices.Equal`.

internal/controller/servermaintenance_controller.go
```diff
@@ -348,22 +422,295 @@ func (r *ServerMaintenanceReconciler) cleanup(ctx context.Context, log logr.Logg
...
+			log.V(1).Info("Configured boot order for server",
+				"Server", server.Name,
+				"BootType", maintenance.Spec.ServerBootConfigurationTemplate.BootType,
+				"BootOrder", bootOrderChanged)
+			// save the current boot order to revert back after maintenance
+			status := &metalv1alpha1.ServerMaintenanceBootOrder{
+				DefaultBootOrder: bootOrder,
+				State:            metalv1alpha1.BootOrderConfigInProgress,
+			}
+			if err := r.patchBootOrderStatus(ctx, log, maintenance, status); err != nil {
```
Nuckal777: Should `patchBootOrderStatus` be called before`SetBootOrder` in case patching the status fails? `SetBootOrder` would be called multiple times then.

internal/controller/servermaintenance_controller.go
```diff
@@ -348,22 +422,295 @@ func (r *ServerMaintenanceReconciler) cleanup(ctx context.Context, log logr.Logg
...
+					}
+					if bootDevice != bootOrder[idx] {
+						log.V(1).Info("boot order for server is not yet reverted to default. Waiting...",
+							"Server", server.Name,
+							"CurrentBootOrder", bootOrder,
+							"DefaultBootOrder", serverMaintenance.Status.BootOrderStatus.DefaultBootOrder)
+						return true, nil
+					}
+				}
+				revertCompleted = true
```
Nuckal777: I think this check doesn't work, if a boot order somewhere in the middle is removed. If we stored in `DefaultBootOrder` `[a,b,c]` and b is removed `b != c` will always be true for `idx = 1`.

internal/controller/bmcsettings_controller.go
```diff
@@ -495,26 +499,41 @@ func (r *BMCSettingsReconciler) checkIfMaintenanceGranted(
 	log logr.Logger,
 	bmcSetting *metalv1alpha1.BMCSettings,
 	bmcClient bmc.BMC,
-) bool {
+) (bool, metalv1alpha1.ServerMaintenanceState) {
```
Nuckal777: It seems this function returns one of three states (`Unknown`, `SomeMaintenanceMissing`, `AllInMaintenance`). Could these be returned as a single instance of an enum?

internal/controller/servermaintenance_controller.go
```diff
@@ -150,59 +167,88 @@ func (r *ServerMaintenanceReconciler) handlePendingState(ctx context.Context, lo
...
-		return ctrl.Result{}, nil
 	}
-	if config.Status.State == metalv1alpha1.ServerBootConfigurationStatePending || config.Status.State == "" {
-		log.V(1).Info("Server boot configuration is pending", "Server", server.Name)
-		return ctrl.Result{}, nil
+
+	// Note: this is to check skip of BootConfiguration
+	// implication of skipping is that the server will still continue to run the OS during Maintenance
+	// might be subjected to reboots
+	val, found := serverMaintenance.GetAnnotations()[metalv1alpha1.OperationSkipBootConfiguration]
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: from what I have seen until now. All of the operations related to maintenance needs server reboot. Idea is to force servermaintenance creator to provide the bootCondifguration spec.
if this is not needed for some extra ordinary conditions, we can skip providing the spec using annotation.

internal/controller/servermaintenance_controller.go
nagadeesh-nagaraja: ah, I see the problem now. if I power on the server with fixed boot order, we might end up booting into disk and showing up online during Maintenance state. 
But according to me, I do not see a way around this, because when the server setup the boot Order. it might actually end up booting into it after it was configured.. 


In my opinion, this should not be a problem if the server is anyway is being returned. if more maintenance are present, we should actually not do this reorder of boot every time. But I guess this improvement can be taken into consideration when we actually solve it using the SPEC definitions.

afritzler:
Closing and moving the discussion to: https://github.com/ironcore-dev/metal-operator/issues/807

PR #492, opened 2025-10-09
Extend `NetworkInterface` status with LLDP neighbor information 

xkonni:
# Proposed Changes

This PR allows the metal-api to hold more information about the server. It is based on #452 and the first commit is actually from that PR, as it was not yet merged.

- update crd 
  - add all networkInterfaces, include ipv4 + ipv6 ips and status
  - add lldp information
- update registry
  - lldp was added in #452 
  - update networkInterface with ipv6 addresses and carrierStatus
- update controller to feed the newly available data in the registry into the crd


internal/controller/server_controller.go
```diff
@@ -825,14 +825,61 @@ func (r *ServerReconciler) extractServerDetailsFromRegistry(ctx context.Context,
...
+		}
+
+		// Process all IP addresses from the single IpAddresses slice
+		var allIPs []metalv1alpha1.IP
+		for _, ipAddr := range s.IpAddresses {
+			if ipAddr != "" {
+				// Parse and validate the IP address
+				ip, err := metalv1alpha1.ParseIP(ipAddr)
+				if err != nil {
+					log.V(1).Info("Invalid IP address, skipping", "interface", s.Name, "ip", ipAddr, "error", err)
```
stefanhipfel: better to log the error as error :)
log.V(1).Error

internal/controller/server_controller.go
```diff
@@ -825,14 +825,61 @@ func (r *ServerReconciler) extractServerDetailsFromRegistry(ctx context.Context,
...
+				}
+				nics[i].Neighbors = neighbors
+				break
+			}
+		}
 	}
+
 	server.Status.NetworkInterfaces = nics
 
+	log.V(1).Info("Found server information in registry", "nics", nics)
```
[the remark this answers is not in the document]
xkonni: that is a debugging leftover. removed entirely.

nagadeesh-nagaraja (commented):
https://github.com/ironcore-dev/metal-operator/pull/452 is still part of this PR, I though the plan was to merge it separately?

afritzler:
@xkonni can you please rebase this PR?

api/v1alpha1/server_types.go
```diff
@@ -290,16 +290,44 @@ type NetworkInterface struct {
...
 
-	// IP is the IP address assigned to the network interface.
-	// The type is specified as string and is schemaless.
-	// +kubebuilder:validation:Type=string
-	// +kubebuilder:validation:Schemaless
-	// +required
-	IP IP `json:"ip"`
+	// IPs is a list of IP addresses (both IPv4 and IPv6) assigned to the network interface.
+	// +optional
+	IPs []IP `json:"ips,omitempty"`
```
afritzler: This will be a breaking change. How about keeping the `ip` field and let the controller `nil`ify it for now and store the actual IP in the `ips` slice? We should also deprecate this `ip` field and remove it in a later version. @Nuckal777 @defo89 are you guys currently using rely on this field?

xkonni: we decided this is fine.

hardikdr: As discussed, agree with @afritzler's suggestion. Lets keep the ip field, and set it to nil for now and deprecate it later, and we will remove the dependency from the boot operator.

Nuckal777 (approved):

PR #495, opened 2025-10-13
Propagate retry operation from `Set` resources to child resources

### LINKED ISSUE #473: Annotation to retry all the Failed child Resources
BIOS/BMC Setting/Version Set Resources create child object which might end up in failed state. 

Provide as ability to create a re-try annotation which will be propogated to the child object and restart the failed child resources.

nagadeesh-nagaraja:
Fixes #473

### DIFF
--- a/api/v1alpha1/constants.go
+++ b/api/v1alpha1/constants.go
@@ -62,8 +62,8 @@ const (
 	// ForceOnServerPower indicates to force powerOn the baremetal server power.
 	ForceOnServerPower = "force-on-server"
 
-	// GracefulShutdownBMCPower indicates to gracefully restart the baremetal server's BMC's power.
-	GracefulRestartBMCPower = "graceful-restart-bmc"
+	// GracefulShutdownBMC indicates to gracefully restart the baremetal server's BMC's power.
+	GracefulRestartBMC = "graceful-restart-bmc"
 )
 
 var AnnotationToRedfishMapping = map[string]redfish.ResetType{
@@ -72,5 +72,5 @@ var AnnotationToRedfishMapping = map[string]redfish.ResetType{
 	PowerCycleServerPower:      redfish.PowerCycleResetType,
 	ForceOffServerPower:        redfish.ForceOffResetType,
 	ForceOnServerPower:         redfish.ForceOnResetType,
-	GracefulRestartBMCPower:    redfish.GracefulRestartResetType,
+	GracefulRestartBMC:         redfish.GracefulRestartResetType,
 }
--- a/internal/controller/bmc_controller_test.go
+++ b/internal/controller/bmc_controller_test.go
@@ -478,7 +478,7 @@ var _ = Describe("BMC Reset", func() {
 		By("Resetting the BMC")
 		Eventually(Update(bmc, func() {
 			bmc.Annotations = map[string]string{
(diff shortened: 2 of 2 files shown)


internal/controller/biossettingsset_controller.go
```diff
@@ -162,8 +167,68 @@ func (r *BIOSSettingsSetReconciler) handleIgnoreAnnotationPropagation(
...
+					annotations := biosSettings.GetAnnotations()
+					log.V(1).Info("Retry operation deleted on child object", "BIOSSettings", biosSettings.Name)
+					delete(annotations, metalv1alpha1.OperationAnnotation)
+					delete(annotations, metalv1alpha1.OperationAnnotationPropagated)
+					biosSettings.SetAnnotations(annotations)
+				}
+				return nil
+			})
+			if err != nil {
+				errs = append(errs, fmt.Errorf("failed to patch BIOSSettings annotations Propogartion removal: %w for: %v", err, biosSettings.Name))
```
Nuckal777: typo in `propagation`

internal/controller/biossettingsset_controller.go
```diff
@@ -100,6 +100,11 @@ func (r *BIOSSettingsSetReconciler) delete(
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to update current BIOSSettingsSet Status %w", err)
 		}
+		// handle propagartion of retry annotation to child when parent is being deleted
```
Nuckal777: typo in propagation

Nuckal777 (approved):
PR is fine from my side. @afritzler maybe you want to take another look?

api/v1alpha1/constants.go
```diff
@@ -3,27 +3,75 @@
...
+	GracefulRestartServerPower: redfish.OnResetType,
+	HardRestartServerPower:     redfish.ForceRestartResetType,
+	PowerCycleServerPower:      redfish.PowerCycleResetType,
+	ForceOffServerPower:        redfish.ForceOffResetType,
+	ForceOnServerPower:         redfish.ForceOnResetType,
+}
+
+const (
+	// ForceResetBMCOperationAnnotation forces a reset of BMC before next operation
+	ForceResetBMCOperationAnnotation = "force-reset-BMC"
```
afritzler: Do we need an own force reset operation for BMCs? Why not just having a global `force-reset` operation for all applicable resources?

afritzler (changes requested):
Looks good from my side - some minor things. Just as a side node, why do decided to go with a dedicated `ForceReset` annotation value for BMCs? I can't remember anymore.

api/v1alpha1/constants.go
nagadeesh-nagaraja: I am guessing that, this was named different (and was not using the regular reset operation similar to server) because initial idea was this does both, tries regular reset of server, if not then force reset. this was suppose to be one operation doing everything? 
but while we figure out the kubeConfig seperation to run jobs, this was probably left behind... 

Should I change it back to this? 
`OperationAnnotationForceReset = "ForceReset"`

currently, this is only handled in BMC controller which is doing graceful reset ( server resource uses its own operation)

afritzler (changes requested):

Issue #509, opened 2025-10-27
metalprobe docker container needs networkctl


xkonni:
**Describe the bug**

Since #452 was merged, the `metalprobe` cmd depends on `networkctl` to be available. This is not the case in the current `metalprobe` docker container

**To Reproduce**

```
❯ make docker-build-metalprobe
...


❯ docker run --rm -it metalprobe:latest --server-uuid=123 --registry-url=123
...
2025-10-27T12:44:26Z    ERROR   setup   failed to initialize agent      {"error": "networkctl is not present: exec: \"networkctl\": executable file not found in $PATH"}
```

**Expected behavior**

The docker image can execute `networkctl` and collect all information required.

PR #512, opened 2025-10-28
Switch from `networkctl` to `lldpd` in `metalprobe`

xkonni:
As the current metalprobe needs a full systemd with networkctl running and we can not (properly) provide this within a docker, this is a proposed switch from systemd-networkctl to lldpd-lldpctl.

# Proposed Changes

- use `lldpctl -f json` instead of `networkct --json=short`
- parse `lldpctl -f json` data
- add tests for `lldpctl -f json` data
- re-run the lldp discovery in every iteration
- update metalprobe docker to start lldpd before starting metalprobe
  - no longer using distroless nonroot, discovery needs root privileges
  - add launch script
  - exec `/metalprobe $args`

# Caveats

The output of `lldpctl` is way more complex than what `networkctl lldp` does, therefore the parsing is more complex as well. Tried to keep the resulting structure simple.

For reference: 
[networkctl.json](https://github.com/user-attachments/files/23191131/networkctl.json)
[lldpctl.json](https://github.com/user-attachments/files/23191136/lldpctl.json)
 
# Fixes

 - #509 docker container started via ignition on gardenlinux live **should** be happy now. I have no means to test this as a whole, but all individual steps work
 - lldp data was just fetched once and may not have been available after fresh container start. now it get's pulled with each iteration


internal/api/registry/lldp_test.go
```diff
@@ -0,0 +1,138 @@
...
+	if n.ChassisID != "aa:bb:cc:dd:ee:ff" {
+		t.Errorf("unexpected chassis id in partial fixture: %s", n.ChassisID)
+	}
+	// missing fields should be empty strings
+	if n.PortID != "" {
+		t.Errorf("expected empty port id in partial fixture, got %s", n.PortID)
+	}
+}
+
+func validateSingle(t *testing.T, parsed LLDP) {
```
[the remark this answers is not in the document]
xkonni: it is testing against lldpctl output json files from test/data. If those are changed, the test needs to changed, but it does not depend on the runtime environment.

internal/probe/probe.go
```diff
@@ -118,15 +121,45 @@ func (a *Agent) Start(ctx context.Context) error {
...
+	if a == nil || a.Server == nil {
+		return nil
+	}
+	lldp, err := collectLLDPInfo(ctx, a.LLDPSyncInterval, a.LLDPSyncDuration)
+	if err != nil {
+		a.log.Error(err, "collectLLDPInfo failed")
+		return err
+	}
+	a.Server.LLDP = lldp.Interfaces
+	a.log.Info("refreshed LLDP info", "interfaces", len(a.Server.LLDP))
```
afritzler:
```suggestion
	a.log.Info("Refreshed LLDP info", "interfaces", len(a.Server.LLDP))
```

internal/probe/probe.go
```diff
@@ -55,8 +55,10 @@ func (a *Agent) Init(ctx context.Context) error {
 
 	LLDPInfo, err := collectLLDPInfo(ctx, a.LLDPSyncInterval, a.LLDPSyncDuration)
 	if err != nil {
+		a.log.Error(err, "failed to collect LLDP info")
 		return err
 	}
+	a.log.Info("collected LLDP info", "interfaces", len(LLDPInfo.Interfaces))
```
afritzler:
```suggestion
	a.log.Info("Collected LLDP info", "interfaces", len(LLDPInfo.Interfaces))
```

afritzler (changes requested):

PR #515, opened 2025-10-31
Remove `ServerMaintainenceRefs` from Spec in `Set` CRDs

nagadeesh-nagaraja:
# Proposed Changes
Removes serverMaintenanceRef from Spec in the Set CRD.

### DIFF
--- a/internal/controller/bmcversionset_controller.go
+++ b/internal/controller/bmcversionset_controller.go
@@ -179,7 +179,7 @@ func (r *BMCVersionSetReconciler) reconcile(
 		return ctrl.Result{}, fmt.Errorf("failed to delete orphaned BMCVersion resources %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


internal/controller/bmcversionset_controller.go
```diff
@@ -266,29 +265,27 @@ func (r *BMCVersionSetReconciler) deleteOrphanBMCVersions(
...
 	bmcVersionList *metalv1alpha1.BMCVersionList,
 ) error {
 	if len(bmcVersionList.Items) == 0 {
 		log.V(1).Info("No BMCVersion found, skipping spec template update")
 		return nil
 	}
+	bmcNameMap := make(map[string]metalv1alpha1.BMC)
+	for _, bmc := range bmcList.Items {
+		bmcNameMap[bmc.Name] = bmc
+	}
```
Nuckal777: Appears to be unused currently (might be some side effect of another cleanup).

PR #516, opened 2025-11-03
De-normalize tests and ensure clean state before each `It()` run

afritzler:
# Proposed Changes

- Use `RunnableFunc` to start registry in `manager`
- Denormalize tests by ensuring each `It()` block leaves the API server clean

In future we should run `EnsureCleanState` after each `It()` block for new tests to ensure that we have a clean state to start the tests. The problem in our tests are `ClusterScoped` objects which might cause side effects when left around.

For that reason we are having now a:

```
AfterEach(func(ctx SpecContext) {
        ...
	EnsureCleanState()
})
```

Additionally: We should postpone the introduction of complex helper methods for now even when that means that we have code duplication in our tests.

### DIFF
--- a/internal/controller/bmcversionset_controller_test.go
+++ b/internal/controller/bmcversionset_controller_test.go
@@ -124,6 +124,48 @@ var _ = Describe("BMCVersionSet Controller", func() {
 	AfterEach(func(ctx SpecContext) {
 		bmc.UnitTestMockUps.ResetBMCVersionUpdate()
 
+		maintenanceList := &metalv1alpha1.ServerMaintenanceList{}
+		Eventually(List(maintenanceList)).Should(Succeed())
+		for _, maintenance := range maintenanceList.Items {
+			Expect(k8sClient.Delete(ctx, &maintenance)).To(Succeed())
+		}
+
+		server01 := &metalv1alpha1.Server{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: bmcutils.GetServerNameFromBMCandIndex(0, bmc01),
+			},
+		}
+		Eventually(UpdateStatus(server01, func() {
+			server01.Status.State = metalv1alpha1.ServerStateAvailable
+		})).Should(Succeed())
+		Expect(k8sClient.Delete(ctx, bmc01)).To(Succeed())
+		Expect(k8sClient.Delete(ctx, server01)).To(Succeed())
+		Eventually(Get(server01)).Should(Satisfy(apierrors.IsNotFound))
+
+		server02 := &metalv1alpha1.Server{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: bmcutils.GetServerNameFromBMCandIndex(0, bmc02),
+			},
+		}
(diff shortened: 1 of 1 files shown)


internal/controller/biossettings_controller_test.go
```diff
@@ -68,15 +63,23 @@ var _ = Describe("BIOSSettings Controller", func() {
 				},
 			},
 		}
-		TransitionServerFromInitialToAvailableState(ctx, k8sClient, server, ns.Name)
+		Expect(k8sClient.Create(ctx, server)).To(Succeed())
+
+		By("Ensuring that the Server is in available state")
+		Eventually(UpdateStatus(server, func() {
+			server.Status.State = metalv1alpha1.ServerStateAvailable
+		})).To(Succeed())
```
nagadeesh-nagaraja: are we going back to patching the state of server and not allowing it to transition to Available? 

I have seen that this might cause flakiness in server state later. and cause the server Status to not be populated..

afritzler: I would do the transitioning via `agent` and `registry` setup in the `BMC` and `Server` reconciler as this is their core responsibility.  Repeating this test in other controllers seems arbitrary. Ideally the `BIOS/BMCSettings/Version` controller only rely on the contract that a `Server` has to be available.

internal/controller/biossettings_controller_test.go
```diff
@@ -68,15 +63,23 @@ var _ = Describe("BIOSSettings Controller", func() {
...
 	})
 
 	AfterEach(func(ctx SpecContext) {
-		DeleteAllMetalResources(ctx, ns.Name)
 		bmc.UnitTestMockUps.ResetBIOSSettings()
+
+		Expect(k8sClient.Delete(ctx, bmcSecret)).To(Succeed())
+		Expect(k8sClient.Delete(ctx, server)).To(Succeed())
+		EnsureCleanState()
```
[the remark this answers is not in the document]
stefanhipfel: I think cleaning up after each `It` block sounds good and we would be checking `Delete` reconciles as well.

internal/controller/bmcversionset_controller_test.go
```diff
@@ -19,384 +21,488 @@ import (
...
+		Expect(k8sClient.Delete(ctx, bmcVersion01)).To(Succeed())
+		Expect(k8sClient.Delete(ctx, bmcVersion02)).To(Succeed())
+
+		maintenanceList := &metalv1alpha1.ServerMaintenanceList{}
+		Eventually(List(maintenanceList)).Should(Succeed())
+		for _, maintenance := range maintenanceList.Items {
+			Expect(k8sClient.Delete(ctx, &maintenance)).To(Succeed())
+		}
+
+		server01 := &metalv1alpha1.Server{
```
Nuckal777: I would argue that the BMC and Server cleanup could be moved in the `AfterEach` block for deduplication, because the BMCs are setup in the `BeforeEach` block.

Nuckal777 (approved):

PR #522, opened 2025-11-07
Add workflow to test in darwin

xkonni:
# Proposed Changes

this adds a workflow to run `make test` in a macos runner image

# fixes

potential oversights when implementing new features


Nuckal777:
Why? Only linux is practically relevant for Kubernetes. Failures related to local setups will be found during the normal dev cycle. 🤔

PR #530, opened 2025-11-14
Report error on unknown BMC protocol

maxmoehl:


Nuckal777:
Thanks, wouldn't the error be logged twice here? iirc, returning the error should cause it to be logged by `controller-runtime`.

PR #536, opened 2025-11-24
Customization of default discovery Ignition

asergeant01:
# Proposed Changes

The current default ignition is hardcoded. This change allows it to be optionally overridden by a ConfigMap.

## Summary by CodeRabbit

* **New Features**
  * Add support for customizable Ignition templates via a new configuration flag and provide a default template included in the manager image.

* **Documentation**
  * Add detailed Ignition template customization guide with Helm examples, required template variables, and troubleshooting.

* **Improvements**
  * Make probe data collection more resilient by treating certain collection errors as non-fatal.

* **Chores / Tests**
  * Harden build recipes for safer tool invocation and add unit tests covering file-based Ignition template handling.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,7 @@ FROM gcr.io/distroless/static:nonroot AS manager
 LABEL source_repository="https://github.com/ironcore-dev/metal-operator"
 WORKDIR /
 COPY --from=manager-builder /workspace/manager .
+COPY config/manager/ignition-template.yaml /etc/metal-operator/ignition-template.yaml
 USER 65532:65532
 
 ENTRYPOINT ["/manager"]
--- a/dist/chart/values.yaml
+++ b/dist/chart/values.yaml
@@ -91,3 +91,14 @@ certmanager:
 # [NETWORK POLICIES]: To enable NetworkPolicies set true
 networkPolicy:
   enable: false
+
+# [IGNITION]: Ignition template configuration
+# By default, the operator uses a template file baked into the container at /etc/metal-operator/ignition-template.yaml
+# Enable this to override the template by mounting a ConfigMap at that location
+ignition:
+  # Set to true to override the default ignition template with a custom one.
+  override: false
+  # Template content that can be customized - this will be created as a ConfigMap
+  # and mounted to override the default template
+  # template: |
+    
--- a/internal/controller/server_controller.go
+++ b/internal/controller/server_controller.go
@@ -94,6 +94,7 @@ type ServerReconciler struct {
 	DiscoveryTimeout        time.Duration
 	MaxConcurrentReconciles int
 	Conditions              *conditionutils.Accessor
+	DiscoveryIgnitionPath   string
 }
 
 // +kubebuilder:rbac:groups=metal.ironcore.dev,resources=bmcs,verbs=get;list;watch
@@ -720,14 +721,20 @@ func (r *ServerReconciler) generateDefaultIgnitionDataForServer(flags string, ss
 		return nil, fmt.Errorf("failed to generate password hash: %w", err)
 	}
 
-	ignitionData, err := ignition.GenerateDefaultIgnitionData(ignition.Config{
+	config := ignition.Config{
 		Image:        r.ProbeImage,
 		Flags:        flags,
 		SSHPublicKey: string(sshPublicKey),
 		PasswordHash: string(passwordHash),
-	})
+	}
+
+	// Load ignition template from file
+	ignitionData, err := ignition.GenerateIgnitionDataFromFile(
+		r.DiscoveryIgnitionPath,
+		config,
+	)
 	if err != nil {
-		return nil, fmt.Errorf("failed to generate default ignition data: %w", err)
+		return nil, fmt.Errorf("failed to generate ignition data from file %s: %w", r.DiscoveryIgnitionPath, err)
 	}
 
 	return ignitionData, nil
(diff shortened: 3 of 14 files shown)


cmd/manager/main.go
```diff
@@ -51,6 +55,15 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
+// getCurrentNamespace reads the namespace from the service account token
+func getCurrentNamespace() (string, error) {
+	namespaceBytes, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
nagadeesh-nagaraja: I am not sure if its a good idea to hard-code the path here? Can we guarantee the location in every scenario?

Nuckal777:
I got two questions, sorry if I am missing something obvious:
- What is the use case of using a custom ignition for the discovery boot?
- Could the custom ignition be mounted into the metal-operator pod and consumed via `os.ReadFile()`?

cmd/manager/main.go
asergeant01: Ok. I reverted it back

cmd/manager/main.go
```diff
@@ -114,12 +128,15 @@ func main() { // nolint: gocyclo
...
+	flag.StringVar(&ignitionConfigMapName, "ignition-configmap-name", "",
+		"Name of the ConfigMap containing the ignition template. If empty, uses hardcoded template.")
+	flag.StringVar(&ignitionConfigMapKey, "ignition-configmap-key", "ignition-template.yaml",
+		"Key in the ConfigMap containing the ignition template.")
 	flag.StringVar(&registryURL, "registry-url", "", "The URL of the registry.")
 	flag.StringVar(&registryProtocol, "registry-protocol", "http", "The protocol to use for the registry.")
 	flag.IntVar(&registryPort, "registry-port", 10000, "The port to use for the registry.")
 	flag.StringVar(&probeImage, "probe-image", "", "Image for the first boot probing of a Server.")
 	flag.StringVar(&probeOSImage, "probe-os-image", "", "OS image for the first boot probing of a Server.")
-	flag.StringVar(&managerNamespace, "manager-namespace", "default", "Namespace the manager is running in.")
```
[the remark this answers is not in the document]
asergeant01: I've reverted it back

internal/controller/server_controller_test.go
```diff
@@ -793,6 +854,286 @@ var _ = Describe("Server Controller", func() {
...
+  users:
+    - name: metal
+      password_hash: {{.PasswordHash}}
+      ssh_authorized_keys: [ {{.SSHPublicKey}} ]`
+
+			tmpFile, err := os.CreateTemp("", "default-ignition-*.yaml")
+			Expect(err).NotTo(HaveOccurred())
+			defer func() {
+				Expect(os.Remove(tmpFile.Name())).To(Succeed())
+			}()
```
afritzler: Isn't there a better way to do that via `ginkgo` helper?

internal/controller/server_controller_test.go
```diff
@@ -793,6 +854,286 @@ var _ = Describe("Server Controller", func() {
...
+  users:
+    - name: e2e-user
+      password_hash: {{.PasswordHash}}
+      ssh_authorized_keys: [ {{.SSHPublicKey}} ]`
+
+			tmpFile, err := os.CreateTemp("", "e2e-ignition-*.yaml")
+			Expect(err).NotTo(HaveOccurred())
+			defer func() {
+				Expect(os.Remove(tmpFile.Name())).To(Succeed())
+			}()
```
afritzler: Same here.

internal/probe/probe.go
```diff
@@ -45,39 +45,53 @@ func (a *Agent) Init(ctx context.Context) error {
 	}
 	systeminfo, err := collectSystemInfoData()
 	if err != nil {
-		return err
+		a.log.Info("System info collection skipped (DMI/SMBIOS not accessible)", "error", err.Error())
```
afritzler: I would recommend using here `log.Error`.

Dockerfile
```diff
@@ -39,6 +39,7 @@ FROM gcr.io/distroless/static:nonroot AS manager
 LABEL source_repository="https://github.com/ironcore-dev/metal-operator"
 WORKDIR /
 COPY --from=manager-builder /workspace/manager .
+COPY config/manager/ignition-template.yaml /etc/metal-operator/ignition-template.yaml
```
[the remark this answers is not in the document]
asergeant01: We could, however I think it's clearer for an end user that the discovery ignition is always loaded from a file. To me it reduces ambiguity, easier debugging and reduces code.

internal/probe/probe.go
asergeant01: Yes, agree. thank you

internal/controller/test_helper.go
```diff
@@ -1,51 +0,0 @@
-// SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and IronCore contributors
```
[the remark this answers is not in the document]
asergeant01: So this was my attempt at reducing the sporadic test failures by making the helper a test in itself so I'd have access to the k8sclient. I'm going to revert this.

dist/chart/values.yaml
```diff
@@ -89,3 +89,59 @@ certmanager:
...
+
+# [IGNITION]: Ignition template configuration
+# By default, the operator uses a template file baked into the container at /etc/metal-operator/ignition-template.yaml
+# Enable this to override the template by mounting a ConfigMap at that location
+ignition:
+  # Set to true to override the default ignition template with a custom one
+  override: false
+  # Template content that can be customized - this will be created as a ConfigMap
+  # and mounted to override the default template
+  template: |
```
Nuckal777: Should we leave template empty in the default values and `require` it in the `ConfigMap`? Otherwise, the ignition here will eventually desync with the content of default ignition baked in to the image.

internal/controller/server_controller.go
```diff
@@ -1133,6 +1145,20 @@ func (r *ServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+// validateDiscoveryIgnitionPath ensures the DiscoveryIgnitionPath is set and accessible
+func (r *ServerReconciler) validateDiscoveryIgnitionPath() error {
```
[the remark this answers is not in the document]
asergeant01: I don't see the harm in doing both?
I'm thinking of that first Server reconciliation if the file is missing or unreadable - is failing fast not preferable?

asergeant01:
@afritzler Verified helm deployment with no override

PR #571, opened 2025-12-16
Refactor `BIOSSettingsSetReconciler` and improve watch setup

afritzler:
# Proposed Changes

Refactor `BIOSSettingsSetReconciler` and improve watch setup.


internal/controller/helper.go
```diff
@@ -286,19 +280,15 @@ func handleIgnoreAnnotationPropagation(
 			log.V(1).Info("Patched Child's annotations for ignore operation", "ChildResource", childObj.GetName(), "Operation", opResult)
 		}
 		return nil
-	})
+	}); errs != nil {
+		errs = append(errs, err)
```
Nuckal777: If we use the error from `meta.EachListItem()` the `errs` slice isn't needed anymore.

Issue #596, opened 2026-01-16
Inconsistency in the state reported by the CRDs


nagadeesh-nagaraja:
**Describe the issue**
(BIOS & BMC) Version CRD reports InProgress when waiting on maintenance. while BIOS Settings CRD reports Pending state while doing the same (waiting for maintenance to be granted). make it consistent across CRDs.

Issue #593, opened 2026-01-15
BMC Version CRD seems to panic


nagadeesh-nagaraja:
**Describe the bug**
see a panic in the metal operator logs 

**Expected behavior**
no panics in the controller

**Screenshots**
```
2026-01-15T08:36:09Z	DEBUG	Child's annotations check	{"controller": "bmcversionset", "controllerGroup": "metal.ironcore.dev", "controllerKind": "BMCVersionSet", "BMCVersionSet": {"name":"dell-poweredge-r660-7-20-60-50"}, "namespace": "", "name": "dell-poweredge-r660-7-20-60-50", "reconcileID": "58d921ef-3d0e-4cda-91a2-de43a671f80e", "ChildResource": "bmc-version-set-tstt4"}
2026-01-15T08:36:21Z	DEBUG	Removing orphan server Maintenances	{"controller": "bmcversion", "controllerGroup": "metal.ironcore.dev", "controllerKind": "BMCVersion", "BMCVersion": {"name":"bmc-version-set-tstt4"}, "namespace": "", "name": "bmc-version-set-tstt4", "reconcileID": "986dc1e7-2d7c-46db-acda-32a09aef26d3"}
2026-01-15T08:36:21Z	DPANIC	odd number of arguments passed as key-value pairs for logging	{"controller": "bmcversion", "controllerGroup": "metal.ironcore.dev", "controllerKind": "BMCVersion", "BMCVersion": {"name":"bmc-version-set-tstt4"}, "namespace": "", "name": "bmc-version-set-tstt4", "reconcileID": "986dc1e7-2d7c-46db-acda-32a09aef26d3", "ignored key": "bmc-version-8kwks"}
[...]

PR #597, opened 2026-01-16
Rename `BIOSVersion` `Processing` state to `InProgress`

xkonni:
# minor fixes

- logging function was passed wrong number of arguments
- verify that the biosversion is transitioning to 'InProgress' and rename for consistency

Fixes

- #593 
- #596 

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected BIOS version in-progress state value from "Processing" to "InProgress" so status now serializes consistently.

* **Tests**
  * Added assertion to verify BIOS/BMC report the InProgress state while waiting on maintenance.

* **Documentation**
  * Updated API docs to reflect the renamed BIOSVersionState enum value (InProgress).

* **Chores**
  * Improved logging to include structured context when related maintenance records can't be fetched.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/api/v1alpha1/biosversion_types.go
+++ b/api/v1alpha1/biosversion_types.go
@@ -16,7 +16,7 @@ const (
 	// BIOSVersionStatePending specifies that the bios upgrade maintenance is waiting
 	BIOSVersionStatePending BIOSVersionState = "Pending"
 	// BIOSVersionStateInProgress specifies that upgrading bios is in progress.
-	BIOSVersionStateInProgress BIOSVersionState = "Processing"
+	BIOSVersionStateInProgress BIOSVersionState = "InProgress"
 	// BIOSVersionStateCompleted specifies that the bios upgrade maintenance has been completed.
 	BIOSVersionStateCompleted BIOSVersionState = "Completed"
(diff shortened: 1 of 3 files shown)


api/v1alpha1/biosversion_types.go
```diff
@@ -16,7 +16,7 @@ const (
 	// BIOSVersionStatePending specifies that the bios upgrade maintenance is waiting
 	BIOSVersionStatePending BIOSVersionState = "Pending"
 	// BIOSVersionStateInProgress specifies that upgrading bios is in progress.
-	BIOSVersionStateInProgress BIOSVersionState = "Processing"
+	BIOSVersionStateInProgress BIOSVersionState = "InProgress"
```
nagadeesh-nagaraja: the bug was not for this.
the issue was that the BIOS/BMCVersion shows Inprogress when its waiting for maintenance to be approved. 
while BIOSSettings and shows in Pending state. 

I think it will be better we synchronize the behavior. to be in Pending state in case of waiting for Maintenance to be approved across all CRDs

afritzler:
@xkonni can you regenerate docs and the Helm chart.

Issue #594, opened 2026-01-15
HPE ILO 6 1.70 version does not publish the BMC power state


nagadeesh-nagaraja:
**Describe the bug**
Power status of BMC is not present. hence causing issue in logic which depend on it

**To Reproduce**
run BMC version on HPE systems with ILO 6 1.70

**Expected behavior**
to vendor to publish power state of BMC

**Screenshots**
```
(qa-de-1|k=a-qa-de-100)I760727@CXRLFLKF6L:~/Workspace/PythonWorkspace%kubectl  logs metal-operator-controller-manager-6479d7d8df-9s4wd -n shoot--cp--met-op-qa-de-100 | grep bmc-version-set-9d6sm
2026-01-15T02:28:16Z	DEBUG	Child's annotations check	{"controller": "bmcversionset", "controllerGroup": "metal.ironcore.dev", "controllerKind": "BMCVersionSet", "BMCVersionSet": {"name":"hpe-proliant-dl380-gen11-ilo6-1-68"}, "namespace": "", "name": "hpe-proliant-dl380-gen11-ilo6-1-68", "reconcileID": "da037433-e798-4765-96e1-d15105e78c85", "ChildResource": "bmc-version-set-9d6sm"}
2026-01-15T02:28:20Z	DEBUG	Reconciling BMCVersion	{"controller": "bmcversion", "controllerGroup": "metal.ironcore.dev", "controllerKind": "BMCVersion", "BMCVersion": {"name":"bmc-version-set-9d6sm"}, "namespace": "", "name": "bmc-version-set-9d6sm", "reconcileID": "89fde99b-0808-4924-80b6-c32ec187227a"}
2026-01-15T02:28:25Z	DEBUG	Issuing upgrade of BMC version	{"controller": "bmcversion", "controllerGroup": "metal.ironcore.dev", "controllerKind": "BMCVersion", "BMCVersion": {"name":"bmc-version-set-9d6sm"}, "namespace": "", "name": "bmc-version-set-9d6sm", "reconcileID": "89fde99b-0808-4924-80b6-c32ec187227a"}
[...]

PR #601, opened 2026-01-20
Add `Unknown` BMC powerstate

xkonni:
# Proposed Changes

if the bmc does not report a bmc powerstate, we set the powerstate to not available and pass checks.

# Fixes
- #594 
- incident on qa-de-200 where hpe servers do not move out of maintenance, because they do not report powerstate and the update is not applied.

## Summary by CodeRabbit

* **New Features**
  * Added an "Unknown" BMC power state to represent when BMC power status is not reported.

* **Bug Fixes**
  * Improved handling for unreported/unknown BMC power states so status updates, settings reconciliation, and upgrade flows continue appropriately; clearer logging and correct condition transitions when state is unknown.

* **Documentation**
  * API reference updated to document the new "Unknown" power state.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/internal/controller/bmc_controller.go
+++ b/internal/controller/bmc_controller.go
@@ -235,12 +235,12 @@ func (r *BMCReconciler) updateBMCStatusDetails(ctx context.Context, log logr.Log
 		bmcBase := bmcObj.DeepCopy()
 		bmcObj.Status.Manufacturer = manager.Manufacturer
 		bmcObj.Status.State = metalv1alpha1.BMCState(string(manager.Status.State))
-		// Set power state, or N/A if not available from BMC
+		// Set power state, or unknown if not available from BMC
 		if manager.PowerState != "" {
 			bmcObj.Status.PowerState = metalv1alpha1.BMCPowerState(string(manager.PowerState))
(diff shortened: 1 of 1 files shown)


asergeant01 (approved):
Logic seems sound to me. Disappointed all vendors do not supply this "PowerState" response.

api/v1alpha1/bmc_types.go
```diff
@@ -163,6 +163,8 @@ const (
 	// PoweringOffPowerState A temporary state between On and Off. The power
 	// off action can take time while the OS is in the shutdown process.
 	PoweringOffPowerState BMCPowerState = "PoweringOff"
+	// NAPowerState indicates that power state is not available for this BMC.
+	NAPowerState BMCPowerState = "N/A"
```
Nuckal777: I would spell out `NotAvailable` (or even `Unknown`, which might be more Kubernetes-Style).

Nuckal777 (commented):
Thanks. Some remarks from my side regarding code hygiene. 🧹

PR #654, opened 2026-02-09
Extend the boot contract with `BootMode` and `BootMethod`

afritzler:
## Summary by CodeRabbit

* **New Features**
  * Added boot method configuration (PXE, HTTPBoot) and boot mode control (Once, Continuous) for servers.
  * Introduced command-line flags to set default boot method and boot mode globally.
  * Enhanced boot configuration flexibility with explicit boot source override parameters.

* **Documentation**
  * Updated API reference documentation with new boot configuration options and defaults.

/cc @hardikdr


afritzler:
@coderabbitai summary

internal/controller/server_controller.go
```diff
@@ -467,6 +475,9 @@ func (r *ServerReconciler) handleMaintenanceState(ctx context.Context, log logr.
 		if err := r.updateServerStatusFromSystemInfo(ctx, log, bmcClient, server); err != nil {
 			return false, fmt.Errorf("failed to update server status system info: %w", err)
 		}
+		if err := r.clearLastNetworkBoot(ctx, server); err != nil {
```
nagadeesh-nagaraja: because we are patching here, there will be a event and will reconcile...
it would be best to return reconcile if there was an update here...like using "modified" flag like we do in other methods

same in other places this method is called.

bmc/redfish.go
```diff
@@ -190,29 +180,43 @@ func (r *RedfishBMC) GetSystems(ctx context.Context) ([]Server, error) {
 	return servers, nil
 }
 
-// SetPXEBootOnce sets the boot device for the next system boot using Redfish.
-func (r *RedfishBMC) SetPXEBootOnce(ctx context.Context, systemURI string) error {
+// SetNetworkBoot configures the network boot source override on the BMC using Redfish.
```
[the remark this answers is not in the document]
maxmoehl: I agree, but maybe `SetBootOverride` or `SetBootOverrideSettings` as this maps to the redfish boot override? There's also persistent boot settings in the BIOS which this does not touch.

api/v1alpha1/serverbootconfiguration_types.go
```diff
@@ -23,6 +49,24 @@ type ServerBootConfigurationSpec struct {
...
+	BootMethod BootMethod `json:"bootMethod,omitempty"`
+
+	// BootMode controls whether the boot method override is applied once or on every boot.
+	// "Once" boots from the network one time (e.g. for OS installation), then the BIOS boot
+	// order takes over. "Continuous" boots from the network on every boot cycle.
+	// Defaults to "Once" if not specified.
+	// +kubebuilder:validation:Enum=Once;Continuous
+	// +kubebuilder:default=Once
+	// +optional
+	BootMode BootMode `json:"bootMode,omitempty"`
```
maxmoehl, line 69: Wouldn't it make more sense to mirror the Redfish behavior here? E.g.:
1. the user sets this to `Once`
2. operator applies setting to BMC
3. BMC eventually sets override back to `Disabled` after booting once
4. operator resets the spec field to `Disabled`

Not sure if the operator would rather right-away reset the field instead of waiting for the BMC, but essentially we are proxying the `BootSourceOverrideEnabled` field.

Using a different mechanism, e.g. keeping it set and tracking whether it was done already or not in the state, seems counter-intuitive to anyone who already has experience with the redfish API.

PR #660, opened 2026-02-10
Add `DNSRecord` template volume to Helm chart

stefanhipfel:
# Proposed Changes

adds a dns template volume to helm-charts. Disabled per default

### DIFF
--- a/dist/chart/templates/manager/manager.yaml
+++ b/dist/chart/templates/manager/manager.yaml
@@ -61,6 +61,11 @@ spec:
         volumeMounts:
         - mountPath: /etc/macdb/
           name: macdb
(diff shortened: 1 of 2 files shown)


asergeant01:
I'm not sure how this is going to work with the helm v2 kubebuilder plugin. We can no longer just add things into the chart and values like this I believe. Its going to have to go into the kustomisation and then the plugin automatically builds the chart.

stefanhipfel:
I know. But as long as we use the old version it just works that way. Once we move to v2 we use patching in kustomize.

afritzler:
@defo89 you mean like a `ConfigMap` stub?

PR #666, opened 2026-02-10
Set container image repository and tag in helm chart

defo89:
# Proposed Changes

- Updating image values in chart for easier chart consumption

## Summary by CodeRabbit

* **Chores**
  * Updated the Helm chart publishing workflow to automatically compute and set the container image repository and a SHA-based tag, and to update chart values accordingly.
  * Added tooling/setup steps in the workflow to support this automation.

### DIFF
--- a/.github/workflows/publish-chart.yml
+++ b/.github/workflows/publish-chart.yml
@@ -76,6 +76,16 @@ jobs:
           curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
           mv kustomize /usr/local/bin
 
+      - name: Install yq
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-chart.yml
```diff
@@ -76,6 +76,16 @@ jobs:
           curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
           mv kustomize /usr/local/bin
 
+      - name: Install yq
+        run: |
+          curl -fsSL https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -o /usr/local/bin/yq && chmod a+x /usr/local/bin/yq
+
+      - name: Set container image repository and tag
+        run: |
+          IMAGE_TAG=sha-$(echo ${{ github.sha }} | cut -c1-7)
```
asergeant01: If this is a properly tagged release, v0.2.2 etc, shouldn't we use that instead of the sha for readability?

PR #704, opened 2026-02-23
Add ServerCleaning controller

### LINKED ISSUE #613: Node Storage Cleanup
# Summary

When `Server` transitions from `Reserved` to `Available` metal-operator should support wiping the storage.

Current behaviour:
- During the rolling update, `ServerClaim` is removed and Server transitions to `Available`.
- Server was using disk install -> data on disk is persisted including possible operational data (kubelet certs, static IP address configuration).
- Server is used by another `ServerClaim` but network boot fails:
  - Server might re-join the cluster without Cluster machinery (Gardener, Cluster-API) being aware of this.
  - Server operates with previously configured IP address which is already "released" by IPAM controller.
  - Cluster operations is disrupted because of two above issues (wrong scheduling, duplicate IP addressing).

Possible options:
- Cleanup via Redfish/BMC
- Cleanup via `ServerBootConfiguration` (risk of network boot failing -> same issues as described above)

There should be 2 policies but this is open for discussion:
- Cleanup root disk only. This will help to redeploy rook/ceph clusters and persists OSDs on node redeployment.
- Multi-tenancy: all storage disks will be wiped on Server and it can be safely used by next tenant.

stefanhipfel:
# Proposed Changes
- Introduced comprehensive unit tests for the ServerCleaning controller, covering various scenarios including creation, reconciliation, and state transitions of ServerCleaning resources.
- Implemented tests to ensure proper handling of server states, cleaning tasks tracking, and finalizer management.
- Updated the test suite setup to include the ServerCleaningReconciler, ensuring it is properly initialized with the test manager.


internal/controller/server_controller.go
```diff
@@ -416,8 +419,17 @@ func (r *ServerReconciler) handleReservedState(ctx context.Context, bmcClient bm
 	// TODO: This needs be reworked later as the Server cleanup has to happen here. For now we just transition the server
 	// 		 back to available state.
 	if server.Spec.ServerClaimRef == nil {
-		if modified, err := r.patchServerState(ctx, server, metalv1alpha1.ServerStateAvailable); err != nil || modified {
-			return true, err
+		// Check if server has taints
```
stefanhipfel: can be removed as well once https://github.com/ironcore-dev/metal-operator/pull/672 is merged

afritzler:
@stefanhipfel I would suggest to close this one in favor of https://github.com/ironcore-dev/metal-operator/pull/888 and track the progress of this feature here: https://github.com/ironcore-dev/metal-operator/issues/613

PR #709, opened 2026-02-25
Fix boot property setting for BMCs using @Redfish.Settings pattern

afritzler:
# Proposed Changes

Some BMCs (e.g., AMI) require boot properties to be set via the @Redfish.Settings URI rather than directly on the system resource. Previously, SetPXEBootOnce and SetBootOrder always used SetBoot which PATCHes the system URI directly, causing a 400 error on these BMCs.

Add a setBoot helper that detects when the ComputerSystem advertises SettingsApplyTimes and routes the request through
UpdateBootAttributesApplyAt, which PATCHes the correct settings URI with OnReset apply time. For BMCs without @Redfish.Settings, the existing SetBoot behavior is preserved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unified boot configuration updates into a single apply path for more consistent behavior across systems.
  * One-time PXE boot now correctly applies and will trigger a restart when required to enforce the change.
  * Boot-order updates use the unified apply path and preserve vendor-specific constraints (e.g., UEFI).
  * Public interfaces remain unchanged while improving reliability of PXE boot and boot-order changes.

### DIFF
--- a/bmc/redfish.go
+++ b/bmc/redfish.go
@@ -212,8 +212,18 @@ func (r *RedfishBMC) SetPXEBootOnce(ctx context.Context, systemURI string) error
 	// TODO: pass logging context from caller
 	log := ctrl.LoggerFrom(ctx)
 	log.V(2).Info("Setting PXE boot once", "SystemURI", systemURI, "Boot settings", setBoot)
-	if err := system.SetBoot(&setBoot); err != nil {
+	resetRequired, err := r.setBoot(system, &setBoot)
+	if err != nil {
 		return fmt.Errorf("failed to set the boot order: %w", err)
+	}
+
+	// When boot properties were applied via @Redfish.Settings (e.g., AMI BMCs),
+	// they are pending and require a system reset to take effect.
+	if resetRequired && system.PowerState == schemas.OnPowerState {
+		log.V(1).Info("Resetting system to apply pending boot settings", "SystemURI", systemURI)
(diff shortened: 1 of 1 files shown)


bmc/redfish.go
```diff
@@ -212,9 +212,19 @@ func (r *RedfishBMC) SetPXEBootOnce(ctx context.Context, systemURI string) error
...
 	}
+
+	// When boot properties were applied via @Redfish.Settings (e.g., AMI BMCs),
+	// they are pending and require a system reset to take effect.
+	if resetRequired && system.PowerState == schemas.OnPowerState {
+		log.V(1).Info("Resetting system to apply pending boot settings", "SystemURI", systemURI)
+		if _, err := system.Reset(schemas.ForceRestartResetType); err != nil {
+			return fmt.Errorf("failed to reset system to apply boot settings: %w", err)
+		}
+	}
```
nagadeesh-nagaraja: I am concerned that this will create flaky issue. 

aA the moment, we only call this "SetPXEBootOnce" when system is turned off. both in "initial" and "reserved" state. I am not sure when the power will be on "On" state. 

If this is called when the power is in "On" state. This reset issued might be interpreted by the Server Reconcile logic and turn the server power back off before the scheduled job is completed. causing flaky issues.  

How about we do this instead?  This method should only be called when power is off. The power On operation should be handled by the caller or outside this method. 

currently we dont use this, but if we want to support setting this when powered on, we could create a new function or provide a flag to allow when powered on.

PR #712, opened 2026-02-27
Ensure `ServerMaintenance` is removed when performing a BMC version upgrade

### LINKED ISSUE #691: ServerMaintenance cleanups miss some resources
## Bug Description: ServerMaintenance Cleanup Logic Failure in BMCVersionReconciler

Had this yesterday. After manually installing and setting a `biosversion` and a `bmcversion` to complete, the `ServerMaintenance` related to the `biosversion` was removed but the one from the `bmcversion` was not and needed to be removed manually.

### The Problem

When a `BMCVersion` reaches the `Completed` state, the controller attempts to clean up associated `ServerMaintenance` objects via the `removeServerMaintenances()` function. However, cleanup fails silently in the following scenario:

**Early Return When `ServerMaintenanceRefs` is Nil:**

```go
func (r *BMCVersionReconciler) removeServerMaintenances(ctx context.Context, bmcVersion *metalv1alpha1.BMCVersion) error {
	// ...
	if bmcVersion.Spec.ServerMaintenanceRefs == nil {
		return nil  // ← EARLY EXIT: No cleanup occurs!
	}
	// Deletion logic never executes
```

### Examples

**Scenario 1: References Never Populated**
- ServerMaintenance objects are created during the upgrade process
- For some reason, `spec.serverMaintenanceRefs` is never populated/persisted back to the BMCVersion spec
- When completion cleanup runs, it sees nil refs and returns early
- **Result**: Orphaned ServerMaintenance objects remain indefinitely

**Scenario 2: References Cleared**
- ServerMaintenance objects exist and were properly referenced in spec
- The refs are cleared/set to nil (either manually or by a bug)
- Cleanup function can no longer find the objects to delete them
- **Result**: Orphaned ServerMaintenance objects with no owner reference path

### Expected Behavior

The cleanup function should:
1. Find all ServerMaintenance objects **controlled by this BMCVersion** (using owner references)
2. Delete them regardless of whether `spec.serverMaintenanceRefs` is populated
3. Only then clear the spec refs

### Actual Behavior

The function blindly trusts that if `serverMaintenanceRefs` is nil, there's nothing to clean up. This assumption is false when:
- Refs were never populated despite objects existing
- Refs were cleared/lost

### Impact

- Accumulation of orphaned ServerMaintenance resources
- Potential API server bloat over time
- Manual cleanup required

### Root Cause

The cleanup logic is **ref-dependent** rather than **owner-reference-dependent**. It should use Kubernetes' native owner reference mechanism to find and clean up child objects, not rely on a separate spec field that may not be in sync.

stefanhipfel:
## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable cleanup of maintenance resources when references are cleared: owned maintenance items are removed reliably, non-owned items are preserved; deletions handle missing items gracefully and include improved logging while finalization and status updates remain intact.

* **Tests**
  * Added comprehensive tests covering orphan cleanup, idempotent reconciliation, ownership enforcement, and various reference/state transition scenarios.

Fixes #691

### DIFF
--- a/internal/controller/bmcversion_controller_test.go
+++ b/internal/controller/bmcversion_controller_test.go
@@ -503,13 +503,11 @@ var _ = Describe("BMCVersion Controller", func() {
 		Eventually(Get(serverMaintenance)).Should(Succeed())
 
 		By("Ensuring that spec.serverMaintenanceRefs is populated")
-		Eventually(func(g Gomega) bool {
-			Expect(Get(bmcVersion)()).To(Succeed())
-			if len(bmcVersion.Spec.ServerMaintenanceRefs) == 0 {
-				return false
-			}
-			return bmcVersion.Spec.ServerMaintenanceRefs[0].Name == serverMaintenance.Name
-		}).Should(BeTrue())
+		Eventually(Object(bmcVersion)).Should(
+			HaveField("Spec.ServerMaintenanceRefs", ContainElement(
+				HaveField("Name", serverMaintenance.Name),
+			)),
+		)
 
 		By("Simulating cleanup by mocking upgrade completion")
 		Eventually(UpdateStatus(bmcVersion, func() {


internal/controller/bmcversion_controller.go
```diff
@@ -118,41 +118,31 @@ func (r *BMCVersionReconciler) delete(ctx context.Context, bmcVersion *metalv1al
...
-		for _, err := range errs {
-			if apierrors.IsNotFound(err) {
-				missingServerMaintenanceRef = append(missingServerMaintenanceRef, err)
-			} else {
+	// Delete all ServerMaintenance objects owned by this BMCVersion
+	for i := range serverMaintenanceList.Items {
+		serverMaintenance := &serverMaintenanceList.Items[i]
+		if serverMaintenance.DeletionTimestamp.IsZero() && metav1.IsControlledBy(serverMaintenance, bmcVersion) {
+			log.V(1).Info("Deleting ServerMaintenance", "ServerMaintenance", client.ObjectKeyFromObject(serverMaintenance))
+			if err := r.Delete(ctx, serverMaintenance); err != nil {
```
nagadeesh-nagaraja: below AI review has a point, the deletion can error due to CRD not found.

internal/controller/bmcversion_controller_test.go
```diff
@@ -419,6 +420,232 @@ var _ = Describe("BMCVersion Controller", func() {
...
+		Expect(k8sClient.Create(ctx, bmcVersion)).To(Succeed())
+
+		By("Setting spec.serverMaintenanceRefs to nil")
+		Eventually(Update(bmcVersion, func() {
+			bmcVersion.Spec.ServerMaintenanceRefs = nil
+		})).Should(Succeed())
+
+		By("Recording the non-owned ServerMaintenance for later verification")
+		nonOwnedMaintenance := &metalv1alpha1.ServerMaintenance{}
+		Expect(k8sClient.Get(ctx, client.ObjectKey{Name: orphanMaintenance.Name, Namespace: ns.Name}, nonOwnedMaintenance)).To(Succeed())
```
afritzler: `komega` `Get()`.

internal/controller/bmcversion_controller_test.go
```diff
@@ -419,6 +420,232 @@ var _ = Describe("BMCVersion Controller", func() {
...
+		Eventually(Update(bmcVersion, func() {
+			metautils.SetAnnotation(bmcVersion, "test", "verify-ownership")
+		})).Should(Succeed())
+
+		By("Verifying that the non-owned ServerMaintenance still exists")
+		Consistently(Get(nonOwnedMaintenance)).Should(Succeed())
+
+		By("Verifying no other ServerMaintenance objects were created")
+		var serverMaintenanceList metalv1alpha1.ServerMaintenanceList
+		Expect(k8sClient.List(ctx, &serverMaintenanceList)).To(Succeed())
```
afritzler: ObjectList()

PR #713, opened 2026-02-27
Add SSH-based BMC reset fallback in BMCReconciler

stefanhipfel:
# Proposed Changes

Implements a ssh based bmc reset in case the redfish api is no longer responsive.
The ssh reset is done via a separate goroutine

## Summary by CodeRabbit

* **New Features**
  * Added SSH-based fallback resets when Redfish reset requests fail with server errors.
  * Added support for forcing an SSH reset through the reset workflow.
  * Added a configurable SSH reset timeout, defaulting to two minutes.
* **Bug Fixes**
  * Improved reset handling for unavailable BMCs and transient connection failures.
  * Prevented duplicate reset attempts and improved status reporting for completed or failed resets.

### DIFF
--- a/internal/controller/bmc_controller.go
+++ b/internal/controller/bmc_controller.go
@@ -153,7 +153,7 @@ func (r *BMCReconciler) reconcile(ctx context.Context, bmcObj *metalv1alpha1.BMC
 	if err != nil {
 		if r.shouldResetBMC(bmcObj) {
 			log.V(1).Info("BMC needs reset, resetting", "BMC", bmcObj.Name)
-			if err := r.resetBMC(ctx, bmcObj, bmcClient, ReasonAutoReset, bmcAutoResetMessage); err != nil {
+			if err := r.resetBMC(ctx, bmcObj, bmcClient, nil, ReasonAutoReset, bmcAutoResetMessage); err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to reset BMC: %w", err)
 			}
 			log.V(1).Info("BMC reset initiated", "BMC", bmcObj.Name)
@@ -170,7 +170,7 @@ func (r *BMCReconciler) reconcile(ctx context.Context, bmcObj *metalv1alpha1.BMC
 			if err := r.Patch(ctx, bmcObj, client.MergeFrom(bmcBase)); err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to remove operation annotation: %w", err)
 			}
-			if err := r.resetBMCWithErr(ctx, bmcObj, bmcClient, connErr, ReasonUserReset, bmcUserResetMessage); err != nil {
+			if err := r.resetBMC(ctx, bmcObj, bmcClient, connErr, ReasonUserReset, bmcUserResetMessage); err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to reset BMC: %w", err)
 			}
 			return ctrl.Result{RequeueAfter: r.BMCClientRetryInterval}, nil
@@ -394,7 +394,7 @@ func (r *BMCReconciler) handleAnnotationOperations(ctx context.Context, bmcObj *
 	switch value {
 	case schemas.GracefulRestartResetType:
 		log.V(1).Info("Handling operation", "Operation", operation, "RedfishResetType", value)
-		if err := r.resetBMC(ctx, bmcObj, bmcClient, ReasonUserReset, bmcUserResetMessage); err != nil {
+		if err := r.resetBMC(ctx, bmcObj, bmcClient, nil, ReasonUserReset, bmcUserResetMessage); err != nil {
 			return false, fmt.Errorf("failed to reset BMC: %w", err)
 		}
 		log.Info("Handled operation", "Operation", operation)
@@ -516,11 +516,7 @@ func (r *BMCReconciler) patchBMCStatePending(ctx context.Context, bmcObj *metalv
 	return nil
 }
 
-func (r *BMCReconciler) resetBMC(ctx context.Context, bmcObj *metalv1alpha1.BMC, bmcClient bmc.BMC, reason, message string) error {
-	return r.resetBMCWithErr(ctx, bmcObj, bmcClient, nil, reason, message)
-}
-
-func (r *BMCReconciler) resetBMCWithErr(ctx context.Context, bmcObj *metalv1alpha1.BMC, bmcClient bmc.BMC, clientErr error, reason, message string) error {
+func (r *BMCReconciler) resetBMC(ctx context.Context, bmcObj *metalv1alpha1.BMC, bmcClient bmc.BMC, clientErr error, reason, message string) error {
 	log := ctrl.LoggerFrom(ctx)
 	if err := r.updateConditions(ctx, bmcObj, true, ConditionReset, corev1.ConditionTrue, reason, message); err != nil {
 		return fmt.Errorf("failed to set BMC resetting condition: %w", err)
--- a/internal/controller/bmc_controller.go
+++ b/internal/controller/bmc_controller.go
@@ -60,6 +60,9 @@ type BMCReconciler struct {
 	// DNSRecordTemplatePath is the path to the file containing the DNSRecord template.
 	DNSRecordTemplate string
 	Conditions        *conditionutils.Accessor
+
+	// SSHResetTimeout defines the timeout for SSH reset operations (dial + command execution).
+	SSHResetTimeout time.Duration
 }
 
 // +kubebuilder:rbac:groups=metal.ironcore.dev,resources=endpoints,verbs=get;list;watch
@@ -122,6 +125,21 @@ func (r *BMCReconciler) reconcile(ctx context.Context, bmcObj *metalv1alpha1.BMC
 		log.V(1).Info("Skipped BMC reconciliation")
 		return ctrl.Result{}, nil
 	}
+	// SSH reset takes priority — process it even during the waitForBMCReset window so the
+	// reset actually runs instead of being deferred until the window expires.
+	if r.hasSSHResetAnnotation(bmcObj) {
+		log.V(1).Info("SSH reset annotation detected on unresponsive BMC", "BMC", bmcObj.Name)
+		bmcBase := bmcObj.DeepCopy()
+		metautils.DeleteAnnotation(bmcObj, metalv1alpha1.OperationAnnotation)
+		if err := r.Patch(ctx, bmcObj, client.MergeFrom(bmcBase)); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to remove SSH reset annotation: %w", err)
+		}
+		if err := r.resetBMCViaSSH(ctx, bmcObj.Name); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to reset BMC via SSH: %w", err)
+		}
+		log.V(1).Info("BMC SSH reset completed", "BMC", bmcObj.Name)
+		return ctrl.Result{RequeueAfter: r.BMCClientRetryInterval}, nil
+	}
 	if r.waitForBMCReset(bmcObj, r.BMCResetWaitTime) {
 		log.V(1).Info("Skipped BMC reconciliation while waiting for BMC reset to complete")
 		if err := r.patchBMCStatePending(ctx, bmcObj); err != nil {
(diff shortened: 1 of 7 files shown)


internal/controller/bmc_controller.go
nagadeesh-nagaraja: is there a way to add unit test to  test this workflow?

internal/controller/bmc_controller.go
```diff
@@ -489,19 +499,94 @@ func (r *BMCReconciler) resetBMC(ctx context.Context, bmcObj *metalv1alpha1.BMC,
...
+		return fmt.Errorf("failed to fetch BMC: %w", err)
+	}
+	bmcBase := currentBMC.DeepCopy()
+	now := metav1.Now()
+	currentBMC.Status.LastResetTime = &now
+	if err := r.Status().Patch(ctx, currentBMC, client.MergeFrom(bmcBase)); err != nil {
+		return fmt.Errorf("failed to patch LastResetTime: %w", err)
+	}
 	return nil
 }
```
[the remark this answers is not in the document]
stefanhipfel: bmc is not namespaces

afritzler (changes requested):

internal/controller/bmc_controller.go
```diff
@@ -145,9 +159,28 @@ func (r *BMCReconciler) reconcile(ctx context.Context, bmcObj *metalv1alpha1.BMC
...
+			resetReason, resetMessage := bmcAutoResetReason, bmcAutoResetMessage
+			hasAnnotation := r.hasResetAnnotation(bmcObj)
+			if hasAnnotation {
+				log.V(1).Info("Reset annotation detected on unresponsive BMC", "BMC", bmcObj.Name)
+				resetReason, resetMessage = bmcUserResetReason, bmcUserResetMessage
+			} else {
+				log.V(1).Info("BMC needs auto-reset", "BMC", bmcObj.Name)
+			}
+
+			// Remove annotation BEFORE triggering reset to prevent duplicate resets
```
atd9876, line 173: Is this correct? doesn't the condition bmcResetConditionType = "Reset" prevent multiple resets from being triggered?

couldn't removing the annotation first lead to a situation where the reset annotation has been added manually, the annotation is deleted, the reset fails and does not retry because the annotation has been removed?

internal/controller/bmc_controller.go
```diff
@@ -627,10 +801,40 @@ func (r *BMCReconciler) enqueueBMCByBMCSecret(ctx context.Context, obj client.Ob
...
+	log.Info("Starting SSH reset worker")
+
+	for {
+		select {
+		case <-ctx.Done():
+			log.Info("Stopping SSH reset worker")
+			return nil
+		case req := <-r.sshResetQueue:
+			// Create a timeout context for each reset operation
+			resetCtx, cancel := context.WithTimeout(ctx, r.SSHResetWorkerTimeout)
```
atd9876, line 834: I may be being pedantic but this sets a context.WithTimeout of 5 minutes which calls a function where the SSH connect sets a context.WithTimeout of 2 minutes. Is there a need for a timeout on both these contexts, won't the SSH connection context always timeout and return to this context and this one will never timeout

internal/controller/bmc_controller.go
stefanhipfel: 1. Annotation is removed BEFORE reset:
   - Prevents duplicate resets from being triggered
   - The Reset condition (not the annotation) becomes the source of truth
 2. Reset state persistence in resetBMC()):
   - The Reset=True condition is set after annotation removal
   - This condition persists even if the reset fails
   - The condition acts as the "in-progress" marker, not the annotation
 3. Retry mechanism:
   - Auto-reset logic in shouldResetBMC() checks the condition, not the annotation
   - If reset fails and condition is cleared, the auto-reset mechanism can trigger again
   - Manual reset retries require re-adding the annotation
 4. Cleanup on reconnection (handlePreviousBMCResetAnnotations():
   - Clears stale Reset conditions when BMC comes back online
   - Removes leftover annotations from previous reset attempts
   - Ensures clean state for future resets

internal/controller/bmc_controller.go
stefanhipfel: In practice, the SSH connection timeout will fire first if the connection fails (2 min), and the command timeout will fire first if the command hangs (2 min). The outer timeout (5 min) serves as a safety net for pathological scenarios where the operation takes longer than expected.

The worst-case scenario is dial timeout (2 min) + command timeout (2 min) = 4 minutes, which is within the 5-minute outer timeout. So yes, the outer timeout will rarely fire in normal operation - it's there for defence against truly exceptional cases (e.g., very slow credential fetching, or network issues that don't trigger the inner timeouts).

internal/controller/bmc_controller.go
stefanhipfel: Yes, the bmcResetConditionType = 'Reset' condition does prevent multiple resets from being triggered - this is by design and is the idempotency guard for the reset operation.

xkonni (approved):
looks good, maybe rebase, fix the pipeline issues and merge it?

atd9876 (approved):

internal/controller/bmc_controller.go
```diff
@@ -478,41 +552,164 @@ func (r *BMCReconciler) shouldResetBMC(bmcObj *metalv1alpha1.BMC) bool {
...
+		if httpErr.HTTPReturnedStatusCode >= 500 && httpErr.HTTPReturnedStatusCode < 600 {
+			// 5xx server error - enqueue SSH reset work
+			// The bmcResetConditionType condition (set to True above) acts as idempotency guard:
+			// - waitForBMCReset() will prevent new resets during the delay period
+			// - After delay, we retry BMC connection
+			// - Condition is cleared when connection succeeds (via handlePreviousBMCResetAnnotations)
+			select {
+			case r.sshResetQueue <- &sshResetRequest{
+				bmcName:      bmcObj.Name,
+				bmcNamespace: bmcObj.Namespace,
```
afritzler: Is that the secret or the BMC itself? BMC is a cluster scoper resource without without a namespace.

internal/controller/bmc_controller.go
```diff
@@ -478,41 +552,164 @@ func (r *BMCReconciler) shouldResetBMC(bmcObj *metalv1alpha1.BMC) bool {
...
+			default:
+				log.Error(nil, "SSH reset queue was full; reset was not scheduled", "BMC", bmcObj.Name)
+				_ = r.updateConditions(ctx,
+					bmcObj,
+					false,
+					bmcResetConditionType,
+					corev1.ConditionFalse,
+					bmcUnknownErrorReason,
+					"SSH reset queue was full; reset not scheduled",
+				)
```
afritzler: Any reason why we are not handling an error here?

afritzler (changes requested):

internal/controller/bmc_controller.go
stefanhipfel: fixed

internal/controller/bmc_controller.go
stefanhipfel: fixed

pkg/bmcutils/bmcutils.go
```diff
@@ -261,15 +265,48 @@ func GetServerNameFromBMCandIndex(index int, bmcObj *metalv1alpha1.BMC) string {
...
+func expandPath(path string) (string, error) {
+	if len(path) > 0 && path[0] == '~' {
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			return "", err
+		}
+		return filepath.Join(homeDir, path[1:]), nil
+	}
+	return path, nil
+}
```
afritzler: @coderabbitai isn't there a go native way of doing a path expansion?

internal/controller/bmc_controller.go
```diff
@@ -40,6 +41,15 @@ const (
...
 	bmcAutoResetMessage = "BMC reset initiated automatically after repeated connection failures. Waiting for it to come back online."
+
+	// BMC-specific condition and reason aliases for shared constants
+	bmcResetConditionType         = ConditionReset
+	bmcAutoResetReason            = ReasonAutoReset
+	bmcUserResetReason            = ReasonUserReset
+	bmcUnknownErrorReason         = ReasonUnknownError
+	bmcAuthenticationFailedReason = ReasonAuthenticationFailed
+	bmcConnectionFailedReason     = ReasonConnectionFailed
+	bmcInternalErrorReason        = ReasonInternalError
```
afritzler: Those can go to the conditions.go and must be potentially harmonized with the existing conditions and reasons.

afritzler (changes requested):

internal/controller/bmc_controller.go
```diff
@@ -60,6 +61,18 @@ type BMCReconciler struct {
 	// DNSRecordTemplatePath is the path to the file containing the DNSRecord template.
 	DNSRecordTemplate string
 	Conditions        *conditionutils.Accessor
+
+	// SSH reset worker infrastructure
+	sshResetQueue chan *sshResetRequest
```
adracus: This queue alongside the `runSSHResetWorker` is effectively a separate controller. In the docs, controller-runtime stated that a channel source is usually the last resort and reconciliations should be driven by the in-cluster state if possible. How about instead of having this queue, we can use the existing `handleAnnotationOperations` function to do the ssh-based reset? Then instead of writing to the channel we could set the operation annotation.

internal/controller/bmc_controller.go
```diff
@@ -122,6 +125,21 @@ func (r *BMCReconciler) reconcile(ctx context.Context, bmcObj *metalv1alpha1.BMC
...
+	// reset actually runs instead of being deferred until the window expires.
+	if r.hasSSHResetAnnotation(bmcObj) {
+		log.V(1).Info("SSH reset annotation detected on unresponsive BMC", "BMC", bmcObj.Name)
+		bmcBase := bmcObj.DeepCopy()
+		metautils.DeleteAnnotation(bmcObj, metalv1alpha1.OperationAnnotation)
+		if err := r.Patch(ctx, bmcObj, client.MergeFrom(bmcBase)); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to remove SSH reset annotation: %w", err)
+		}
+		if err := r.resetBMCViaSSH(ctx, bmcObj.Name); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to reset BMC via SSH: %w", err)
```
afritzler, line 138: Shouldn't we do the removal after the reset happens?

internal/controller/bmc_controller.go
```diff
@@ -484,14 +517,29 @@ func (r *BMCReconciler) patchBMCStatePending(ctx context.Context, bmcObj *metalv
 }
 
 func (r *BMCReconciler) resetBMC(ctx context.Context, bmcObj *metalv1alpha1.BMC, bmcClient bmc.BMC, reason, message string) error {
+	return r.resetBMCWithErr(ctx, bmcObj, bmcClient, nil, reason, message)
+}
+
+func (r *BMCReconciler) resetBMCWithErr(ctx context.Context, bmcObj *metalv1alpha1.BMC, bmcClient bmc.BMC, clientErr error, reason, message string) error {
```
afritzler: @coderabbitai can this method and resetBMCViaSSH be collapsed into one? Does it make sense to have two different reset methods here?

afritzler (changes requested):

internal/controller/bmc_controller.go
afritzler: @stefanhipfel i think we should drop the arbitrary wrapper as suggested here in favor of `resetBMC`.

internal/controller/bmc_controller.go
stefanhipfel: if we run into an error during ssh reset (which is very fragile, since bmc is not responding), it will never remove the annotation and instead keep retrying ..

internal/controller/bmc_controller.go
stefanhipfel: ok

afritzler:
@coderabbitai resume

internal/controller/bmc_controller.go
```diff
@@ -135,14 +153,28 @@ func (r *BMCReconciler) reconcile(ctx context.Context, bmcObj *metalv1alpha1.BMC
...
 			log.V(1).Info("BMC needs reset, resetting", "BMC", bmcObj.Name)
-			if err := r.resetBMC(ctx, bmcObj, bmcClient, ReasonAutoReset, bmcAutoResetMessage); err != nil {
+			if err := r.resetBMC(ctx, bmcObj, bmcClient, nil, ReasonAutoReset, bmcAutoResetMessage); err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to reset BMC: %w", err)
 			}
 			log.V(1).Info("BMC reset initiated", "BMC", bmcObj.Name)
 			return ctrl.Result{
 				RequeueAfter: r.BMCClientRetryInterval,
 			}, nil
 		}
```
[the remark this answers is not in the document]
afritzler: @stefanhipfel can you please check if this is a valid finding?

afritzler (approved):

PR #744, opened 2026-03-18
Improve deletion logic in of `BMC*` and `BIOS*` types

nagadeesh-nagaraja:
# Proposed Changes
Improves deletion logics in some CRDs which have dependencies to another objects.

## Summary by CodeRabbit

* **Bug Fixes**
  * Controllers now allow resource deletion when referenced Servers or BMCs are missing, enabling cleanup to proceed.
  * Log messages clarified to better indicate whether a referenced BMC or Server could not be fetched.

* **Refactor**
  * Server maintenance orchestration extracted into a dedicated handler to simplify reconciliation flow.

* **Other**
  * Simplified error propagation when obtaining BMC clients to reduce wrapped error noise.

### DIFF
--- a/internal/controller/server_controller.go
+++ b/internal/controller/server_controller.go
@@ -221,27 +221,6 @@ func (r *ServerReconciler) reconcile(ctx context.Context, server *metalv1alpha1.
 			log.V(1).Info("BMC is not available, skipping", "BMC", server.Spec.BMCRef.Name, "Server", server.Name, "error", err)
 			return ctrl.Result{RequeueAfter: r.ResyncInterval}, nil
 		}
-		if apierrors.IsNotFound(err) {
-			var statusErr *apierrors.StatusError
-			if errors.As(err, &statusErr) && statusErr.ErrStatus.Details != nil {
-				d := statusErr.ErrStatus.Details
-				log.V(1).Info("Referenced resource not found while creating BMC client",
-					"group", d.Group,
-					"kind", d.Kind,
-					"name", d.Name,
-					"server", server.Name,
-					"message", statusErr.ErrStatus.Message,
-				)
-			} else {
-				log.V(1).Info("BMC or BMCSecret resource not found while creating BMC client",
-					"BMC", server.Spec.BMCRef,
-					"Server", server.Name,
-					"error", err,
-				)
-			}
-			// if BMC is not found, requeue until BMC is available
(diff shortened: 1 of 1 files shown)


internal/controller/biosversion_controller.go
```diff
@@ -92,6 +92,10 @@ func (r *BIOSVersionReconciler) shouldDelete(ctx context.Context, biosVersion *m
 
 	if controllerutil.ContainsFinalizer(biosVersion, BIOSVersionFinalizer) &&
 		biosVersion.Status.State == metalv1alpha1.BIOSVersionStateInProgress {
+		if _, err := GetServerByName(ctx, r.Client, biosVersion.Spec.ServerRef.Name); apierrors.IsNotFound(err) {
+			log.V(1).Info("Server not found, proceeding with deletion", "Server", biosVersion.Spec.ServerRef.Name)
+			return true
+		}
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: logically this can not enter this state without ServerRef being not nil. 

the state will remain in "pending" state (due to check in reconcile method and as this is immutable field.

internal/controller/biosversion_controller.go
```diff
@@ -188,6 +192,17 @@ func (r *BIOSVersionReconciler) transitionState(ctx context.Context, biosVersion
 			log.V(1).Info("BMC is not available, skipping", "BMC", server.Spec.BMCRef.Name, "Server", server.Name, "error", err)
 			return true, nil
 		}
+		if apierrors.IsNotFound(err) {
```
afritzler: Same as above.

internal/controller/biossettings_controller.go
```diff
@@ -275,6 +279,27 @@ func (r *BIOSSettingsReconciler) reconcile(ctx context.Context, settings *metalv
...
+					"server", server.Name,
+					"message", statusErr.ErrStatus.Message,
+				)
+			} else {
+				log.V(1).Info("BMC or BMCSecret resource not found while creating BMC client",
+					"BMC", server.Spec.BMCRef,
+					"Server", server.Name,
+					"error", err,
+				)
+			}
```
afritzler: Why not just logging the error from the client creation method instead of doing the dance here?

afritzler (changes requested):

afritzler (approved):

PR #748, opened 2026-03-19
Prepare deprecation of APIVersion, Kind, UID fields in ObjectReferences

### LINKED ISSUE #625: Discuss automatic mutation of `ref.UID` during cluster migration (e.g. CAPI move)
## Background

We currently use `ref.UID` as defined in
[https://github.com/ironcore-dev/metal-operator/blob/main/api/v1alpha1/common_types.go#L16](https://github.com/ironcore-dev/metal-operator/blob/main/api/v1alpha1/common_types.go#L16)
across many reference fields in the project.

This includes critical fields such as:

* `spec.serverClaimRef`
* `spec.bootConfigurationRef`

## Problem

During scenarios like a **CAPI move**, resources are recreated in the target cluster and therefore receive **new UIDs**. This breaks existing references, since the `UID` stored in `ref.UID` no longer matches the newly created object.

The situation is especially problematic because:

* `spec.serverClaimRef` is **immutable once set**
* This prevents us from simply rewriting the UID post-migration
* As a result, migrated objects may end up with permanently broken references

## Questions / Discussion Points

1. **Automatic UID mutation**

   * Should we support automatic mutation of `ref.UID` during migrations (e.g. via webhook, controller logic, or a dedicated migration step)?
   * If yes, where should this logic live and how do we handle immutable fields safely?

2. **Immutability vs migration**

   * Do we need to relax immutability guarantees for certain ref fields specifically to support migrations?
   * Or do we need a different migration strategy entirely?

3. **Do we need UID references at all?**

   * What is the concrete value of storing `UID` in refs today?
   * Can we transition to **namespace/name-only references** and rely on Kubernetes guarantees instead?
   * Would dropping `UID` refs simplify long-term operability and avoid rewrite issues in future migrations?

## Goal

Align on a clear strategy that:

* Supports cluster migration workflows safely
* Avoids brittle or irreversible references
* Minimizes future operational and design complexity

maxmoehl:
When moving resources between clusters the UID field of our custom references must be manually updated afterwards which is especially tricky given that the references as a whole are immutable.

This commit makes the APIVersion, Kind, and UID fields in the ObjectReference option and drops all code paths which rely on it. Once this change has been released those fields can be completely removed from the code base. During the migration the controller will actively nilify any instance of those fields it encounters.

Resolves: #625

## Summary by CodeRabbit

* **New Features**
  * Introduced an immutable reference type: Name+Namespace are permanent once set; updates that change them are rejected.

* **Changes**
  * Reference matching for claims/maintenance/boot now uses Name+Namespace.
  * apiVersion, kind, and uid are deprecated on references and are no longer required in CRD schemas.
  * Controllers now automatically remove deprecated reference fields and persist cleaned resources early in reconciliation.

* **Documentation**
  * API docs updated to describe the immutable reference type and deprecations.

* **Tests**
  * Tests and webhooks updated/added to validate immutability and new matching.


afritzler (changes requested):

afritzler (approved):

afritzler:
@coderabbitai create an issue to remove the deprecated field and the migration logic.

PR #750, opened 2026-03-23
Secure discovery boot data with jwt signed tokens

### LINKED ISSUE #749: Secure discovery boot data with per-server tokens
## Problem

Currently, the discovery registry accepts data from any machine claiming a given server UUID. There is no mechanism to verify that the discovery data was actually submitted by the legitimate machine and not by a rogue third party in the cluster.

## Proposed Solution

Introduce per-server authentication tokens for the discovery boot flow:

1. **Token creation**: When a server enters discovery boot, generate a unique token tied to the server's UUID.
2. **Secret storage**: Store this token as a Kubernetes `Secret` associated with the server.
3. **Ignition delivery**: Pass the token to the machine via the ignition configuration so the booting machine receives its own token.
4. **Registry enforcement**: The discovery registry must require the matching token in requests for a given server UUID. Requests without a valid token are rejected.

## Security Goal

This ensures that discovery data stored in the registry is authentic — it can only have been submitted by the machine that actually booted with the correct token. A rogue third party in the cluster cannot inject fake discovery data for a server they do not control.

## Acceptance Criteria

- [ ] A token is generated per server at discovery boot time
- [ ] The token is stored in a `Secret` (e.g. named `<server-name>-discovery-token` or similar)
- [ ] The token is included in the ignition config passed to the booting machine
- [ ] The registry validates the token on incoming discovery data submissions, matching it to the server UUID
- [ ] Submissions with missing or incorrect tokens are rejected with an appropriate error

stefanhipfel:
# Proposed Changes

This PR implements JWT-signed discovery tokens to secure the discovery boot data flow between metalprobe agents and the registry server.

## Key Changes

**Security Implementation:**
- **JWT-based authentication** - Implements RFC 7519 standard JWT tokens with HMAC-SHA256 (HS256) signing
- **Signing secret management** - Controller creates and manages a 32-byte shared secret in Kubernetes
- **Token generation** - Controller generates signed tokens for each server, embedded in ignition configs
- **Token validation** - Registry validates tokens on `/register` and `/bootstate` endpoints
- **1-hour token lifetime** - Balances security (limited exposure window) with operational flexibility
- **Automatic expiration** - JWT library enforces expiration validation

**Code Changes:**
- Add `internal/token/` package for JWT token generation and verification
- Update controller to generate per-server discovery tokens
- Update registry to validate discovery tokens
- Update metalprobe to send discovery tokens in API requests
- Add comprehensive unit tests for token functionality
- Add integration tests validating end-to-end token flow

**Documentation:**
- Add `docs/discovery-token-security.md` - Complete operational guide covering:
  - Token format and security properties
  - Secret lifecycle management
  - Troubleshooting and monitoring
  - Testing procedures

**Dependencies:**
- Add `github.com/golang-jwt/jwt/v5` - Industry-standard JWT library

## Review Feedback Addressed

All CodeRabbitAI feedback has been addressed in commit 986d7cc4:
- ✅ Fixed probe authentication retry logic bug
- ✅ Fixed registry secret loading race condition
- ✅ Fixed controller secret creation race
- ✅ Fixed test race condition with bcrypt hashes
- ✅ Improved error logging
- ✅ Added UUID validation
- ✅ Added test mode security warning

## Testing

- ✅ Unit tests: 85/85 passing
- ✅ Token tests: 16/16 passing (100% coverage)
- ✅ Probe tests: All passing
- ✅ Registry tests: All passing
- ✅ Controller tests: All passing
- ✅ Race detector: Clean

## Verification

The implementation provides:
- Authentication: Only controller can generate valid tokens (requires signing secret)
- Integrity: Tokens cannot be tampered with (signature verification)
- Binding: Each token bound to specific systemUUID
- Freshness: Tokens expire after 1 hour
- Standard format: Industry-standard JWT (RFC 7519)

Fixes #749

## Summary by CodeRabbit

* **New Features**
  * Added JWT-based discovery token authentication to secure server registration and validation endpoints, preventing unauthorized server discovery.

* **Documentation**
  * Added security documentation covering token lifecycle, Kubernetes secret management, operational best practices, troubleshooting, and monitoring recommendations.


nagadeesh-nagaraja (approved):
Might need a rebase with the latest changes which went in and check if unexpected silent error or failure is happening :D apart from that LGTM

stefanhipfel:
@xkonni @afritzler maybe we can discuss this on Thursday? The PR has been open for quite some time.

xkonni (commented):
yes. as nag said, rebase and then we can likely get this in.

xkonni (approved):

internal/controller/server_controller.go
```diff
@@ -817,6 +833,99 @@ func (r *ServerReconciler) applyDefaultIgnitionForServer(ctx context.Context, se
...
+			// Another reconciler won the race - fetch the existing secret
+			ctrl.LoggerFrom(ctx).Info("Signing secret already exists, retrieving it")
+
+			secret := &v1.Secret{}
+			if err := r.Get(ctx, types.NamespacedName{
+				Name:      secretName,
+				Namespace: r.ManagerNamespace,
+			}, secret); err != nil {
+				return nil, fmt.Errorf("failed to retrieve existing signing secret: %w", err)
+			}
```
nagadeesh-nagaraja, line 1016: This Get call might return the same error as earlier `Get` call because of the cache. 
we need to Get from the API bypassing the cache directly. 

it might be simpler to return and come back after next reconcile to check on it again rather than recalling Get here.

internal/controller/server_controller_test.go
```diff
@@ -30,100 +34,33 @@ import (
...
-						Name: bmcSecret.Name,
-					},
-				},
-				ReclaimPolicy: metalv1alpha1.ServerReclaimPolicyRetain,
-			},
-		}
-		Expect(k8sClient.Create(ctx, server)).To(Succeed())
+// getSignedDiscoveryToken retrieves the signing secret and generates a signed discovery token
+// for the given systemUUID. This is a test helper to avoid code duplication.
+func getSignedDiscoveryToken(ctx SpecContext, k8sClient client.Client, ns, systemUUID string) (string, error) {
```
nagadeesh-nagaraja: This should ideally move to helper_test.go.

internal/token/token.go
```diff
@@ -0,0 +1,121 @@
...
+	// Validate UUID format to prevent delimiter injection
+	if !uuidRegex.MatchString(systemUUID) {
+		return "", fmt.Errorf("systemUUID must be a valid RFC 4122 UUID format (got: %s)", systemUUID)
+	}
+
+	// Create JWT claims
+	claims := jwt.MapClaims{
+		"sub": systemUUID,                                        // Subject: scoped to this server
+		"iat": jwt.NewNumericDate(time.Now()),                    // Issued at
+		"exp": jwt.NewNumericDate(time.Now().Add(1 * time.Hour)), // Expires in 1 hour
```
nagadeesh-nagaraja: this should be configurable. hard coding is not recommended.

internal/token/token.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	// Create token with HS256 signing method
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
```
nagadeesh-nagaraja: signing method should also be parametarised, taking enum as a func parameter here is good. it easier to maintain the code in future if things gets changed or change this method if needed.

internal/token/token.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/golang-jwt/jwt/v5"
+)
+
+// uuidRegex matches RFC 4122 UUID format (8-4-4-4-12 hex digits)
+var uuidRegex = regexp.MustCompile(`^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$`)
+
+// GenerateSigningSecret generates a cryptographically secure signing secret
+// for HMAC token generation. Returns a 32-byte (256-bit) secret.
+func GenerateSigningSecret() ([]byte, error) {
+	secret := make([]byte, 32)
```
nagadeesh-nagaraja, line 27: It is better to not hard code here. can use it as a const or pass it in the paramters.

nagadeesh-nagaraja (changes requested):
Please make sure the "Make tilt-up" function is still working with these changes.

internal/controller/server_controller.go
stefanhipfel: using APIReader.Get() here would also solve the issue

internal/token/token.go
stefanhipfel: we already have more than 40 flags in our manager. I don't want to add more to that. We need a different solution in a separate PR.
@coderabbitai create an issue for config-file/ComponentConfig-style struct rather than more flag.Var calls.

PR #771, opened 2026-04-01
feat: Add  enhanced server metadata management (using servermetadata)

stefanhipfel:
# Proposed Changes


afritzler:
Closing as #762 has been closed an will be revisited. Also those topology specific labels we should move to the maintenance operator (or another 3rd party component).

Issue #772, opened 2026-04-01
ServerClaim race condition causes repeated claim/release loop


xkonni:
# Bug: ServerClaim race condition causes repeated claim/release loop

**Component:** `server` controller, `serverclaim` controller, `bmc` controller  
**Severity:** High  
**Branch:** `bug/serverclaim_race_condition`

## Summary

A server enters a continuous `Available → Reserved → Available` oscillation loop.
Multiple `ServerClaim` objects attempt to claim it in rapid succession, each overwriting
the previous claim's `ServerClaimRef` before the server has a chance to transition to
`Reserved`. The server never durably settles in `Reserved` state for any single claim,
and the affected nodes are never provisioned.

## Observed timeline

Relative timestamps, T+0 = start of observed window.

| Time | Controller | Event |
|------|-----------|-------|
| T+0s | `serverclaim` | Claim A reads stale cache → sees `ServerClaimRef` with wrong UID → logs "Server claim ref UID does not match claim" (×4 over several seconds) |
| T+0s | `server` | Runs Available handler: "Updated BIOS boot order", "Ensured initial boot configuration is deleted", transitions to `Reserved` |
| T+0s | `serverclaim` | Claim B successfully binds: "Ensured PowerState for Server" |
| T+0s | `bmc` | "Created or patched Server" — **Operation: unchanged** (no-op write, still bumps `ResourceVersion`) |
| T+2–5s | `server` | Transitions `Reserved` → **`Available`** |
| T+9s | `serverclaim` | Claim C: "Matching server found" → "Patched ServerClaimRef" → immediately sees "Server is not in reserved state" (state=`Available`) |
| T+11s | `server` | Transitions `Reserved` → **`Available`** |
| T+17s | `serverclaim` | Claim D: "Matching server found" → "Patched ServerClaimRef" → "Server is not in reserved state" |
| T+18s | `server` | Transitions `Reserved` → **`Available`** |
| T+24s | `serverclaim` | Claims E and F race: **both** find server unclaimed simultaneously. E wins the patch. F logs: `ERROR failed to patch claim ref for server: the object has been modified; please apply your changes to the latest version and try again` |
| T+26s | `server` | Transitions `Reserved` → **`Available`** |
| T+27s | `serverclaim` | Claim F retries: "Matching server found" → "Patched ServerClaimRef" → immediately sees "Server is not in reserved state" (state=`Available`) |
| T+28s | `server` | Transitions `Reserved` → **`Available`** |
[...]

PR #773, opened 2026-04-01
Fix ServerClaim race condition and BMC-driven state oscillation

xkonni:
# Fix ServerClaim race condition and BMC-driven state oscillation

## Summary

Under load, multiple `ServerClaim` reconcilers could simultaneously find the same server
unclaimed due to a stale informer cache, overwrite each other's `ServerClaimRef`, and
leave no claim successfully bound. A periodic BMC reconcile compounded this by patching
the `Server` object even when nothing changed, re-triggering the server state machine and
repeatedly resetting the server to `Available` — reopening the race window on every cycle.

See issue #772 

### Fix ServerClaim race: bypass cache before claiming a server

`ensureObjectRefForServer` now re-fetches the `Server` directly from the API server
(via `APIReader`) before writing `ServerClaimRef`. Concurrent claim reconcilers working
off a stale informer cache can no longer both see `ServerClaimRef == nil` and overwrite
each other's claim — the loser sees the already-set ref on the fresh read and bails out
cleanly.

### Skip BMC server patch when discovered fields are unchanged

`discoverServers` now only mutates the `Server` object when `SystemUUID`, `SystemURI`,
`BMCRef`, or labels have actually changed. Previously, `CreateOrPatch` was called
unconditionally on every BMC reconcile, issuing a write even when nothing differed
(`Operation: unchanged`). This bumped `ResourceVersion` and re-triggered the server
controller every ~7–10 seconds, keeping the oscillation loop alive.

### Fix Available→Reserved transition on stale cache in server controller

`handleAvailableState` now re-fetches the `Server` from the API server before checking
`ServerClaimRef`. Without this, a BMC-triggered reconcile arriving with a stale cache
snapshot would find `ServerClaimRef == nil`, skip the `Reserved` transition, and run the
Available boilerplate — resetting the server state even though a claim had already been
written.

## Test plan

- [ ] Verify under concurrent claim load that a server transitions cleanly to `Reserved`
      without oscillating back to `Available`
- [ ] Confirm BMC reconcile no longer logs repeated `Operation: unchanged` patches
      followed by server controller re-entries into the Available state machine

## Summary by CodeRabbit

* **Bug Fixes**
  * Reconciliation now re-fetches fresh resource state to avoid acting on stale cached data and to detect concurrent claims, deferring when conflicts occur.

* **Improvements**
  * Skip unnecessary updates when resources already match desired configuration.
  * More granular logging for created/updated/no-op outcomes.

* **Tests**
  * Test setup aligned with runtime reconciler behavior.

### DIFF
--- a/internal/controller/bmc_controller.go
+++ b/internal/controller/bmc_controller.go
@@ -291,7 +291,14 @@ func (r *BMCReconciler) discoverServers(ctx context.Context, bmcClient bmc.BMC, 
 			errs = append(errs, fmt.Errorf("failed to create or patch server %s: %w", server.Name, err))
 			continue
 		}
-		log.V(1).Info("Created or patched Server", "Server", server.Name, "Operation", opResult)
+		switch opResult {
+		case controllerutil.OperationResultCreated:
+			log.V(1).Info("Created Server", "Server", server.Name)
+		case controllerutil.OperationResultUpdated:
+			log.V(1).Info("Updated Server", "Server", server.Name)
+		default:
+			log.V(1).Info("Server already up to date", "Server", server.Name)
+		}
 	}
 	if len(errs) > 0 {
 		return fmt.Errorf("errors occurred during server discovery: %v", errs)
--- a/internal/controller/server_controller.go
+++ b/internal/controller/server_controller.go
@@ -81,6 +81,7 @@ const (
 // ServerReconciler reconciles a Server object
 type ServerReconciler struct {
 	client.Client
+	APIReader               client.Reader
 	Scheme                  *runtime.Scheme
 	Insecure                bool
 	ManagerNamespace        string
@@ -449,6 +450,18 @@ func (r *ServerReconciler) handleAvailableState(ctx context.Context, bmcClient b
 	if err := r.ensureIndicatorLED(ctx, server); err != nil {
 		return false, fmt.Errorf("failed to ensure server indicator led: %w", err)
 	}
+
+	// Re-fetch directly from the API server before checking ServerClaimRef.
+	// The object passed into this handler may be from the informer cache and
+	// could be stale if a ServerClaim controller just wrote the ref. Without
+	// this, a BMC-triggered reconcile that arrives with a stale cache snapshot
+	// would skip the Reserved transition even though the claim already landed.
+	fresh := &metalv1alpha1.Server{}
+	if err := r.APIReader.Get(ctx, client.ObjectKeyFromObject(server), fresh); err != nil {
+		return false, client.IgnoreNotFound(err)
+	}
+	*server = *fresh
+
 	if server.Spec.ServerClaimRef != nil {
 		if modified, err := r.patchServerState(ctx, server, metalv1alpha1.ServerStateReserved); err != nil || modified {
 			return true, err
--- a/internal/controller/serverclaim_controller.go
+++ b/internal/controller/serverclaim_controller.go
@@ -32,6 +32,7 @@ const (
 // ServerClaimReconciler reconciles a ServerClaim object
 type ServerClaimReconciler struct {
 	client.Client
+	APIReader               client.Reader
 	Cache                   cache.Cache
 	Scheme                  *runtime.Scheme
 	MaxConcurrentReconciles int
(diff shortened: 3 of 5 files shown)


internal/controller/bmc_controller.go
```diff
@@ -281,17 +282,35 @@ func (r *BMCReconciler) discoverServers(ctx context.Context, bmcClient bmc.BMC,
...
+			bmcRef := &corev1.LocalObjectReference{Name: bmcObj.Name}
+			if server.Spec.SystemUUID == uuid &&
+				server.Spec.SystemURI == s.URI &&
+				server.Spec.BMCRef != nil && server.Spec.BMCRef.Name == bmcRef.Name &&
+				maps.Equal(server.Labels, bmcObj.Labels) {
+				return controllerutil.SetControllerReference(bmcObj, server, r.Scheme)
+			}
 			metautils.SetLabels(server, bmcObj.Labels)
-			server.Spec.SystemUUID = strings.ToLower(s.UUID)
+			server.Spec.SystemUUID = uuid
```
nagadeesh-nagaraja: This is going to cause problem with all the existing server (if they have uppercase letter) 

we will need to think about migration.

internal/controller/bmc_controller.go
```diff
@@ -281,17 +282,35 @@ func (r *BMCReconciler) discoverServers(ctx context.Context, bmcClient bmc.BMC,
...
+			// reconcile does not bump ResourceVersion and re-trigger the
+			// server controller unnecessarily.
+			uuid := strings.ToLower(s.UUID)
+			bmcRef := &corev1.LocalObjectReference{Name: bmcObj.Name}
+			if server.Spec.SystemUUID == uuid &&
+				server.Spec.SystemURI == s.URI &&
+				server.Spec.BMCRef != nil && server.Spec.BMCRef.Name == bmcRef.Name &&
+				maps.Equal(server.Labels, bmcObj.Labels) {
+				return controllerutil.SetControllerReference(bmcObj, server, r.Scheme)
+			}
```
nagadeesh-nagaraja: this is already taken care in `Controllerutil.CreateOrPatch` right? why manually doing this so?

internal/controller/server_controller.go
```diff
@@ -449,6 +450,18 @@ func (r *ServerReconciler) handleAvailableState(ctx context.Context, bmcClient b
...
+	// Re-fetch directly from the API server before checking ServerClaimRef.
+	// The object passed into this handler may be from the informer cache and
+	// could be stale if a ServerClaim controller just wrote the ref. Without
+	// this, a BMC-triggered reconcile that arrives with a stale cache snapshot
+	// would skip the Reserved transition even though the claim already landed.
+	fresh := &metalv1alpha1.Server{}
+	if err := r.APIReader.Get(ctx, client.ObjectKeyFromObject(server), fresh); err != nil {
+		return false, client.IgnoreNotFound(err)
+	}
+	*server = *fresh
```
nagadeesh-nagaraja, line 463: @coderabbitai is there a known or better kubernetes pattern to make sure we get fresh object here?

internal/controller/serverclaim_controller.go
```diff
@@ -189,22 +190,44 @@ func (r *ServerClaimReconciler) reconcile(ctx context.Context, claim *metalv1alp
...
-			Namespace: claim.Namespace,
-			Name:      claim.Name,
-		}
-		if err := r.Patch(ctx, server, client.MergeFromWithOptions(serverBase, client.MergeFromWithOptimisticLock{})); err != nil {
-			return fmt.Errorf("failed to patch claim ref for server: %w", err)
+	if ref := fresh.Spec.ServerClaimRef; ref != nil {
+		if ref.Name == claim.Name && ref.Namespace == claim.Namespace {
+			// Already claimed by us — propagate fresh state to caller and continue.
+			*server = *fresh
+			return nil
```
nagadeesh-nagaraja: duplicated check, its already check in parent function (claimServer L318)

internal/controller/serverclaim_controller.go
```diff
@@ -189,22 +190,44 @@ func (r *ServerClaimReconciler) reconcile(ctx context.Context, claim *metalv1alp
...
+		*server = *fresh
+		return nil
 	}
+
+	if fresh.Status.State != metalv1alpha1.ServerStateAvailable {
+		log.V(1).Info("Server no longer in available state after re-fetch, skipping",
+			"Server", fresh.Name, "State", fresh.Status.State)
+		*server = *fresh
+		return nil
+	}
```
nagadeesh-nagaraja: same check in `isServerClaimable`! 
returning a nil here will make the server be claimed by this serverclaim....?

should return a error so that the server will be treated as non claimable.

internal/controller/serverclaim_controller.go
```diff
@@ -189,22 +190,44 @@ func (r *ServerClaimReconciler) reconcile(ctx context.Context, claim *metalv1alp
...
-		log.V(1).Info("Server is already claimed", "Server", server.Name, "Claim", server.Spec.ServerClaimRef.Name)
-		return nil
+
+	// Re-fetch directly from the API server, bypassing the potentially stale
+	// cache. This closes the window where two concurrent claim reconcilers both
+	// see ServerClaimRef==nil from cache and race to write their own claim.
+	fresh := &metalv1alpha1.Server{}
+	if err := r.APIReader.Get(ctx, client.ObjectKeyFromObject(server), fresh); err != nil {
+		return client.IgnoreNotFound(err)
 	}
```
nagadeesh-nagaraja: I think this should be done in the parent method..
isserverClaimable needs this the most...? most of the duplicate code here can be mitigated through this

internal/controller/serverclaim_controller.go
```diff
@@ -354,13 +377,20 @@ func (r *ServerClaimReconciler) claimServer(ctx context.Context, claim *metalv1a
...
 	err := r.ensureObjectRefForServer(ctx, claim, selectedServer)
 	if err != nil {
 		return nil, err
 	}
+	// After the direct re-fetch, check the ref was actually set to our claim.
+	// If the server was claimed concurrently, treat it as no server found and
+	// let the claim requeue when the winning server's state changes.
+	if ref := selectedServer.Spec.ServerClaimRef; ref == nil || ref.Name != claim.Name || ref.Namespace != claim.Namespace {
+		log.V(1).Info("Could not claim server, it was taken concurrently", "Server", selectedServer.Name)
+		return nil, nil
```
nagadeesh-nagaraja: this will stop the serverClaim reconcile completely... 
is that the intended behavior?

nagadeesh-nagaraja (changes requested):

internal/controller/bmc_controller.go
xkonni: it is lowercase'd in line 288

```
uuid := strings.ToLower(s.UUID)
```

 and then used later on

nagadeesh-nagaraja (approved):

davidgrun (approved):

PR #795, opened 2026-04-09
Centralize and harmonize condition/reason constants across controllers

afritzler:
# Proposed Changes

Condition and reason constants were scattered across controller files with inconsistent naming conventions (mixed prefixes, unexported names, BIOS/BMC-specific names for generic concepts) and true duplicates using identical string values. This made it hard to discover existing constants and led to accidental duplication when adding new controllers.

Consolidate shared condition and reason constants into a central conditions.go file and normalize naming to a consistent Condition<Name>/Reason<Name> pattern. Generalize BIOS/BMC-specific names and string values where the concepts are generic (e.g., ConditionBMCResetIssued → ConditionResetIssued,
"BIOSSettingsTimedOut" → "SettingsTimedOut").

Since this change would be breaking, a migration logic has been introduced which migrate the legacy conditions to the new format.

## Summary by CodeRabbit

* **Refactor**
  * Standardized and consolidated status condition and reason identifiers across controllers; introduced shared condition definitions and cleaned up constant declarations.
  * Renamed several finalizer identifiers to unexported names for consistency.

* **Chores / Migration**
  * Added automatic reconciliation-time migration to map legacy condition types and reasons to new identifiers, preserving resource status.

* **Tests**
  * Updated tests to use the renamed condition and reason identifiers.

### DIFF
--- a/internal/controller/biossettings_controller.go
+++ b/internal/controller/biossettings_controller.go
@@ -63,6 +63,19 @@ const (
 	ReasonSettingsInvalid               = "SettingsProvidedAreNotValid"
 )
 
+// legacyBIOSSettingsConditionTypes maps old condition type strings to their new values.
+// This ensures backward compatibility with existing CRs created before the rename.
+var legacyBIOSSettingsConditionTypes = map[string]string{
+	"BIOSSettingsCheckPendingSettings":    ConditionPendingSettingsCheck,
+	"BIOSSettingsDuplicateKeys":           ConditionSettingsDuplicateKeys,
+	"BIOSSettingUpdateStartTime":          ConditionSettingsUpdateStartTime,
+	"BIOSSettingsTimedOut":                ConditionSettingsTimedOut,
+	"ServerPowerOnCondition":              ConditionSettingsServerPowerOn,
+	"ServerRebootPostUpdateHasBeenIssued": ConditionSettingsRebootPostUpdate,
+	"BMCResetIssued":                      ConditionResetIssued,
+	"BIOSVersionUpdatePending":            ConditionVersionUpdatePending,
+}
+
 // BIOSSettingsReconciler reconciles a BIOSSettings object
 type BIOSSettingsReconciler struct {
 	client.Client
@@ -93,7 +106,28 @@ func (r *BIOSSettingsReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	if err := r.Get(ctx, req.NamespacedName, biosSettings); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	if err := r.migrateLegacyConditions(ctx, biosSettings); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to migrate legacy conditions: %w", err)
+	}
 	return r.reconcileExists(ctx, biosSettings)
+}
+
+// migrateLegacyConditions renames legacy condition type strings on existing CRs.
+// TODO: Remove this migration in the next release once all CRs have been reconciled.
+func (r *BIOSSettingsReconciler) migrateLegacyConditions(ctx context.Context, settings *metalv1alpha1.BIOSSettings) error {
+	settingsBase := settings.DeepCopy()
+	migrated := migrateConditionTypes(settings.Status.Conditions, legacyBIOSSettingsConditionTypes)
+	for i := range settings.Status.FlowState {
+		if migrateConditionTypes(settings.Status.FlowState[i].Conditions, legacyBIOSSettingsConditionTypes) {
+			migrated = true
+		}
+	}
+	if !migrated {
+		return nil
+	}
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("Migrated legacy condition types on BIOSSettings")
+	return r.Status().Patch(ctx, settings, client.MergeFrom(settingsBase))
 }
 
 func (r *BIOSSettingsReconciler) reconcileExists(ctx context.Context, settings *metalv1alpha1.BIOSSettings) (ctrl.Result, error) {
--- a/internal/controller/biossettings_controller.go
+++ b/internal/controller/biossettings_controller.go
@@ -66,14 +66,15 @@ const (
 // legacyBIOSSettingsConditionTypes maps old condition type strings to their new values.
 // This ensures backward compatibility with existing CRs created before the rename.
 var legacyBIOSSettingsConditionTypes = map[string]string{
-	"BIOSSettingsCheckPendingSettings":    ConditionPendingSettingsCheck,
-	"BIOSSettingsDuplicateKeys":           ConditionSettingsDuplicateKeys,
-	"BIOSSettingUpdateStartTime":          ConditionSettingsUpdateStartTime,
-	"BIOSSettingsTimedOut":                ConditionSettingsTimedOut,
-	"ServerPowerOnCondition":              ConditionSettingsServerPowerOn,
-	"ServerRebootPostUpdateHasBeenIssued": ConditionSettingsRebootPostUpdate,
-	"BMCResetIssued":                      ConditionResetIssued,
-	"BIOSVersionUpdatePending":            ConditionVersionUpdatePending,
+	"BIOSSettingsCheckPendingSettings":     ConditionPendingSettingsCheck,
+	"BIOSSettingsDuplicateKeys":            ConditionSettingsDuplicateKeys,
+	"BIOSSettingUpdateStartTime":           ConditionSettingsUpdateStartTime,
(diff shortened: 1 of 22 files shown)


internal/controller/conditions.go
```diff
@@ -0,0 +1,74 @@
...
+	ReasonConnected = "BMCConnected"
+	// ReasonMaintenanceCreated indicates a ServerMaintenance resource has been created.
+	ReasonMaintenanceCreated = "ServerMaintenanceHasBeenCreated"
+	// ReasonMaintenanceDeleted indicates a ServerMaintenance resource has been deleted.
+	ReasonMaintenanceDeleted = "ServerMaintenanceHasBeenDeleted"
+	// ReasonMaintenanceWaiting indicates waiting for ServerMaintenance approval.
+	ReasonMaintenanceWaiting = "ServerMaintenanceWaitingOnApproval"
+	// ReasonMaintenanceApproved indicates ServerMaintenance has been approved.
+	ReasonMaintenanceApproved = "ServerMaintenanceApproval"
+)
```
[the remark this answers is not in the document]
afritzler: Added migration logic for the legacy condition types.

internal/controller/helper.go
```diff
@@ -98,6 +88,20 @@ func GetCondition(acc *conditionutils.Accessor, conditions []metav1.Condition, c
...
 }
 
+// migrateConditionTypes renames legacy condition type strings in-place.
+// Returns true if any conditions were migrated.
+// TODO: Remove this function in the next release once all CRs have been reconciled.
+func migrateConditionTypes(conditions []metav1.Condition, migrations map[string]string) bool {
+	migrated := false
+	for i := range conditions {
+		if newType, ok := migrations[conditions[i].Type]; ok {
+			conditions[i].Type = newType
```
nagadeesh-nagaraja, line 98: I would also update the Reason string here. 
Unfortunately, we use "Reason" for some of the logic. ideally we need to move away from that. 
but, for the current PR for migration.. i would also update the Reason string here. 


one example where i found Reason causing issue is "BMCResetIssued" which was changed 

have a migrationConditionReason  method as well. 
```
var legacyBIOSSettingsConditionReasons = map[string]string{
	"BMCResetIssued": ReasonResetIssued,
}
```

afritzler:
@coderabbitai review

internal/controller/conditions.go
```diff
@@ -0,0 +1,78 @@
...
+	// ReasonUnknownError indicates an unknown error.
+	ReasonUnknownError = "UnknownError"
+	// ReasonConnectionFailed indicates a connection failure.
+	ReasonConnectionFailed = "ConnectionFailed"
+	// ReasonUserReset indicates a user-requested reset.
+	ReasonUserReset = "UserRequested"
+	// ReasonAutoReset indicates an automatic reset.
+	ReasonAutoReset = "AutoResetting"
+	// ReasonConnected indicates a successful connection.
+	ReasonConnected = "BMCConnected"
```
stefanhipfel, line 67: should we also just use `Connected` here and drop BMC? Same as the` ConnectionFailed`

internal/controller/helper.go
```diff
@@ -98,6 +88,20 @@ func GetCondition(acc *conditionutils.Accessor, conditions []metav1.Condition, c
 	return condition, nil
 }
 
+// migrateConditionTypes renames legacy condition type strings in-place.
+// Returns true if any conditions were migrated.
+// TODO: Remove this function in the next release once all CRs have been reconciled.
+func migrateConditionTypes(conditions []metav1.Condition, migrations map[string]string) bool {
```
stefanhipfel, line 94: what would be the easiest way to check if all migrations happened successfully, before removing the migration code? Do we check each region and search for legacy condition type strings?

internal/controller/conditions.go
afritzler: Thanks a lot for the catch @stefanhipfel! I will change that.

internal/controller/helper.go
afritzler: Well ideally you need to check that all regions have the upcoming version v0.5.0 rolled out. Additionally a proper testing in dev/qa needs to be done e.g. by having BIOS/BMC objects present before the update.

afritzler: A kubectl one-liner like:                         
```
kubectl get biossettings -A -o json | jq '[.items[].status.conditions[]?.type] | map(select(. == "BMCResetIssued" or ...))'
```

internal/controller/helper.go
afritzler: Ok, I addressed backward compatibility also for reasons.

stefanhipfel (approved):

afritzler:
@coderabbit add an issue to remove the reason and condition migration logic for the next stable version.

afritzler:
@coderabbitai review

internal/controller/biossettings_controller.go
```diff
@@ -97,9 +114,36 @@ func (r *BIOSSettingsReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		return ctrl.Result{}, fmt.Errorf("failed to migrate legacy conditions: %w", err)
+	}
 	return r.reconcileExists(ctx, biosSettings)
 }
 
+// migrateLegacyConditions renames legacy condition type strings on existing CRs.
+// TODO: Remove this migration in the next release once all CRs have been reconciled.
+func (r *BIOSSettingsReconciler) migrateLegacyConditions(ctx context.Context, settings *metalv1alpha1.BIOSSettings) error {
+	settingsBase := settings.DeepCopy()
+	migrated := migrateConditionTypes(settings.Status.Conditions, legacyBIOSSettingsConditionTypes)
```
nagadeesh-nagaraja, line 137: nit: can simplify this logic?

```
migrated := migrateConditionTypes(settings.Status.Conditions, legacyBIOSSettingsConditionTypes) || migrateConditionReasons(settings.Status.Conditions, legacyBIOSSettingsConditionReasons)
```

internal/controller/biossettings_controller.go
```diff
@@ -97,9 +114,36 @@ func (r *BIOSSettingsReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	if migrateConditionReasons(settings.Status.Conditions, legacyBIOSSettingsConditionReasons) {
+		migrated = true
+	}
+	for i := range settings.Status.FlowState {
+		if migrateConditionTypes(settings.Status.FlowState[i].Conditions, legacyBIOSSettingsConditionTypes) {
+			migrated = true
+		}
+		if migrateConditionReasons(settings.Status.FlowState[i].Conditions, legacyBIOSSettingsConditionReasons) {
+			migrated = true
+		}
```
nagadeesh-nagaraja, line 147: nit: Could be simplified as well?

internal/controller/bmcsettings_controller.go
```diff
@@ -75,6 +86,18 @@ func (r *BMCSettingsReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 	log := ctrl.LoggerFrom(ctx)
+	// TODO: Remove this migration in the next release once all CRs have been reconciled.
+	settingsBase := settings.DeepCopy()
+	migrated := migrateConditionTypes(settings.Status.Conditions, legacyBMCSettingsConditionTypes)
+	if migrateConditionReasons(settings.Status.Conditions, legacyBMCSettingsConditionReasons) {
+		migrated = true
+	}
```
nagadeesh-nagaraja, line 96: nit: Could be simplifed using || operator?

internal/controller/endpoint_controller.go
```diff
@@ -22,7 +22,7 @@ import (
 )
 
 const (
-	EndpointFinalizer = "metal.ironcore.dev/endpoint"
```
nagadeesh-nagaraja, line 25: should we make it non exportable (remove Caps)?

some of the finalizers (from other controller) are used in webhook folder.. would it make sense to leave it all uniform? with this change, some controller finalizer can be used outside, some dont.

nagadeesh-nagaraja (approved):
I think its good to go from my side, few nit and please check on the finalizer variable capitalization making them common across all controllers.

internal/controller/endpoint_controller.go
afritzler: Done.

internal/controller/bmcsettings_controller.go
afritzler: We will remove this block anyways once the migration is done.

internal/controller/biossettings_controller.go
afritzler: Same here.

PR #798, opened 2026-04-10
Add TTL support for temporary BMC users

stefanhipfel:
This commit adds automatic expiration functionality for BMCUser resources, enabling temporary users for debugging, maintenance windows, and temporary access scenarios.

Features:
- Dual expiration approach: TTL (relative) and ExpiresAt (absolute)
- Immutable expiration time (calculated once at creation)
- Automatic cleanup of both K8s objects and BMC accounts
- Warning period with condition changes (Active -> ExpiringSoon -> Expired)

## Summary by CodeRabbit

* **New Features**
  * Temporary BMC users with automatic expiration via TTL or absolute timestamp; status shows controller-calculated expiration.
  * Expiration lifecycle conditions: Active, ExpiringSoon, Expired; pre-expiration warnings and coordinated requeueing.

* **Validation**
  * Server-side validation enforcing TTL vs ExpiresAt mutual exclusivity, TTL limits, and future timestamps; update-time warnings when expiration was already calculated.

* **Documentation**
  * Usage guide, README features entry, and sample manifests illustrating temporary user scenarios and monitoring.


README.md
```diff
@@ -12,6 +12,15 @@ Metal-operator is a project built using Kubebuilder and controller-runtime to fa
...
 
+## Features
+
+- **Automated Server Discovery**: Discover and manage bare metal servers through BMC interfaces
+- **Server Provisioning**: Automated server boot configuration and provisioning
+- **BIOS and BMC Management**: Configure and update BIOS/BMC settings and firmware
+- **Temporary Users with TTL**: Automatic expiration and cleanup of temporary BMC users for debugging and maintenance
+- **Server Claims**: Reserve and claim servers for specific workloads
+- **Server Maintenance**: Coordinate maintenance operations across server fleets
+
```
afritzler, line 23: Can we leave this feature out of the landing page (for now).

docs/usage/bmcuser-ttl.md
```diff
@@ -0,0 +1,185 @@
+# Temporary BMC Users with TTL
```
afritzler, line 1: Can we add the expiration of users to the bmsuser.md?

docs/.vitepress/config.mts
```diff
@@ -84,6 +84,7 @@ export default withMermaid({
         collapsed: false,
         items: [
           { text: 'metalctl', link: '/usage/metalctl' },
+          { text: 'Temporary BMC Users', link: '/usage/bmcuser-ttl' },
```
afritzler, line 87: As mentioned in the actual doc file: I would move the doc to the concept section of BMCUser.

afritzler (changes requested):
I have a couple of concerns with the current expiration approach.

Hard deletion on expiration feels too aggressive. Once the TTL fires, the BMC account is removed from the hardware immediately. If someone is still mid-debugging session, they lose access with no way to recover. And since the expiration is immutable, they can't extend it either. The Expired condition is also effectively unobservable: it gets set and then `r.Delete()` is called in the same reconcile, so by the time any monitoring could react, the object is already gone.

I think a two-phase approach would be safer:
1. On expiration: disable the account on the BMC, keep the K8s object around
2. After a grace period (or manual cleanup): actually delete (I would delegate the cleanup in the first step to somebody else)

This gives operators time to react, makes the `ExpiringSoon` warning actually actionable, and preserves an audit trail of who had access and when.

Why a condition instead of a state? The lifecycle here is linear: `Active -> ExpiringSoon -> Expired`. That's a state machine, not an independent observation about the resource. A `status.state` field would model this more naturally than an `Active` condition that's `True` with reason `ExpiringSoon` and then `False` with reason `Expired`. Conditions are better suited for orthogonal signals (like Ready, Degraded) that can be `true/false` independently of each other.

stefanhipfel:
> I have a couple of concerns with the current expiration approach.
> 
> Hard deletion on expiration feels too aggressive. Once the TTL fires, the BMC account is removed from the hardware immediately. If someone is still mid-debugging session, they lose access with no way to recover. And since the expiration is immutable, they can't extend it either. The Expired condition is also effectively unobservable: it gets set and then `r.Delete()` is called in the same reconcile, so by the time any monitoring could react, the object is already gone.
> 
> I think a two-phase approach would be safer:
> 
> 1. On expiration: disable the account on the BMC, keep the K8s object around
> 2. After a grace period (or manual cleanup): actually delete (I would delegate the cleanup in the first step to somebody else)
> 
> This gives operators time to react, makes the `ExpiringSoon` warning actually actionable, and preserves an audit trail of who had access and when.
> 
> Why a condition instead of a state? The lifecycle here is linear: `Active -> ExpiringSoon -> Expired`. That's a state machine, not an independent observation about the resource. A `status.state` field would model this more naturally than an `Active` condition that's `True` with reason `ExpiringSoon` and then `False` with reason `Expired`. Conditions are better suited for orthogonal signals (like Ready, Degraded) that can be `true/false` independently of each other.


TTL:
I would however expect a Deletion when using TTL, because this is exactly what it means: time to live. I am afraid of having many disabled users in our bmc after a while, that nobody knows about. (And no one dares to delete) Most would just create a new user, once the old one is deleted/disabled anyway. I don't see the benefit of disabling and needing a clean up process. 
As a compromise, I would rather keep the BMCUser CRD around with state expired maybe? But I would prefer the User to be deleted on the BMC.

State:
We discussed state some time ago. I think it should come from the BMC itself, e.g. Active, Disabled etc, hence I decided to use Conditions instead, otherwise we would overwrite them.
I could just remove the conditions. BMC User states could be implemented in a separate PR which fetches the actual state from the BMC.

afritzler:
@stefanhipfel can we close this one as the `BMCUser` resource will move to the maintenance operator?

stefanhipfel:
Will be moved to maintenance operator

PR #812, opened 2026-04-16
Add variable resolution and substitution to BMCSettings controller

### LINKED ISSUE #781: Controller Changes to support dynamic Settings in BMCSettings CRD
Implement Controller code for DynamicSettings in CRD

nagadeesh-nagaraja:
# Proposed Changes

Fixes #781

## Summary by CodeRabbit

* **New Features**
  * Enhanced variable substitution in BMCSettings: fieldRef/configMap/Secret selectors and values support ordered/chained expansion and a literal escape `$$(KEY)`; settings are applied after resolution and BMCSettings requeue when referenced Secrets/ConfigMaps change.

* **Behavior**
  * Reconciliation now awaits convergence (less terminal failure on verify) and emits reduced-setting logging (attribute keys only).

* **Documentation**
  * Clarified substitution semantics, fieldRef string-only constraint, and simplified state/workflow with drift handling.

* **Tests**
  * Added comprehensive tests for resolution, chaining, escapes, and fieldRef behavior.

* **RBAC**
  * Added permission to read ConfigMaps for reconciliation.


internal/controller/bmcsettings_controller.go
```diff
@@ -732,7 +732,14 @@ func (r *BMCSettingsReconciler) handleFailedState(ctx context.Context, settings
 
 func (r *BMCSettingsReconciler) getBMCSettingsDifference(ctx context.Context, settings *metalv1alpha1.BMCSettings, bmcObj *metalv1alpha1.BMC, bmcClient bmc.BMC) (diff schemas.SettingsAttributes, err error) {
```
afritzler, line 734: The controller is missing a proper RBAC for configmap setup and is also missing a watch setup for configmaps/secrets.

internal/controller/helper.go
```diff
@@ -461,3 +463,123 @@ func labelChangeOrAnyFieldChangeInObject(e event.UpdateEvent, oldFields, newFiel
...
+	out := make(map[string]string, len(settingsMap))
+	for k, v := range settingsMap {
+		out[k] = substituteVars(v, resolved)
+	}
+	return out
+}
+
+// substituteVars replaces $(KEY) with the resolved value.
+// The escape sequence $$(KEY) produces a literal $(KEY) in the output.
+func substituteVars(s string, resolved map[string]string) string {
```
afritzler, line 505: The $(VAR) substitution syntax is borrowed from Kubernetes, and Kubernetes already ships a battle-tested, single-pass implementation for exactly this:

`k8s.io/kubernetes/third_party/forked/golang/expansion` - specifically Expand() and MappingFuncFor().

This method becomes essentially:
```go
func substituteVars(s string, resolved map[string]string) string {
   return expansion.Expand(s, expansion.MappingFuncFor(resolved))
}
```
Since pulling from k8s.io/kubernetes directly is too heavy on transitive deps, we can just copy expand.go into an internal/expansion/ package — it's Apache-2.0 and designed to be forked (it's literally in third_party/forked/golang/). We need to adjust here the REUSE.toml to reflect this third party copy.

internal/controller/helper.go
```diff
@@ -461,3 +463,123 @@ func labelChangeOrAnyFieldChangeInObject(e event.UpdateEvent, oldFields, newFiel
...
+		}
+		return val, nil
+	default:
+		return "", fmt.Errorf("no source specified in valueFrom")
+	}
+}
+
+// resolveFieldRef navigates a dotted fieldPath on the given object using the
+// unstructured accessor (e.g. "spec.BMCRef.name").
+func resolveFieldRef(obj client.Object, fieldPath string) (string, error) {
```
afritzler, line 566: This works for string fields, but `unstructured.NestedString` will error on anything that isn't a string (ints, bools, maps). That's probably fine for the current scope, but worth a brief comment or doc note saying fieldRef only supports string-valued fields. If the scope grows later, Kubernetes has `k8s.io/apimachinery/pkg/fieldpath` (ExtractFieldPathAsString) which handles the type coercion and covers the same field paths as the downward API.

docs/concepts/bmcsettings.md
```diff
@@ -71,59 +75,71 @@ stateDiagram-v2
...
+   - Pre-apply BMC reset to establish a stable state.
+   - Issue settings update and track progress via conditions.
+6. **Reboot/verification path** (`InProgress`):
+   - Post-apply BMC reset/reboot when required by vendor behavior.
+   - Re-run variable resolution and diff check to verify convergence.
+   - Transition to `Applied` when diff is empty, otherwise stay `InProgress` and retry.
+7. **Drift detection** (`Applied`):
+   - On each reconcile, re-resolve variables and re-check diff against the BMC.
+   - If drift is detected, transition back to `Pending` for a new apply cycle.
+8. **Terminalisation and cleanup** (`Applied`):
```
afritzler: Typo.

docs/concepts/bmcsettings.md
```diff
@@ -31,12 +31,16 @@ Note: The API field is `BMCRef` (capitalized) in this CRD schema.
 `spec.variables` allows setting values to be resolved dynamically at apply time.
```
afritzler, line 31: The docs say a missing Secret/ConfigMap/field transitions to Failed, but looking at the controller code, ResolveVariables errors bubble up as a reconcile error - which means the controller requeues with backoff rather than moving to the Failed state.

afritzler (changes requested):

third_party/expansion/expand.go
```diff
@@ -0,0 +1,90 @@
+// Originally from https://github.com/kubernetes/kubernetes/blob/master/third_party/forked/golang/expansion/expand.go
```
afritzler, line 1: Should we also include the test file?

internal/controller/bmcsettings_controller.go
```diff
@@ -730,18 +731,36 @@ func (r *BMCSettingsReconciler) handleFailedState(ctx context.Context, settings
 	return nil
 }
 
+// settingKeys returns only the map keys of a SettingsAttributes map for
+// safe logging. Values are omitted because they may contain secrets resolved
+// from SecretKeyRef variables.
+func settingKeys(attrs schemas.SettingsAttributes) []string {
```
afritzler: I would suggest to move this to the helper.go.

internal/controller/helper_test.go
```diff
@@ -0,0 +1,381 @@
...
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	metalv1alpha1 "github.com/ironcore-dev/metal-operator/api/v1alpha1"
+)
+
+var _ = Describe("Variable templating", func() {
+
+	// ─── substituteVars ────────────────────────────────────────────────────────
```
afritzler: Can we remove those comments.

afritzler (changes requested):
Some small nitpicks from my side.

internal/controller/bmcsettings_controller.go
```diff
@@ -730,18 +731,36 @@ func (r *BMCSettingsReconciler) handleFailedState(ctx context.Context, settings
...
+	keys := make([]string, 0, len(attrs))
+	for k := range attrs {
+		keys = append(keys, k)
+	}
+	return keys
+}
```
afritzler: Can you please remove this part as this now went to the helper.go.

afritzler (changes requested):

afritzler:
@coderabbitai review

afritzler (approved):

PR #830, opened 2026-04-23
Rename maintenance approval metadata label 

### LINKED ISSUE #831: Rename ServerMaintenanceApprovalKey to match naming convention
**Context**
The `ServerMaintenance` API in metal-operator defines label keys for maintenance lifecycle states. The naming convention uses past-tense forms describing state: `maintenance-needed`, `maintenance-requested`. The approval key breaks this pattern.

```go
const (
    ServerMaintenanceNeededLabelKey    = "metal.ironcore.dev/maintenance-needed"
    ServerMaintenanceRequestedLabelKey = "metal.ironcore.dev/maintenance-requested"
    ServerMaintenanceApprovalKey       = "metal.ironcore.dev/maintenance-approval"  // inconsistent
)
```

Two inconsistencies:
- Const name missing `Label` suffix and uses noun form (`Approval` vs `Approved`)
- Label value uses `maintenance-approval` (noun/process) instead of `maintenance-approved` (state)

valeryia-hurynovich:
# Proposed Changes

- Constant renamed: `ServerMaintenanceApprovalKey` → `ServerMaintenanceApprovedLabelKey`
- Label value renamed: `metal.ironcore.dev/maintenance-approval `→ `metal.ironcore.dev/maintenance-approved`
- All references updated across metal-operator codebase

Fixes #831

## Summary by CodeRabbit

* **Chores**
  * Added a new maintenance approval metadata label and marked the previous approval metadata key as deprecated.
  * Maintenance teardown now removes the new approval label as part of cleanup.

* **Bug Fixes**
  * Approval handling accepts either the new label or the deprecated key so approval flows remain reliable.

* **Tests**
  * Updated tests to use the new approval label and added coverage verifying backward compatibility with the deprecated key.


api/v1alpha1/servermaintenance_types.go
```diff
@@ -13,8 +13,8 @@ const (
 	ServerMaintenanceNeededLabelKey = "metal.ironcore.dev/maintenance-needed"
 	// ServerMaintenanceReasonAnnotationKey is an annotation key that is used to store the reason for a server maintenance.
 	ServerMaintenanceReasonAnnotationKey = "metal.ironcore.dev/maintenance-reason"
-	// ServerMaintenanceApprovalKey is a label key that is used to store the approval status for a server maintenance.
-	ServerMaintenanceApprovalKey = "metal.ironcore.dev/maintenance-approval"
+	// ServerMaintenanceApprovedLabelKey is a label key that is used to store the approved status for a server maintenance.
+	ServerMaintenanceApprovedLabelKey = "metal.ironcore.dev/maintenance-approved"
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: I would recommend to address this as suggested by AI.

maxmoehl:
Thank you for your contribution, but before implementing changes we ask contributors to first open an issue to outline the reasoning behind a change and discuss possible solutions to it.

internal/controller/servermaintenance_controller_test.go
nagadeesh-nagaraja, line 1: How about we add a test to check the deprecated workflow? 

a example simple test:

create a serverMaintenance with old Keys and then assert that the old keys are automatically removed.
this test will make sure that we will not break the production clusters :)

nagadeesh-nagaraja:
@coderabbitai Can you please create a follow up issue to remove the deprecated keys and logic in the next major release version? 
cc: @valeryia-hurynovich

nagadeesh-nagaraja (commented):
I think the test would be nice here, we can remove the test when we get rid of old labels/keys

nagadeesh-nagaraja:
@coderabbitai 
under the issue:  https://github.com/ironcore-dev/metal-operator/issues/835 create a sub-issue  to create webhook handle to check for the deprecated ServerMaintenance key and reject those updates on ServerClaim. This would help us to throw error if someone is using that key still before removing the deprecated key. Else, if key is removed all of a sudden, it might fail silently which is not nice.

nagadeesh-nagaraja (approved):
I think as next steps we might need to introduce a webhook to reject deprecated key usage, and then later delete the deprecated key. 

I have created issue to track them.

api/v1alpha1/servermaintenance_types.go
```diff
@@ -13,7 +13,10 @@ const (
 	ServerMaintenanceNeededLabelKey = "metal.ironcore.dev/maintenance-needed"
 	// ServerMaintenanceReasonAnnotationKey is an annotation key that is used to store the reason for a server maintenance.
 	ServerMaintenanceReasonAnnotationKey = "metal.ironcore.dev/maintenance-reason"
-	// ServerMaintenanceApprovalKey is a label key that is used to store the approval status for a server maintenance.
+	// ServerMaintenanceApprovedLabelKey is a label key that is used to store the approved status for a server maintenance.
+	ServerMaintenanceApprovedLabelKey = "metal.ironcore.dev/maintenance-approved"
+	// ServerMaintenanceApprovalKey is the previous maintenance approval metadata key.
+	// Deprecated: Use ServerMaintenanceApprovedLabelKey instead. Will be removed in a future release.
```
hardikdr, line 19: Looks like a typo here, `ServerMaintenanceApprovedLabelKey` instead of `ServerMaintenanceApprovalKey`?

valeryia-hurynovich: It looks like everything is right, we will use ServerMaintenanceApprovedLabelKey instead of ServerMaintenanceApprovalKey and in comments it also describes this, doesn't it?

hardikdr: That's true, may bad.

hardikdr (approved):

PR #839, opened 2026-04-27
Add Server Label to metrics

asergeant01:
# Proposed Changes

Update Metal Operator metrics to have server label. 

e.g. 

from:

`metal_server_power_state{endpoint="https", instance="127.0.0.1:8443", job="metal-operator-controller-manager-metrics-service", namespace="my-namespace", pod="metal-operator-controller-manager-5fff57cd7c-nqhwq", power_state="Off", service="metal-operator-controller-manager-metrics-service"}
`

to:

`metal_server_power_state{endpoint="https", instance="127.0.0.1:8443", job="metal-operator-controller-manager-metrics-service", namespace="my-namespace", pod="metal-operator-controller-manager-5fff57cd7c-nqhwq", power_state="Off", server="my-server-1",service="metal-operator-controller-manager-metrics-service"}
`

## Summary by CodeRabbit

* **New Features**
  * Per-server metrics: state, power state, and condition gauges that emit enum-style series with a server label.
  * Added alert rule test/check targets and local promtool/yq install support.
* **Bug Fixes**
  * Fixed PrometheusRule nesting and revised alert expressions/annotations; replaced metric-absence alert with scrape-health check.
* **Tests**
  * Updated unit tests for per-server metrics and added Prometheus alert rule tests.
* **Documentation**
  * Updated metrics docs, PromQL/Grafana examples, and cardinality guidance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,6 +131,16 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 lint-config: golangci-lint ## Verify golangci-lint linter configuration
 	"$(GOLANGCI_LINT)" config verify
 
+.PHONY: test-alerts
+test-alerts: promtool yq ## Run Prometheus alert rule tests.
+	@"$(YQ)" '.spec' config/prometheus/server_alerts.yaml > config/prometheus/server_alerts_rules.yaml
+	"$(PROMTOOL)" test rules config/prometheus/server_alerts_test.yaml
+
+.PHONY: check-alerts
+check-alerts: promtool yq ## Validate Prometheus alert rules syntax.
+	@"$(YQ)" '.spec' config/prometheus/server_alerts.yaml > config/prometheus/server_alerts_rules.yaml
+	"$(PROMTOOL)" check rules config/prometheus/server_alerts_rules.yaml
+
 .PHONY: startdocs
 startdocs: ## Start the local mkdocs based development environment.
 	docker build -t $(IMAGE) -f docs/Dockerfile . --load
@@ -270,6 +280,8 @@ GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 GOIMPORTS ?= $(LOCALBIN)/goimports
 METALCTL ?= $(LOCALBIN)/metalctl
 GINKGO ?= $(LOCALBIN)/ginkgo
+PROMTOOL ?= $(LOCALBIN)/promtool
+YQ ?= $(LOCALBIN)/yq
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.8.1
@@ -285,6 +297,8 @@ CRD_REF_DOCS_VERSION ?= v0.2.0
 KUBEBUILDER_VERSION ?= v4.13.0
 ADDLICENSE_VERSION ?= v1.1.1
 GINKGO_VERSION ?= $(shell go list -m -f '{{.Version}}' github.com/onsi/ginkgo/v2)
+PROMTOOL_VERSION ?= 3.11.3
+YQ_VERSION ?= v4.44.3
 
 .PHONY: ginkgo
 ginkgo: $(GINKGO) ## Download ginkgo locally if necessary.
@@ -359,6 +373,16 @@ $(MOCKSERVER_BIN): $(LOCALBIN)
 .PHONY: metalctl
 metalctl: $(METALCTL) ## Build metalctl locally if necessary.
 	go build -o $(METALCTL) ./cmd/metalctl
+
+.PHONY: promtool
+promtool: $(PROMTOOL) ## Download promtool locally if necessary.
+$(PROMTOOL): $(LOCALBIN)
+	@hack/install-promtool.sh "$(PROMTOOL_VERSION)" "$(LOCALBIN)"
+
+.PHONY: yq
+yq: $(YQ) ## Download yq locally if necessary.
+$(YQ): $(LOCALBIN)
+	$(call go-install-tool,$(YQ),github.com/mikefarah/yq/v4,$(YQ_VERSION))
 
 # go-install-tool will 'go install' any package with custom target and name of binary, if it doesn't exist
 # Note: All paths are quoted to work in directories containing spaces or parentheses.
--- a/hack/install-promtool.sh
+++ b/hack/install-promtool.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and IronCore contributors
+# SPDX-License-Identifier: Apache-2.0
+
+# install-promtool.sh downloads a pre-built promtool binary from the
+# Prometheus GitHub release artifacts with checksum verification.
+#
+# Usage: install-promtool.sh <version> <output-dir>
+#   version    - Prometheus release version (e.g. 2.54.0, without leading 'v')
+#   output-dir - Directory to place the promtool binary in
+
+set -euo pipefail
+
+VERSION="${1:?Usage: install-promtool.sh <version> <output-dir>}"
+OUTPUT_DIR="${2:?Usage: install-promtool.sh <version> <output-dir>}"
+
+# Strip leading 'v' if present
+VERSION="${VERSION#v}"
+
+BINARY="${OUTPUT_DIR}/promtool"
+
+if [[ -f "${BINARY}" ]] && "${BINARY}" --version 2>&1 | grep -q "${VERSION}"; then
+    echo "promtool ${VERSION} already installed at ${BINARY}"
+    exit 0
+fi
+
+OS="$(uname -s | tr '[:upper:]' '[:lower:]')"
+ARCH="$(uname -m)"
+case "${ARCH}" in
+    x86_64)  ARCH="amd64" ;;
+    aarch64) ARCH="arm64" ;;
+esac
+
+TARBALL="prometheus-${VERSION}.${OS}-${ARCH}.tar.gz"
+RELEASE_URL="https://github.com/prometheus/prometheus/releases/download/v${VERSION}"
+
+TMP_DIR="$(mktemp -d)"
+trap 'rm -rf "${TMP_DIR}"' EXIT
+
+echo "Downloading promtool ${VERSION} for ${OS}/${ARCH}..."
+curl -sSfL "${RELEASE_URL}/${TARBALL}" -o "${TMP_DIR}/${TARBALL}"
+curl -sSfL "${RELEASE_URL}/sha256sums.txt" -o "${TMP_DIR}/sha256sums.txt"
+
+cd "${TMP_DIR}"
+if ! grep -F "${TARBALL}" sha256sums.txt | sha256sum -c -; then
+    echo "ERROR: SHA256 checksum verification failed for ${TARBALL}" >&2
+    exit 1
+fi
+
+tar -xzf "${TARBALL}" -C "${TMP_DIR}"
+mv "prometheus-${VERSION}.${OS}-${ARCH}/promtool" "${BINARY}"
+chmod +x "${BINARY}"
+echo "Installed promtool ${VERSION} at ${BINARY}"
--- a/internal/metrics/server_metrics.go
+++ b/internal/metrics/server_metrics.go
@@ -13,6 +13,25 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
+// AllServerStates defines all possible server states for enum metrics
+var AllServerStates = []metalv1alpha1.ServerState{
+	metalv1alpha1.ServerStateInitial,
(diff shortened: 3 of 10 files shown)


internal/metrics/server_metrics.go
```diff
@@ -47,20 +47,20 @@ func NewServerStateCollector(c client.Client) *ServerStateCollector {
...
 		),
 		conditionDesc: prometheus.NewDesc(
 			"metal_server_condition_status",
-			"Count of servers with each condition status",
-			[]string{"condition_type", "status"},
+			"Current condition status of a server (value is always 1)",
+			[]string{"server", "namespace", "condition_type", "status"},
 			nil,
 		),
 	}
```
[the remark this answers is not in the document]
asergeant01: @stefanhipfel As the original creator what are your thoughts on this. I think without the server label the counts aren't that informative - particularly with alerting.

xkonni (approved):

.github/workflows/test-alerts.yml
```diff
@@ -0,0 +1,31 @@
+name: Test Alerts
```
afritzler: Is there a reason why we have an own workflow for testing alerts? Why not including it into `make tests`?

afritzler: Actually I would add `make tests-alerts` to the general test workflow.

afritzler (changes requested):

Makefile
```diff
@@ -135,6 +135,20 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 lint-config: golangci-lint ## Verify golangci-lint linter configuration
 	"$(GOLANGCI_LINT)" config verify
 
+.PHONY: test-alerts
+test-alerts: ## Run Prometheus alert rule tests (requires promtool and yq)
+	@command -v promtool >/dev/null 2>&1 || { echo "promtool not found. Install from https://github.com/prometheus/prometheus/releases"; exit 1; }
```
afritzler: `promtool` should be installed like any other tools - see below in the `Makefile`.

afritzler (changes requested):

Makefile
```diff
@@ -135,6 +135,20 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
...
+	@command -v yq >/dev/null 2>&1 || { echo "yq not found. Install from https://github.com/mikefarah/yq/releases"; exit 1; }
```
afritzler: Same for `yq`.

afritzler (changes requested):

.github/workflows/test-alerts.yml
asergeant01: I was keeping it seperate so it was only run when there where changes to the prometheus configs... I'm not sure of the value in running every time when there are no changes?

Makefile
asergeant01: Done

Makefile
asergeant01: Done

internal/metrics/server_metrics.go
```diff
@@ -90,63 +91,40 @@ func (c *ServerStateCollector) Collect(ch chan<- prometheus.Metric) {
...
-		// Count by power state
+		// Emit per-server power state metric
 		if server.Status.PowerState != "" {
-			powerStateCounts[string(server.Status.PowerState)]++
+			ch <- prometheus.MustNewConstMetric(
+				c.powerStateDesc,
+				prometheus.GaugeValue,
+				1,
+				serverName,
+				string(server.Status.PowerState),
```
stefanhipfel: Adding things like status as labels, makes the metrics messy:
When a server moves from Available to Reserved, you don’t get one series updating — you get the old series going stale and a new series appearing.

One solution would be:
metal_server_state{server="srv-01", state="Available"} 1
metal_server_state{server="srv-01", state="Reserved"} 0
metal_server_state{server="srv-01", state="Error"}    0
metal_server_state{server="srv-01", state="Initial"}  0

stefanhipfel: This way we would not have stale metrics that disappear

asergeant01: You are absolutely right - great catch

Makefile
```diff
@@ -360,6 +374,26 @@ $(MOCKSERVER_BIN): $(LOCALBIN)
...
 
+.PHONY: promtool
+promtool: $(PROMTOOL) ## Download promtool locally if necessary.
+$(PROMTOOL): $(LOCALBIN)
+	@[ -f "$(PROMTOOL)-$(PROMTOOL_VERSION)" ] || { \
+		set -e; \
+		echo "Downloading promtool $(PROMTOOL_VERSION)" ; \
+		rm -f "$(PROMTOOL)" || true ; \
+		TMP_DIR=$$(mktemp -d) ; \
+		curl --retry $(CURL_RETRIES) -fsL "https://github.com/prometheus/prometheus/releases/download/v$(PROMTOOL_VERSION)/prometheus-$(PROMTOOL_VERSION).$(GOOS)-$(GOARCH).tar.gz" | tar -xz -C "$$TMP_DIR" ; \
```
afritzler: I would recommend using the `go-install-tool` helper and use `https://github.com/prometheus/prometheus/tree/main/cmd/promtool` + version. Similar to the `yq` installation below.

afritzler (changes requested):

.github/workflows/test-alerts.yml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  test-alerts:
+    name: Test Prometheus alert rules
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone the code
+        uses: actions/checkout@v6
+
+      - name: Install yq
+        uses: mikefarah/yq@v4
```
afritzler: This is not needed if we just use the `setup-go` action + the `Makefile` target as we install the tools over there.

afritzler (changes requested):

Makefile
asergeant01: Thanks for the suggestion. I tried using `go-install-tool` with `github.com/prometheus/prometheus/cmd/promtool`, but unfortunately `go install` doesn't work for promtool because Prometheus's `go.mod` contains `replace` directives (for k8s logging integration). 

See: https://github.com/prometheus/prometheus/issues/9958 (open since Dec 2021, still unresolved)

The curl-based approach downloading pre-built binaries seems like the right solution for now?

.github/workflows/test-alerts.yml
maxmoehl: The main issue is that workflows which don't always run cannot be made required in the PR checks. The preferred solution for now is to run them always even if we waste a few CI minutes for every PR that way.

afritzler: Agree with @maxmoehl on that one. Lets move that to the `make test` flow.

Makefile
```diff
@@ -360,6 +374,35 @@ $(MOCKSERVER_BIN): $(LOCALBIN)
...
+			echo "ERROR: SHA256 checksum verification failed for $${TARBALL}" ; \
+			rm -rf "$$TMP_DIR" ; \
+			exit 1 ; \
+		} ; \
+		tar -xzf "$$TMP_DIR/$${TARBALL}" -C "$$TMP_DIR" ; \
+		mv "$$TMP_DIR/prometheus-$(PROMTOOL_VERSION).$(GOOS)-$(GOARCH)/promtool" "$(PROMTOOL)-$(PROMTOOL_VERSION)" ; \
+		rm -rf "$$TMP_DIR" ; \
+	}
+	ln -sf "$(PROMTOOL)-$(PROMTOOL_VERSION)" "$(PROMTOOL)"
+	chmod +x "$(PROMTOOL)"
```
afritzler: @coderabbitai how would you install this tool similar to other tools in this makefile using the go-install-tool helper?

afritzler (changes requested):

Makefile
asergeant01: @coderabbitai I tested your suggestion with v0.54.0 but it still fails with the same replace directives error:

The issue is that all Prometheus module versions have replace directives in their go.mod (for k8s logging integration), not just the v2.x tags. This is a fundamental limitation documented in https://github.com/prometheus/prometheus/issues/9958 (open since Dec 2021, still unresolved as of Nov 2025).

.github/workflows/test-alerts.yml
asergeant01: ok

Makefile
afritzler: Thanks @asergeant01 for clarifying that. Then maybe it is better to have a `hack/install-promtool.sh` helper instead of bloating the `Makefile`. Wdyt?

afritzler (changes requested):

Makefile
afritzler: We can the call the installer via 
```shell
 @hack/install-promtool.sh "$(PROMTOOL_VERSION)" "$(LOCALBIN)"
```

Makefile
```diff
@@ -131,6 +131,16 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 lint-config: golangci-lint ## Verify golangci-lint linter configuration
 	"$(GOLANGCI_LINT)" config verify
 
+.PHONY: test-alerts
+test-alerts: promtool yq ## Run Prometheus alert rule tests.
+	@"$(YQ)" '.spec' config/prometheus/server_alerts.yaml > config/prometheus/server_alerts_rules.yaml
+	cd config/prometheus && "$(PROMTOOL)" test rules server_alerts_test.yaml
```
afritzler:
```suggestion
	"$(PROMTOOL)" test rules config/prometheus/server_alerts_test.yaml
```

Makefile
```diff
@@ -360,6 +374,16 @@ $(MOCKSERVER_BIN): $(LOCALBIN)
 metalctl: $(METALCTL) ## Build metalctl locally if necessary.
 	go build -o $(METALCTL) ./cmd/metalctl
 
+.PHONY: promtool
+promtool: $(PROMTOOL) ## Download promtool locally if necessary.
+$(PROMTOOL): $(LOCALBIN)
+	@hack/install-promtool.sh "$(PROMTOOL_VERSION)" "$(LOCALBIN)" "$(CURL_RETRIES)"
```
afritzler: I would frankly remove the retry logic.

hack/install-promtool.sh
```diff
@@ -0,0 +1,47 @@
...
+if ! grep -F "${TARBALL}" sha256sums.txt | sha256sum -c -; then
+    echo "ERROR: SHA256 checksum verification failed for ${TARBALL}" >&2
+    exit 1
+fi
+
+tar -xzf "${TARBALL}" -C "${TMP_DIR}"
+mv "prometheus-${VERSION}.${GOOS}-${GOARCH}/promtool" "${PROMTOOL}-${VERSION}"
+
+ln -sf "${PROMTOOL}-${VERSION}" "${PROMTOOL}"
+chmod +x "${PROMTOOL}"
```
afritzler: Removing the retry logic, but adding multi arch fetching:

```shell
#!/usr/bin/env bash
set -euo pipefail

# install-promtool.sh downloads a pre-built promtool binary from the
# Prometheus GitHub release artifacts.
#
# Usage: install-promtool.sh <version> <output-dir>
#   version    - Prometheus release version (e.g. 3.11.3, without leading 'v')
#   output-dir - Directory to place the promtool binary in

VERSION="${1:?Usage: install-promtool.sh <version> <output-dir>}"
OUTPUT_DIR="${2:?Usage: install-promtool.sh <version> <output-dir>}"

# Strip leading 'v' if present
VERSION="${VERSION#v}"

BINARY="${OUTPUT_DIR}/promtool"

if [ -f "${BINARY}" ] && "${BINARY}" --version 2>&1 | grep -q "${VERSION}"; then
    echo "promtool ${VERSION} already installed at ${BINARY}"
    exit 0
fi

OS="$(uname -s | tr '[:upper:]' '[:lower:]')"
ARCH="$(uname -m)"
case "${ARCH}" in
    x86_64)  ARCH="amd64" ;;
    aarch64) ARCH="arm64" ;;
esac

TARBALL="prometheus-${VERSION}.${OS}-${ARCH}.tar.gz"
URL="https://github.com/prometheus/prometheus/releases/download/v${VERSION}/${TARBALL}"

TMP="$(mktemp -d)"
trap 'rm -rf "${TMP}"' EXIT

echo "Downloading promtool ${VERSION} for ${OS}/${ARCH}..."
curl -sSfL "${URL}" -o "${TMP}/${TARBALL}"
tar -xzf "${TMP}/${TARBALL}" -C "${TMP}" --strip-components=1
mv "${TMP}/promtool" "${BINARY}"
chmod +x "${BINARY}"
echo "Installed promtool at ${BINARY}"

```

afritzler (changes requested):

afritzler (approved):

PR #846, opened 2026-04-28
Enh/fallback allowable values simpleupdate

davidgrun:
## Fallback transfer protocol for firmware updates

Some BMC implementations only support a subset of transfer protocols for the Redfish `SimpleUpdate` action. When a firmware image URI is configured with a protocol the BMC doesn't support, the update would silently fail or error with no clear recovery path.

This PR adds a fallback mechanism to `ImageSpec`: alongside the existing `transferProtocol` and `uri` fields, operators can now configure `fallbackTransferProtocol` and `fallbackURI`. Before issuing a firmware update, the controller queries the BMC's `UpdateService` for its `TransferProtocol@Redfish.AllowableValues` list and selects which protocol to use. The BMC then fetches the firmware image directly from the resolved URI.

Protocol resolution:

- If the BMC reports no restrictions, the primary protocol is used as-is.
- If the primary protocol is supported, it is used.
- If not, and a fallback is configured and supported, the fallback is used.
- If neither is supported, reconciliation fails with a clear error.

This applies to both `BMCVersion` and `BIOSVersion` controllers. CEL validation rules enforce that `fallbackTransferProtocol` and `fallbackURI` are always set together and non-empty, and that a fallback requires a non-empty primary protocol to be set.

## Example

```yaml
kind: BMCVersion
metadata:
  name: my-bmc-update
spec:
  bmcRef:
    name: my-bmc
  bmcVersionTemplate:
    version: "1.2.3"
    image:
      uri: "https://firmware.example.com/bmc-v1.2.3.bin"
      transferProtocol: "HTTPS"
      fallbackTransferProtocol: "TFTP"
      fallbackURI: "tftp://firmware-internal.example.com/bmc-v1.2.3.bin"
```

The fallback fields are optional — if omitted, the controller behaves exactly as before.

### DIFF
--- a/api/v1alpha1/biosversion_types.go
+++ b/api/v1alpha1/biosversion_types.go
@@ -65,6 +65,8 @@ type BIOSVersionSpec struct {
 	ServerRef *corev1.LocalObjectReference `json:"serverRef,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="!has(self.fallbackTransferProtocol) || (has(self.transferProtocol) && size(self.transferProtocol) > 0)",message="fallbackTransferProtocol requires a non-empty transferProtocol"
+// +kubebuilder:validation:XValidation:rule="(has(self.fallbackTransferProtocol) && has(self.fallbackURI)) || (!has(self.fallbackTransferProtocol) && !has(self.fallbackURI))",message="fallbackTransferProtocol and fallbackURI must both be set or both be unset"
 type ImageSpec struct {
 	// SecretRef is a reference to the Secret containing the credentials to access the image URI.
 	// +optional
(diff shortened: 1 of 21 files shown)


nagadeesh-nagaraja:
do we happen to know the reason behind this change? Can not find issue or a description.

nagadeesh-nagaraja:
@coderabbitAI what do you think is most k8 native way to solve the problem this PR is addressing? ( lets say 90% of the server supported all protocol, 10% support subset of it. Hence the need of fallback implemented here) 

Does it make sense to implement it this way? 
or should we handle it creating multiple BIOSSettingsSet CRD with appropriate labels to identify the vendor/mode/BMC which does support all protocol and then create SET CRD for each of these with respective supported protocols?

afritzler:
@coderabbitai please create an issue for this feature in https://github.com/ironcore-dev/metal-maintenance-operator

afritzler:
Closing this PR as this feature will be implemented in the maintenance operator.

/ref https://github.com/ironcore-dev/metal-maintenance-operator/issues/138

PR #848, opened 2026-04-29
Implement webhook validation to reject ServerClaim deprecated updates

### LINKED ISSUE #838: Add webhook validation to reject ServerClaim updates using deprecated maintenance-approval label key
## Summary

This is a sub-issue of #835 ("Remove deprecated `ServerMaintenanceApprovalKey` and backward-compatibility logic").

Before removing the deprecated `ServerMaintenanceApprovalKey` constant (value: `metal.ironcore.dev/maintenance-approval`) in the next major release, a webhook validation should be added to **explicitly reject** any `ServerClaim` create/update that still carries the old label or annotation key. This ensures consumers receive a clear, actionable error rather than experiencing silent failures when the key is eventually removed.

## Problem

When `ServerMaintenanceApprovalKey` (`metal.ironcore.dev/maintenance-approval`) is removed without a webhook guard, any consumer still writing the old key will silently have their approval ignored — the maintenance will never proceed and no error will surface.

## Proposed Solution

Implement a validating admission webhook for `ServerClaim` that:

- On **create** and **update** operations, inspects `metadata.annotations` and `metadata.labels`.
- If either contains the deprecated key `metal.ironcore.dev/maintenance-approval`, **reject** the request with a descriptive error message, e.g.:
  > `ServerClaim uses deprecated maintenance approval key "metal.ironcore.dev/maintenance-approval"; please migrate to "metal.ironcore.dev/maintenance-approved"`

## Tasks

- [ ] Add a `ValidateCreate` / `ValidateUpdate` check in the `ServerClaim` webhook (or create a new webhook if one does not already exist) that detects the presence of `metal.ironcore.dev/maintenance-approval` in annotations and/or labels
- [ ] Return a clear validation error directing users to migrate to `metal.ironcore.dev/maintenance-approved`
- [ ] Add unit/integration tests for the webhook validation
- [ ] Update documentation to mention the webhook guard and migration path

## Context

- Deprecated constant: `ServerMaintenanceApprovalKey` = `metal.ironcore.dev/maintenance-approval` (in `api/v1alpha1/servermaintenance_types.go`)
- New constant: `ServerMaintenanceApprovedLabelKey` = `metal.ironcore.dev/maintenance-approved`
- Introduced/renamed in: PR `#830`
- Deprecation tracking issue: `#831`
- Parent cleanup issue: `#835`
- Requested by: `@nagadeesh-nagaraja`

## Notes

This webhook guard should be shipped **before** the removal of the deprecated constant (tracked in #835) so that consumers have a clear migration window with explicit feedback.

valeryia-hurynovich:
# Proposed Changes

- Created a new webhook that detects the presence of metal.ironcore.dev/maintenance-approval in annotations and labels
- Returned a clear validation error directing users to migrate to metal.ironcore.dev/maintenance-approved
- Added unit/integration tests for the webhook validation

Fixes #838

## Summary by CodeRabbit

* **New Features**
  * ServerClaim resources now validate and reject deprecated approval keys during creation and updates.

* **Tests**
  * Added comprehensive test coverage for ServerClaim webhook validation scenarios.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -650,6 +650,13 @@ func main() { // nolint: gocyclo
 	}
 	// nolint:goconst
 	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
+		if err := webhookv1alpha1.SetupServerClaimWebhookWithManager(mgr); err != nil {
+			setupLog.Error(err, "Failed to create webhook", "webhook", "ServerClaim")
+			os.Exit(1)
+		}
+	}
+	// nolint:goconst
+	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
 		if err := webhookv1alpha1.SetupBIOSSettingsWebhookWithManager(mgr); err != nil {
 			setupLog.Error(err, "Failed to create webhook", "webhook", "BIOSSettings")
 			os.Exit(1)
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -148,3 +148,23 @@ webhooks:
     resources:
     - servers
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /validate-metal-ironcore-dev-v1alpha1-serverclaim
+  failurePolicy: Fail
+  name: vserverclaim-v1alpha1.kb.io
+  rules:
+  - apiGroups:
+    - metal.ironcore.dev
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - serverclaims
+  sideEffects: None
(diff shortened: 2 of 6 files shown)


afritzler:
I don't think we need a webhook for this.

A webhook adds real operational complexity (certs, availability, fail-open vs fail-closed decisions) just to tell someone they're using an old annotation key. That's a lot of infrastructure for a deprecation notice.

Also, rejecting resource creation because of a specific annotation is kind of an anti-pattern. Annotations and labels are user-owned metadata. Controllers should ignore what they don't recognize, not block the entire request.

The "silent failure" concern isn't really a thing either. Once we remove the old key from the controller, it just becomes inert metadata. The maintenance won't proceed, the user notices, checks the changelog. That's how every Kubernetes deprecation works.

If we want to be helpful during the migration window, we can emit a warning Event or set a condition on the ServerClaim when the controller sees the old key during reconciliation. That gives users clear feedback without putting a new failure mode in the admission path.

afritzler:
If we want to prohibit the creation of objects having the wrong keys, we should deny this via a policy. So https://kyverno.io/ e.g. would be an example on how to enforce those kind of policies. That way you can decouple those concerns.

PR #849, opened 2026-04-29
Improved handling of BMC attribute values by consulting registry type definitions

### LINKED ISSUE #847: Improved handling of BMC attribute values by consulting registry type definitions
**Which component is affected?**
BMCSettings

**Describe the bug**
BMC controller fails to apply Integer-type attributes because it doesn't convert string values to integers before sending them to iDRAC.

What Happens:
CRD Schema: BMCSettings uses [map[string]string](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) - all values must be strings

Dell Registry: Dell ManagerAttributeRegistry defines ADGroup.X.Privilege as type Integer

Controller Behavior: SetBMCAttributesImmediately() (line 335) copies string values directly to PATCH payload without checking attribute types or converting

Validation Failure: CheckBMCAttributes() validates types and rejects strings for Integer attributes:

Impact:
Critical - Blocks configuration of all Integer-type Dell iDRAC attributes including:

ADGroup.X.Privilege (0-511 for AD group permissions)
ActiveDirectory.1.AuthTimeout
NTP, SysLog timeout values
Any other Integer-type attributes in Dell registry
Note: Only affects settings that are changing - if the value already matches current iDRAC state, no PATCH is sent and no error occurs.

atd9876:
Fix Dell GetBMCAttributeValues: convert Integer attributes from float64 to int

JSON numbers unmarshaled into map[string]any always produce float64 in Go.
When the registry declares an attribute as IntegerAttributeType, convert
the float64 value to int so downstream checkAttributes validation passes.

# Proposed Changes

- Convert value type to match Dell Registry Attribute type

Fixes #847

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved attribute handling: integer attributes from BMC responses are now normalized to proper integer types, enumeration attributes continue to map correctly, and attribute-type branching is more robust—reducing misinterpreted values and improving configuration accuracy.

### DIFF
--- a/bmc/redfish_dell.go
+++ b/bmc/redfish_dell.go
@@ -248,7 +248,8 @@ func (r *DellRedfishBMC) GetBMCAttributeValues(ctx context.Context, bmcUUID stri
 				continue
 			}
 		}
-		if strings.EqualFold(string(entry.Type), string(schemas.EnumerationAttributeType)) {
+		switch entry.Type {
+		case schemas.EnumerationAttributeType:
 			currentVal, hasCurrentVal := mergedBMCAttributes[name]
 			if !hasCurrentVal {
 				errs = append(errs, fmt.Errorf("enum attribute '%v' not found in any DellAttributes endpoint", name))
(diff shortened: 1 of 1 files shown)


nagadeesh-nagaraja:
@atd9876 Can we move this to draft until we figure out the exact root cause?

nagadeesh-nagaraja (approved):

bmc/redfish_dell.go
```diff
@@ -270,7 +270,19 @@ func (r *DellRedfishBMC) GetBMCAttributeValues(ctx context.Context, bmcUUID stri
 						currentVal, name, entry.Value))
 			}
 		} else {
-			result[name] = mergedBMCAttributes[name]
```
nagadeesh-nagaraja, line 273: nit: might be the right time to convert to switch case right here looks like we are doing the same anyway

```
switch entry.Type {

case schemas.IntegerAttributeType:
    ....
case schemas.EnumerationAttributeType:
    ....
}
```

atd9876: Yes. that will tidy that section of code

PR #851, opened 2026-05-04
Add ServerMaintenance validating webhook with tests and manifests

### LINKED ISSUE #629: Validate `serverRef` on `serverMaintenances` on creation
# Summary

Forbid creation of `ServerMaintenances` targeting a non-existent `Server` object via a validation webhook.

## Motivation

Warn operators early when applying an invalid configuration.

Pooja-Sangle:
# Proposed Changes

- Introduced a validating admission webhook for ServerMaintenance
- Enforced validation to ensure spec.serverRef.name is provided
- Added check to verify that the referenced Server resource exists
- Integrated webhook setup with controller manager
- Updated webhook manifests and Helm chart configuration
- Added unit tests for success and failure cases

Fixes #629 

## Release Notes

* **New Features**
  * Added validation webhook for ServerMaintenance resources that enforces server reference validation during creation and updates.
  * Webhook behavior can be controlled via the `ENABLE_WEBHOOKS` environment variable.

* **Tests**
  * Added comprehensive test suite for ServerMaintenance webhook validation.

### DIFF
--- a/internal/webhook/v1alpha1/servermaintenance_webhook.go
+++ b/internal/webhook/v1alpha1/servermaintenance_webhook.go
@@ -1,0 +1,96 @@
+// SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and IronCore contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha1
+
+import (
+	"context"
+	"fmt"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	metalv1alpha1 "github.com/ironcore-dev/metal-operator/api/v1alpha1"
+)
+
+// -------------------------
+// Setup webhook registration
+// -------------------------
+
+func SetupServerMaintenanceWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr, &metalv1alpha1.ServerMaintenance{}).
+		WithValidator(&ServerMaintenanceValidator{Client: mgr.GetClient()}).
+		Complete()
+}
+
+// +kubebuilder:webhook:path=/validate-metal-ironcore-dev-v1alpha1-servermaintenance,mutating=false,failurePolicy=fail,sideEffects=None,groups=metal.ironcore.dev,resources=servermaintenances,verbs=create;update,versions=v1alpha1,name=vservermaintenance-v1alpha1.kb.io,admissionReviewVersions=v1
+
+// -------------------------
+// Validator struct
+// -------------------------
+
+type ServerMaintenanceValidator struct {
+	Client client.Client
+}
+
+// -------------------------
+// CREATE validation
+// -------------------------
+
+func (v *ServerMaintenanceValidator) ValidateCreate(
+	ctx context.Context,
+	obj *metalv1alpha1.ServerMaintenance,
+) (admission.Warnings, error) {
+
+	// 1. ServerRef must exist
+	if obj.Spec.ServerRef == nil || obj.Spec.ServerRef.Name == "" {
+		return nil, fmt.Errorf("serverRef.name must be set")
+	}
+
+	// 2. Check that referenced Server exists
+	server := &metalv1alpha1.Server{}
+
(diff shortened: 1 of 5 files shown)


internal/webhook/v1alpha1/servermaintenance_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+    Name: obj.Spec.ServerRef.Name,
+	}, server)
+
+
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("referenced Server %q does not exist", obj.Spec.ServerRef.Name)
+		}
+		return nil, err
+	}
```
nagadeesh-nagaraja, line 67: What if there is a transient error or cache not updated on the newly added server? 

retry will solve the issue, but its not a pleasant experience. 


@coderabbitai what is a generic approach k8 uses for handling checking for referenced object before creating the current object patterns?

internal/webhook/v1alpha1/servermaintenance_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+// UPDATE validation
+// -------------------------
+
+func (v *ServerMaintenanceValidator) ValidateUpdate(
+	ctx context.Context,
+	oldObj, newObj *metalv1alpha1.ServerMaintenance,
+) (admission.Warnings, error) {
+
+	// reuse create logic
+	return v.ValidateCreate(ctx, newObj)
```
nagadeesh-nagaraja, line 82: nitpick: might be better to create a common function and use in both ValidateUpdate and ValidateCreate rather than calling ValidateCreate from Validate update..

internal/webhook/v1alpha1/servermaintenance_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		return nil, err
+	}
+
+	return nil, nil
+}
+
+// -------------------------
+// UPDATE validation
+// -------------------------
```
nagadeesh-nagaraja, line 74: nit: these comments could be removed.

internal/webhook/v1alpha1/servermaintenance_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+// CREATE validation
+// -------------------------
+
+func (v *ServerMaintenanceValidator) ValidateCreate(
+	ctx context.Context,
+	obj *metalv1alpha1.ServerMaintenance,
+) (admission.Warnings, error) {
+
+	// 1. ServerRef must exist
+	if obj.Spec.ServerRef == nil || obj.Spec.ServerRef.Name == "" {
```
nagadeesh-nagaraja, line 50: isnt this already done using kubebuilder logic in api definition?

xkonni: while `serverRef` itself is required, the name has no `minLengh` or `required` marker and can be "". the 2nd condition can actually trigger.

afritzler:
Hi @Pooja-Sangle, thanks a lot for taking this on! 

After looking through the changes, I think we can actually solve this without a webhook, and I wanted to share the reasoning in case it's helpful:

**1. `serverRef` is already required**

In `api/v1alpha1/servermaintenance_types.go`, `ServerRef` is already marked `+required` with no `omitempty`, so the CRD schema rejects objects missing the field. The only gap is an empty `Name`, which we can close declaratively right on the field. No webhook needed:

```go
// +kubebuilder:validation:Required
// +kubebuilder:validation:XValidation:rule="self.name != ''",message="serverRef.name must be set"
ServerRef *corev1.LocalObjectReference `json:"serverRef"`
```

**2. Checking that the referenced `Server` exists is a bit of an anti-pattern in Kubernetes**

The API is eventually consistent, so cross-object existence checks at admission time tend to cause more problems than they solve:

- GitOps/kustomize often applies a `Server` and a `ServerMaintenance` together. Apply order isn't guaranteed, so a perfectly valid bundle could get rejected.
- The webhook read isn't transactional with the create that follows, so it's inherently racy.
- If the `Server` shows up a moment later, the controller would have handled it naturally via requeue anyway.

The idiomatic place for this is the reconciler: if the referenced `Server` isn't found, requeue and surface it via a status condition. That way the relationship resolves itself as resources converge, which matches how the rest of the operator works.

**Suggestion**

Would you be up for closing this PR and instead opening a small one that adds the `XValidation` marker to `ServerRef`? And if the controller doesn't already handle a missing `Server` gracefully, that'd be a great follow-up too. Happy to help review either way!

Thanks again for the contribution and for digging into #629 — really appreciate it. 

@coderabbitai can you please create an issue for the outlined problem here.

PR #861, opened 2026-05-06
Rework condition handling during settings application and BMC reset

### LINKED ISSUE #857: Bug: ConditionBMCResetPostSettingApply never set for HPE — causes infinite re-apply loop
## Which component is affected?

BMCSettings controller (`internal/controller/bmcsettings_controller.go` — `updateSettingsAndVerify()`) and HPE BMC implementation (`bmc/redfish_hpe.go` — `CheckBMCAttributes()`).

## Describe the bug

After the BMCSettings controller successfully applies settings to an HPE iLO BMC, it calls `CheckBMCAttributes()` to determine whether a post-apply BMC reset is needed. The HPE implementation unconditionally returns `(false, nil)` — meaning "no reset required." When no reset is required, the code returns `nil` without setting `ConditionBMCResetPostSettingApply`. On the next reconcile (~2 minutes later), `updateSettingsAndVerify()` fetches this condition, finds it unset (empty Reason), and re-enters the apply branch — re-applying all settings unconditionally. This creates an infinite re-apply loop for any HPE BMC.

## To reproduce

1. Apply a `BMCSettingsSet` targeting an HPE iLO BMC with any settings key (PATCH or POST)
2. Wait for the `BMCSettings` child to enter `InProgress` and apply settings successfully
3. Observe that the BMCSettings never transitions to `Applied` — it stays `InProgress` forever
4. Check controller logs — settings are re-issued every ~2 minutes

<details>
<summary>Resource definitions used</summary>

```yaml
kind: BMCSettingsSet
metadata:
  name: ad-ldap-hpe-ilo
spec:
  clusterScope: true
  bmcSelector:
    matchLabels:
      manufacturer: hpe
  template:
    version: "1.68"
    serverMaintenancePolicy: Enforced
    settingsMap:
      "PATCH /redfish/v1/AccountService": |
        {
          "ActiveDirectory": {
            "ServiceEnabled": true,
            "ServiceAddresses": ["ldaps.example.com"]
          }
        }
```

</details>

## Expected behavior

After settings are applied and `CheckBMCAttributes()` returns `(false, nil)` (no reset needed), the controller should record that the apply phase is complete — either by setting `ConditionBMCResetPostSettingApply` with a "no reset needed" reason or by using a different mechanism to track that settings have been issued. The controller should then proceed directly to verification and eventually transition to `Applied`.

## Actual behavior

The controller enters an infinite re-apply loop:

1. Settings applied successfully via `SetBMCAttributesImmediately()`
2. `CheckBMCAttributes()` returns `(false, nil)` for HPE — no reset required
3. `resetBMCReq` is `false` → the `if resetBMCReq { handleBMCReset(...) }` block is skipped
4. `ConditionBMCResetPostSettingApply` is never written to BMCSettings status
5. Function returns `nil` — reconcile ends
6. Next reconcile: `GetCondition(ConditionBMCResetPostSettingApply)` finds nothing in status → creates a fresh in-memory condition with `Reason: ""`
7. Check at line 432: `resetBMC.Reason != ReasonResetIssued` → `"" != "ResetIssued"` → `true`
8. Re-enters the apply branch → re-applies ALL settings again
9. Repeat indefinitely

For PATCH keys this is wasteful but harmless (idempotent). For POST keys this creates duplicates (see Issue 4).

<details>
<summary>Controller logs</summary>

```
# Settings issued — repeats every ~2 minutes:
{"level":"debug","msg":"BMC settings issued successfully","SettingKeys":["PATCH /redfish/v1/AccountService"]}
{"level":"debug","msg":"BMC settings issued successfully","SettingKeys":["PATCH /redfish/v1/AccountService"]}
{"level":"debug","msg":"BMC settings issued successfully","SettingKeys":["PATCH /redfish/v1/AccountService"]}

# Never see "BMC settings changes have been verified" or state transition to Applied
```

</details>

## Environment

- metal-operator version/commit: main (pre-PR-812)
- Kubernetes version: v1.30
- Deployment method: Helm
- Hardware / BMC type: HPE iLO 6 firmware v1.68 (also affects Lenovo XClarity which has the same `return false, nil` implementation)

## Additional context

The relevant code path in `updateSettingsAndVerify()`:

```go
// Line 432 — entry gate
if resetBMC.Reason != ReasonResetIssued {
    // ... applies settings ...

    if resetBMCReq {  // FALSE for HPE/Lenovo
        if ok, err := r.handleBMCReset(ctx, settings, bmcObj, ConditionBMCResetPostSettingApply); !ok || err != nil {
            return ctrl.Result{}, err
        }
    }
    // When resetBMCReq is false, nothing sets ConditionBMCResetPostSettingApply
    // Falls through to verification, but on next reconcile the gate check fails again
}
```

The `!needsReset` path needs to record that the apply phase completed. Possible fixes:

- **Option A**: Set `ConditionBMCResetPostSettingApply` with a new reason like `ReasonNoResetNeeded` when `resetBMCReq` is false
- **Option B**: Change the gate check to look at `ConditionBMCSettingsChangesIssued` (which IS set for all vendors) instead of `ConditionBMCResetPostSettingApply`
- **Option C**: Skip directly to verification after applying when no reset is required, without returning nil

This issue affects ALL HPE and Lenovo BMCs managed by the BMCSettings controller. Dell is not affected because its `CheckBMCAttributes()` can return `true`, triggering `handleBMCReset` which sets the condition.

atd9876:
# Proposed Changes

- Change condition handling to prevent apply loop

Fixes #857

## Summary by CodeRabbit

* **Bug Fixes**
  * Better drift detection: when settings diverge, prior apply state is cleared so reconciliation reliably retries.
  * Improved reset handling: post-apply reset needs are recorded and unnecessary reset attempts are avoided.

* **Improvements**
  * Phased apply-and-verify flow with clearer verification-pending transitions.
  * Status updates now include generation info when applicable.
  * Added a distinct reset-required reason for clearer reset state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ironcore-dev/metal-operator/pull/861)

### DIFF
--- a/internal/controller/bmcsettings_controller.go
+++ b/internal/controller/bmcsettings_controller.go
@@ -423,13 +423,15 @@ func (r *BMCSettingsReconciler) handleSettingInProgressState(ctx context.Context
 
 func (r *BMCSettingsReconciler) updateSettingsAndVerify(ctx context.Context, settings *metalv1alpha1.BMCSettings, bmcObj *metalv1alpha1.BMC, settingsDiff schemas.SettingsAttributes, bmcClient bmc.BMC) (ctrl.Result, error) {
 	log := ctrl.LoggerFrom(ctx)
-	resetBMC, err := GetCondition(r.Conditions, settings.Status.Conditions, ConditionBMCResetPostSettingApply)
+
+	// Phase 1: Issue settings to the BMC (gated by ConditionBMCSettingsChangesIssued)
+	changesIssued, err := GetCondition(r.Conditions, settings.Status.Conditions, ConditionBMCSettingsChangesIssued)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to get condition for reset of BMC of server: %w", err)
+		return ctrl.Result{}, fmt.Errorf("failed to get condition for settings changes issued: %w", err)
 	}
 
-	if resetBMC.Reason != ReasonResetIssued {
-		// apply the BMC Settings if not done.
+	if changesIssued.Reason != ReasonBMCSettingsChangesIssued {
+		// Settings not yet issued — apply them now.
 		if ok, err := r.handleBMCPowerState(ctx, bmcObj, settings); err != nil || ok {
 			return ctrl.Result{}, err
 		}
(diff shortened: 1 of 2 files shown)


atd9876:
@coderabbitai review

internal/controller/bmcsettings_controller.go
```diff
@@ -471,34 +471,55 @@ func (r *BMCSettingsReconciler) updateSettingsAndVerify(ctx context.Context, set
...
+	resetCond, err := GetCondition(r.Conditions, settings.Status.Conditions, ConditionBMCResetPostSettingApply)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get condition for reset of BMC: %w", err)
+	}
+	if resetCond.Reason == ReasonResetRequired || resetCond.Reason == ReasonResetIssued {
+		log.V(1).Info("Handling BMC reset post applying BMC settings")
 		if ok, err := r.handleBMCReset(ctx, settings, bmcObj, ConditionBMCResetPostSettingApply); !ok || err != nil {
 			return ctrl.Result{}, err
 		}
 	}
```
nagadeesh-nagaraja: I think this if condition here is not needed...`handleBMCReset` should handle it (as we already checking and reseting or waiting on the BMC obj. 

also I think handleBMCReset needs to be updated to work with `ReasonResetRequired` Reason.

internal/controller/bmcsettings_controller.go
```diff
@@ -546,7 +567,36 @@ func (r *BMCSettingsReconciler) handleSettingAppliedState(ctx context.Context, s
 		return fmt.Errorf("failed to fetch and check BMCSettings: %w", err)
 	}
 	if len(settingsDiff) > 0 {
-		return r.updateBMCSettingsStatus(ctx, settings, "", nil)
+		// Drift detected — clear the "changes issued" and "changes verified" gates
+		// so the next InProgress reconcile will re-enter the apply phase.
```
nagadeesh-nagaraja: is this extra work needed?
when the this is hit r.updateBMCSettingsStatus(ctx, settings, "", nil) --> resets all condition anyway. 

if manually resetting conditions, you might need to reset all conditions here else might have issues with reconcile :)

internal/controller/bmcsettings_controller.go
atd9876: Yes. will move into handleBMCReset

internal/controller/bmcsettings_controller.go
atd9876: Yes. that is better

atd9876:
@coderabbitai review

internal/controller/bmcsettings_controller.go
```diff
@@ -471,34 +471,32 @@ func (r *BMCSettingsReconciler) updateSettingsAndVerify(ctx context.Context, set
...
-		log.V(1).Info("Waiting for BMC reset post applying BMC settings")
-		if ok, err := r.handleBMCReset(ctx, settings, bmcObj, ConditionBMCResetPostSettingApply); !ok || err != nil {
-			return ctrl.Result{}, err
-		}
+
+		return ctrl.Result{RequeueAfter: r.ResyncInterval}, nil
+	}
+
+	// Phase 2: Handle BMC reset if required
+	log.V(1).Info("Handling BMC reset post applying BMC settings")
```
xkonni: this log fires even if `handleBMCReset` bails. maybe remove it/put in that function?

internal/controller/bmcsettings_controller.go
```diff
@@ -471,34 +471,32 @@ func (r *BMCSettingsReconciler) updateSettingsAndVerify(ctx context.Context, set
...
+				return ctrl.Result{}, err
 			}
+			return ctrl.Result{}, nil
 		}
-	} else {
-		log.V(1).Info("Waiting for BMC reset post applying BMC settings")
-		if ok, err := r.handleBMCReset(ctx, settings, bmcObj, ConditionBMCResetPostSettingApply); !ok || err != nil {
-			return ctrl.Result{}, err
-		}
+
```
xkonni, line 489: this happens silently, maybe add some log that explains what happens. e.g.

```
log.V(1).Info("Pending BMC attributes exist, waiting for them to clear before applying settings")
```

xkonni (approved):
like it. minor clarifications possible.

internal/controller/bmcsettings_controller.go
atd9876: will update

internal/controller/bmcsettings_controller.go
atd9876: will update

nagadeesh-nagaraja (approved):

PR #876, opened 2026-05-11
Introduce ErrNotSupported sentinel error for vendor BMC methods

### LINKED ISSUE #735: Introduce ErrNotSupported sentinel error for unimplemented vendor BMC methods
## Summary

Controllers such as `internal/controller/bmcversion_controller.go` and `internal/controller/biosversion_controller.go` call vendor-specific BMC client methods (e.g. `CheckBMCPendingComponentUpgrade`) and currently treat **all** errors identically — including the base `RedfishBaseBMC` fallback that returns a "not supported by this manufacturer" error. This means transient Redfish/FirmwareInventory failures are silently swallowed and the upgrade proceeds, which is the same behavior as a legitimate "not supported" response.

## Motivation

Introducing a dedicated `ErrNotSupported` sentinel error would allow controllers to:
- **Ignore** the "not supported by vendor" case (current behavior, intentional)
- **Requeue** on transient or real check failures instead of blindly proceeding with the upgrade

This improves safety: a temporary FirmwareInventory unavailability would no longer silently bypass the pending-upgrade guard.

## Scope

- Define an `ErrNotSupported` sentinel (e.g. in `bmc/bmc.go` or a shared errors file)
- Update `RedfishBaseBMC.CheckBMCPendingComponentUpgrade` (and similar fallback methods) in `bmc/redfish.go` to wrap/return `ErrNotSupported`
- Update affected controllers (`bmcversion_controller.go`, `biosversion_controller.go`, and any future controllers) to use `errors.Is(err, bmc.ErrNotSupported)` to distinguish the two cases
- Consider applying the same pattern to other vendor-specific BMC interface methods that currently have a "not supported" fallback

## References

- PR #733 (where this was first discussed): https://github.com/ironcore-dev/metal-operator/pull/733
- Specific comment: https://github.com/ironcore-dev/metal-operator/pull/733#discussion_r2910355313

/cc @davidgrun

stefanhipfel:
Fixes #735

### DIFF
--- a/bmc/bmc.go
+++ b/bmc/bmc.go
@@ -6,10 +6,14 @@ package bmc
 import (
 	"context"
 	"encoding/json"
+	"errors"
 
 	"github.com/stmcginnis/gofish/schemas"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
+
+// ErrNotSupported is returned when a BMC operation is not supported by the vendor.
+var ErrNotSupported = errors.New("operation not supported by this vendor")
 
 type Manufacturer string
 
--- a/bmc/redfish.go
+++ b/bmc/redfish.go
@@ -905,14 +905,14 @@ func (r *RedfishBaseBMC) UpgradeBMCVersion(_ context.Context, _ string, _ *schem
 }
 
 // CheckBMCPendingComponentUpgrade is a fallback for unknown vendors.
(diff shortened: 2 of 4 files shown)


afritzler:
@coderabbitai review

afritzler (approved):

PR #888, opened 2026-05-13
Add disk cleaning to `metalprobe`

### LINKED ISSUE #899: Add disk cleaning feature in `metalprobe`
## Summary

This issue tracks the disk cleaning feature implementation inside the `metal-probe` image.

## Background

Disk cleaning during server discovery was discussed as a simple and quick solution to ensure disks are wiped before a server becomes available. PR `#888` introduced an initial implementation of disk cleaning within the metal-probe, and the discussion (see [`#613` (comment)](https://github.com/ironcore-dev/metal-operator/issues/613#issuecomment-4442412514)) suggests focusing the metal-probe changes here while the operator-side logic that triggers cleaning can be addressed in a follow-up.

## Proposed Scope

- Implement disk cleaning inside the `metal-probe` image
  - Smart erasure: use `blkdiscard` for SSDs and `shred`/`dd` for HDDs
  - Concurrency and timeouts: up to 4 disks cleaned in parallel; per-disk timeouts (10 min for quick, 24 hours for secure mode)
  - Persistent state: marker file to prevent re-cleaning after restart
  - Security: path validation and command-injection prevention
  - Cleaning occurs before the agent sends discovery, so the Server will not become available while disk cleaning is still in progress

## Out of Scope (to be handled in a separate follow-up)

- Operator logic / controller that triggers the cleaning (e.g. a `ServerCleaning` controller)

## References

- PR `#888`
- Issue `#704`
- Discussion: https://github.com/ironcore-dev/metal-operator/issues/613#issuecomment-4442412514

stefanhipfel:
## Summary

Adds disk cleaning inside the metal probe image

## Features

- **Smart erasure**: blkdiscard for SSDs, shred/dd for HDDs
- **Concurrent**: Up to 4 disks in parallel with timeouts (10min/24h)
- **Persistent state**: Marker file prevents re-cleaning on restart
- **Security**: Path validation, command injection prevention
- Happens before agent sends discovery, so Server will not become available while disk cleaning is still happening.

## Testing

- Tested on Linux VM with "real disks" (via Lima on Mac)

Fixes #899

## Summary by CodeRabbit

* **New Features**
  * Added disk cleaning functionality with `--disk-cleaning` CLI flag to automatically clean disks after server registration.
  * Introduced two cleaning modes: "quick" (basic wipe) and "secure" (comprehensive secure erase).
  * Added tracking for disk cleaning completion status.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ironcore-dev/metal-operator/pull/888?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


afritzler:
Thanks @stefanhipfel for the PR. I commented in the referenced issue how we could simplify the cleanup phase: https://github.com/ironcore-dev/metal-operator/issues/613#issuecomment-4442412514 I would suggest to take the discussion there.

api/v1alpha1/server_types.go
```diff
@@ -81,6 +81,49 @@ type BootOrder struct {
 	Device string `json:"device"`
 }
 
+// DiskCleaningMode defines the disk cleaning method.
+// +kubebuilder:validation:Enum=quick;secure
+type DiskCleaningMode string
```
afritzler: I would hide this behind a global flag for now. E.g. `--enable-disk-sanitatation` or `--disk-sanitazation-policy` (None by default, secure or fast).

internal/controller/server_controller.go
```diff
@@ -418,6 +420,10 @@ func (r *ServerReconciler) handleDiscoveryState(ctx context.Context, bmcClient b
 		return false, err
 	}
 
+	if err := r.handleDiskCleaningCompletion(ctx, server); err != nil {
```
afritzler: This task I would put into an own controller creating claims with tolerations for those cleanup tasks.

afritzler (changes requested):

stefanhipfel:
> Thanks @stefanhipfel for the PR. I commented in the referenced issue how we could simplify the cleanup phase: [#613 (comment)](https://github.com/ironcore-dev/metal-operator/issues/613#issuecomment-4442412514) I would suggest to take the discussion there.

@afritzler 
Would it then make sense to only implement the disk cleaning in the metal-probe in this PR? 
Then add a separate PR to implement the operator logic which triggers the cleaning.
fyi: I remember that we discussed to do the disk cleaning during discovery, to make it very simple for a quick solution.

cmd/metalprobe/main.go
```diff
@@ -53,11 +58,18 @@ func main() {
 		os.Exit(1)
 	}
 
+	if diskCleaningMode != "quick" && diskCleaningMode != "secure" {
```
nagadeesh-nagaraja: Should this strings "quick" and "secure" be converted to enums or constants?

internal/probe/diskcleaning_linux.go
```diff
@@ -0,0 +1,426 @@
...
+	log := logr.FromContextOrDiscard(ctx)
+
+	// Check if disk cleaning was already completed
+	if wasDiskCleaningCompleted() {
+		log.Info("Disk cleaning already completed, skipping")
+		return nil
+	}
+
+	// Validate mode upfront before launching goroutines
+	if mode != "quick" && mode != "secure" {
```
nagadeesh-nagaraja: can be converted to enum or constants.

cmd/metalprobe/main.go
```diff
@@ -34,6 +36,9 @@ func main() {
 		"Duration of time to wait between lldpctl runs.")
 	flag.DurationVar(&LLDPSyncDuration, "lldp-sync-duration", 30*time.Second,
 		"Timeout for the lldpctl run.")
+	flag.BoolVar(&cleanDisks, "clean-disks", false, "Enable disk cleaning during discovery.")
+	flag.StringVar(&diskCleaningMode, "disk-cleaning-mode", "quick",
+		"Disk cleaning mode: 'quick' or 'secure'.")
```
nagadeesh-nagaraja: I would suggest merging these two flags? there is no need for 2. we can do with one...

internal/controller/conditions.go
```diff
@@ -31,6 +31,8 @@ const (
 	ConditionReady = "Ready"
 	// ConditionRetryOfFailedResourceIssued indicates a retry of a failed resource has been issued.
 	ConditionRetryOfFailedResourceIssued = "RetryOfFailedResourceIssued"
+	// ConditionDiskCleaningCompleted indicates that disk cleaning was completed
+	ConditionDiskCleaningCompleted = "DiskCleaningCompleted"
```
afritzler: Can we remove this from this PR and introduce this condition in a subsequent PR.

internal/controller/server_controller.go
```diff
@@ -709,6 +714,7 @@ func (r *ServerReconciler) applyDefaultIgnitionForServer(ctx context.Context, se
 	log.V(1).Info("Applied SSH keypair secret", "SSHKeyPair", sshKeyPairNamespacedName)
 
 	probeFlags := fmt.Sprintf("--registry-url=%s --server-uuid=%s", registryURL, server.Spec.SystemUUID)
+
```
afritzler: I would remove this newline in this PR.

internal/probe/probe.go
```diff
@@ -136,6 +164,11 @@ func (a *Agent) Start(ctx context.Context) error {
 	}
 	a.log.Info("Server registered", "uuid", a.SystemUUID)
 
+	// Perform disk cleaning after registration if enabled
+	if err := a.PerformDiskCleaning(ctx); err != nil {
+		a.log.Error(err, "Disk cleaning failed, continuing with periodic sync")
+	}
+
```
afritzler: We are performing the disk cleaning after we post the `Server` info back to the registry. E.g. if the disk wipe takes 2 hours the server already appears in the API. I would reverse this order.

afritzler: Also we should not swallow an error here. Instead we should exit with -1 and let systemd restart this container and re-run the cleaning. We could thing about storing the already wiped disk to `/tmp` as a hint for a container restart.

internal/probe/diskcleaning_linux.go
```diff
@@ -0,0 +1,427 @@
...
+	path := fmt.Sprintf("/sys/block/%s/queue/rotational", baseName)
+
+	data, err := os.ReadFile(path)
+	if err != nil {
+		// If we can't tell, assume true (rotational) to be safe and use shred
+		return true
+	}
+
+	return strings.TrimSpace(string(data)) == "1"
+}
```
afritzler, line 279: Multipath/RAID device support will be missed here: 

- `/sys/block/mpatha/queue/rotational` doesn't exist, `isRotational` silently returns true, sending an SSD multipath device through the slow shred/dd path (or missing the discard).
- `/sys/class/block/mpatha/ro` may not exist, `isReadOnly` returns an error and the disk is skipped entirely.

We should either resolve the underlying device for multipath/dm cases or document that multipath isn't supported and tighten the regex back.

internal/probe/probe.go
```diff
@@ -114,6 +116,32 @@ func (a *Agent) Init(ctx context.Context) error {
...
+	a.log.Info("Starting disk cleaning", "mode", a.DiskCleaningMode)
+
+	// Add logger to context for cleanDisks
+	ctx = logr.NewContext(ctx, a.log)
+	if err := cleanDisks(ctx, a.DiskCleaningMode); err != nil {
+		return err
+	}
+
+	a.log.Info("Disk cleaning completed successfully")
+	a.diskCleaningComplete = true
```
afritzler: We are writing the file and setting the internal `diskCleaningComplete` var to `true`. This sounds like a split brain setup here. Can't we just rely on one method to keep track of what has been wiped? I would suggest to write in tmpfs the devices which are already covered and as outlined in my other comment handle failures as blocking and let systemd restart the process.

internal/probe/diskcleaning_linux.go
```diff
@@ -0,0 +1,427 @@
...
+	}
+
+	if !isRotational(diskName) {
+		log.V(1).Info("Detected non-rotational flash storage. Using blkdiscard.", "disk", diskName)
+		if err := executeBlkDiscard(ctx, devicePath, true); err == nil {
+			return rereadPartitionTable(ctx, devicePath)
+		} else {
+			log.Error(err, "blkdiscard failed, falling back to shred/dd", "disk", diskName)
+		}
+	}
```
afritzler: We silently swallow the fact that `blkdiscard` could not be used for wiping. This kind of waters down the "secure" aspect of the wipe process. Can we add this to the flag description that this will happen only if bkldiscard is supported by the firmware/device.

internal/probe/diskcleaning_linux.go
```diff
@@ -0,0 +1,427 @@
...
+		"of="+devicePath,
+		"bs=1M")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		outputStr := string(output)
+		if !strings.Contains(outputStr, "No space left on device") {
+			return fmt.Errorf("dd failed: %w, output: %s", err, outputStr)
+		}
+		log.V(1).Info("dd completed (expected 'No space left' at end)", "disk", diskName)
+	}
```
afritzler, line 239: `dd if=/dev/urandom of=/dev/sdX bs=1M` can run for hours under the 24h timeout with zero log output. We might want to add `status=progress` (and `oflag=direct` to avoid page-cache pollution), or wrap with periodic logging. A human operator who might be debugging it and is staring at a stuck-looking probe will appreciate it :).

internal/probe/diskcleaning_linux.go
```diff
@@ -0,0 +1,427 @@
...
+		return fmt.Errorf("failed to get disk size: %w", err)
+	}
+
+	if sizeBytes > 10*1024*1024 {
+		log.V(1).Info("Wiping disk footer", "disk", diskName)
+		offset := sizeBytes - (10 * 1024 * 1024)
+		if err := wipeDiskRangeNative(ctx, devicePath, offset, 10*1024*1024); err != nil {
+			return fmt.Errorf("failed to wipe disk footer: %w", err)
+		}
+	}
```
afritzler: We already enumerated the disks via `ghw` and have `disk.SizeBytes` in scope at the call site. Using it avoids a fork/exec and removes one place where command output parsing can go wrong.

internal/controller/server_controller_test.go
```diff
@@ -247,7 +248,7 @@ var _ = Describe("Server Controller", func() {
 		))
 
 		By("Starting the probe agent")
-		probeAgent := probe.NewAgent(GinkgoLogr, server.Spec.SystemUUID, registryURL, 100*time.Millisecond, 1*time.Second, 50*time.Millisecond, 250*time.Millisecond)
+		probeAgent := probe.NewAgent(GinkgoLogr, server.Spec.SystemUUID, registryURL, 100*time.Millisecond, 1*time.Second, 50*time.Millisecond, 250*time.Millisecond, "")
```
afritzler: Since the `NewAgent` signature is growing here now by one additional attribute I would suggest to introduce an `AgentConfig` struct and pass that in. That will also eliminate the empty string passing here and in other places.

afritzler (changes requested):
A few other things we might address in a separate PR:
 
1. Concurrency machinery is overbuilt. `sync.WaitGroup` + `chan struct{} semaphore` + `sync.Mutex` + custom `DiskCleaningResult` struct + manual context-cancel-during-acquire dance. We only ever use `Success` to count.  `golang.org/x/sync/errgroup` with `SetLimit(4)` collapses the whole orchestration to ~10 lines:

```golang
g, ctx := errgroup.WithContext(ctx)
g.SetLimit(4)
for _, d := range disks {
    d := d
    g.Go(func() error { return cleanOne(ctx, d, mode) })
}
return g.Wait()
```

2. `wipeDiskRangeNative` is 30 lines avoiding a problem we don't have. The comment says it exists "to avoid integer truncation bugs with dd seek". Linux `dd` has handled 64-bit offsets for ~20 years. And if we want native Go,   `os.WriteAt` does `seek+write` atomically; no need to hand-roll a chunked write loop with a context check (it's writing 10 MiB, it'll finish in milliseconds).

3. Quick mode is reinventing `wipefs`. Wiping the first and last 10 MiB to "destroy partition tables and filesystem signatures" is exactly what `wipefs -a /dev/sdX` does. It knows about GPT backup headers, LVM, mdraid, LUKS, etc. Replace `quickCleanDisk`+ `wipeDiskRangeNative` + `getDiskSize` with one `exec.CommandContext(ctx, "wipefs", "-a", devicePath)`.

4. Triple mode validation. Validated in `main.go` after `flag.Parse`, again at the top of `cleanDisks`, again in the goroutine's switch default. Pick the boundary and trust internal callers.

5. Three-tier secure fallback (blkdiscard → shred → dd) for marginal gain. `shred` between `blkdiscard` and `dd` doesn't buy much. If `blkdiscard` failed because it's an HDD, just go to dd. Drops a fragile "No space left on device" string-match on the way out too.

internal/controller/server_controller_test.go
stefanhipfel: Done

afritzler:
@stefanhipfel can we close this one in favor of https://github.com/ironcore-dev/metal-maintenance-operator/pull/71

stefanhipfel:
closed in favour of https://github.com/ironcore-dev/metal-maintenance-operator/pull/71

PR #894, opened 2026-05-15
Add additional labels from BMC and Server resources to metrics

### LINKED ISSUE #893: Add configurable label enrichment to Redfish telemetry metrics

Redfish telemetry / metrics (`internal/serverevents`, `cmd/main.go`)

When managing a large number of servers, it is difficult to filter Redfish telemetry dashboards and alerts by topology or location. The `redfish_monitor_reading` and `redfish_event_alert_total` metrics only carry a `hostname` label (the BMC resource name), which provides no organisational context such as region, availability zone, rack, or physical location. Operators running multi-site or multi-zone deployments must either maintain external lookup tables or accept coarse-grained dashboards that cannot be scoped to a subset of the fleet.

**Describe the feature**

Introduce two new CLI flags that allow operators to declare additional Prometheus label dimensions for both Redfish metrics, sourced from labels on the `BMC` or `Server` Kubernetes resources:

- `--redfish-metric-labels-from-bmc` — pulls label values from the `BMC` resource (matched by resource name == `hostname`)
- `--redfish-metric-labels-from-server` — pulls label values from the `Server` resource linked via `spec.bmcRef.name`

The value format for each flag is a comma-separated list of `kubernetes-label-key=prometheus-label-name` pairs. Label lookups are cached per hostname with a 1-hour TTL. Missing label keys always emit as empty strings so that missing data never blocks metric emission.

This makes the enrichment fully dynamic and organisation-agnostic — no label keys are hardcoded in the operator.

**Proposed API or behavior changes**

New CLI flags on the manager binary:

```
--redfish-metric-labels-from-bmc=topology.kubernetes.io/region=region,topology.kubernetes.io/zone=zone
--redfish-metric-labels-from-server=metadata.metal.ironcore.dev/location=location,metadata.metal.ironcore.dev/rack=rack
```

Resulting enriched metric:

```text
redfish_monitor_reading{
  hostname="node001-bmc",
  metric_id="CPU1Temp",
  type="Temperature",
  unit="Cel",
  origin_context="/Chassis/1/Thermal",
  region="eu-de-1",
  zone="eu-de-1a",
  location="building-b",
  rack="row3-rack7"
} 42.5
```

Helm chart configuration in `values.yaml`:

```yaml
redfishLabels:
  bmc:
    topology.kubernetes.io/region: region
    topology.kubernetes.io/zone: zone
  server:
    metadata.metal.ironcore.dev/location: location
    metadata.metal.ironcore.dev/rack: rack
```

**Alternatives considered**

- Hardcoding specific label keys (e.g. `region`, `zone`) — rejected as it ties the operator to a specific organisation's labelling conventions and requires code changes for each deployment.
- Enriching at the Prometheus/recording-rule level using `group_left` joins — possible but requires operators to maintain join rules and knowledge of the Kubernetes API, increasing operational complexity.

**Additional context**

Both `BMC` and `Server` are cluster-scoped resources. The `Server` → `BMC` link is resolved via the existing field index on `spec.bmcRef.name` registered in `internal/controller/index_fields.go`.

asergeant01:
Fixes #893 

## Summary by CodeRabbit

* **New Features**
  * Added CLI flags to enrich Redfish metrics with labels from Kubernetes BMC and Server resources using configurable key-to-label mappings.
  * Enhanced observability with additional dynamic label dimensions for metrics when event monitoring is enabled.

* **Documentation**
  * Added documentation describing Redfish telemetry metrics, label enrichment configuration, and usage examples.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ironcore-dev/metal-operator/pull/894?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -732,7 +732,9 @@ func main() { // nolint: gocyclo
 		}
 		if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
 			setupLog.Info("starting event server for alerts and metrics", "EventURL", eventURL)
-			eventServer := serverevents.NewServer(setupLog, fmt.Sprintf(":%d", eventPort), mgr.GetClient(), bmcLabelMappings, serverLabelMappings)
+			eventServer := serverevents.NewServer(
+				setupLog, fmt.Sprintf(":%d", eventPort), mgr.GetClient(), bmcLabelMappings, serverLabelMappings,
+			)
 			if err := eventServer.Start(ctx); err != nil {
 				return fmt.Errorf("unable to start event server: %w", err)
 			}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -83,6 +83,8 @@ func main() { // nolint: gocyclo
 		eventPort                          int
 		eventURL                           string
 		eventProtocol                      string
+		redfishMetricLabelsFromBMC         string
+		redfishMetricLabelsFromServer      string
 		registryClientTimeout              time.Duration
 		registryDataMaxAge                 time.Duration
 		registryResyncInterval             time.Duration
@@ -143,6 +145,16 @@ func main() { // nolint: gocyclo
 	flag.IntVar(&eventPort, "event-port", 10001, "The port to use for the server events endpoint for alerts and metrics.")
 	flag.StringVar(&eventProtocol, "event-protocol", "http",
 		"The protocol to use for the server events endpoint for alerts and metrics.")
+	flag.StringVar(&redfishMetricLabelsFromBMC, "redfish-metric-labels-from-bmc", "",
+		"Comma-separated list of 'kubernetes-label-key=prometheus-label-name' pairs. "+
+			"Each pair adds an additional label dimension to Redfish telemetry metrics, "+
+			"sourced from the matching label on the BMC resource. "+
+			"Example: topology.kubernetes.io/region=region,topology.kubernetes.io/zone=zone")
+	flag.StringVar(&redfishMetricLabelsFromServer, "redfish-metric-labels-from-server", "",
+		"Comma-separated list of 'kubernetes-label-key=prometheus-label-name' pairs. "+
+			"Each pair adds an additional label dimension to Redfish telemetry metrics, "+
+			"sourced from the matching label on the Server resource linked via spec.bmcRef.name. "+
+			"Example: metadata.metal.ironcore.dev/location=location,metadata.metal.ironcore.dev/rack=rack")
 	flag.StringVar(&probeImage, "probe-image", "", "Image for the first boot probing of a Server.")
 	flag.StringVar(&probeOSImage, "probe-os-image", "", "OS image for the first boot probing of a Server.")
 	flag.StringVar(&managerNamespace, "manager-namespace", "default", "Namespace the manager is running in.")
@@ -708,9 +720,19 @@ func main() { // nolint: gocyclo
 	}
 
 	if eventURL != "" {
+		bmcLabelMappings, err := serverevents.ParseLabelMappings(redfishMetricLabelsFromBMC)
+		if err != nil {
+			setupLog.Error(err, "Invalid --redfish-metric-labels-from-bmc")
+			os.Exit(1)
+		}
+		serverLabelMappings, err := serverevents.ParseLabelMappings(redfishMetricLabelsFromServer)
+		if err != nil {
+			setupLog.Error(err, "Invalid --redfish-metric-labels-from-server")
+			os.Exit(1)
+		}
 		if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
 			setupLog.Info("starting event server for alerts and metrics", "EventURL", eventURL)
-			eventServer := serverevents.NewServer(setupLog, fmt.Sprintf(":%d", eventPort))
+			eventServer := serverevents.NewServer(setupLog, fmt.Sprintf(":%d", eventPort), mgr.GetClient(), bmcLabelMappings, serverLabelMappings)
 			if err := eventServer.Start(ctx); err != nil {
 				return fmt.Errorf("unable to start event server: %w", err)
 			}
--- a/dist/chart/templates/_helpers.tpl
+++ b/dist/chart/templates/_helpers.tpl
@@ -48,3 +48,17 @@ app.kubernetes.io/instance: {{ .Release.Name }}
     $hasValidating = true }}{{- end }}
 {{- end }}
 {{ $hasValidating }}}}{{- end }}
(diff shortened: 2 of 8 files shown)


afritzler:
Thanks for the PR @asergeant01 I think it makes sense to expose topology information,

However, I'm wondering whether we should solve this at the Prometheus layer rather than inside the operator. The standard pattern for this in the Prometheus ecosystem is an info-style metric + `group_left` join:

1. Expose topology labels via a `kube-state-metrics` customresourcestate config (no operator code needed):

```yaml
# kube-state-metrics custom resource state config
spec:
  resources:
    - groupVersionKind:
        group: metal.ironcore.dev
        version: v1alpha1
        kind: BMC
      labelsFromPath:
        bmc_name: [metadata, name]
      metrics:
        - name: metal_bmc_labels
          help: "BMC resource labels for join enrichment"
          each:
            type: Info
            info:
              labelsFromPath:
                region: [metadata, labels, "topology.kubernetes.io/region"]
                zone: [metadata, labels, "topology.kubernetes.io/zone"]
```

2. Join at query time (or materialize as a recording rule):

```
redfish_monitor_reading
  * on(hostname) group_left(region, zone)
  label_replace(metal_bmc_labels, "hostname", "$1", "bmc_name", "(.*)")
```

This avoids:
- Kubernetes API calls on the metrics ingestion hot path
- A TTL cache that can serve stale labels for up to an hour (vs. KSM which watches resources)
- New operator flags that require a restart to change label dimensions
- Cardinality risk from misconfiguration (with recording rules you see the cardinality before it hits long-term storage)

Adding label dimensions becomes a config change in KSM + a recording rule, no operator redeployment needed.

asergeant01:
Thanks for the feedback @afritzler. The KSM + group_left pattern is the right call in many contexts, but I think the in-operator approach is still the better fit here. However, you do raise some valid technical concerns which I think I can address.

The recording rule requirement is a hard blocker for many users - The KSM approach only works ergonomically if a recording rule materialises the join — raw group_left in every dashboard query and alert rule is not realistic. That recording rule must live somewhere that evaluates against the metrics store. For users who remote-write to a managed observability platform (Grafana Cloud, Thanos, VictoriaMetrics, etc.) they often do not control the ruler component, or the recording rule must be deployed separately to the remote ruler rather than alongside the operator. 

The label_replace rename is fragile - the proposed join requires `label_replace(metal_bmc_labels, "hostname", "$1", "bmc_name", "(.*)")` to align the join key. A mismatch between the KSM-emitted bmc_name and the hostname label on Redfish metrics silently returns no data. That's not a great experience for users.

What I think I can fix is your concerns around K8s API calls on the hot path and 1-hour stale labels. They're artifacts of the TTL cache implementation, not of the in-operator approach itself. I think the way KSM does this is with runtime manager's informer-backed client.Client. In our implementation that client is already passed in, already watch-based, and reads are local in-memory lookups — zero API round trips, and labels reflect changes near-instantly. 

The remaining concerns (restart to add a label dimension, separation of concerns) are right but I think are acceptable trade-offs given that topology labels like region and zone are stable once configured.

xkonni (approved):
Nice, the informer cache fix is the right call. Field index is there, parsing is solid. LGTM 👍

asergeant01:
@coderabbitai review

afritzler:
> Thanks for the feedback @afritzler. The KSM + group_left pattern is the right call in many contexts, but I think the in-operator approach is still the better fit here. However, you do raise some valid technical concerns which I think I can address.
> 
> The recording rule requirement is a hard blocker for many users - The KSM approach only works ergonomically if a recording rule materialises the join — raw group_left in every dashboard query and alert rule is not realistic. That recording rule must live somewhere that evaluates against the metrics store. For users who remote-write to a managed observability platform (Grafana Cloud, Thanos, VictoriaMetrics, etc.) they often do not control the ruler component, or the recording rule must be deployed separately to the remote ruler rather than alongside the operator.
> 
> The label_replace rename is fragile - the proposed join requires `label_replace(metal_bmc_labels, "hostname", "$1", "bmc_name", "(.*)")` to align the join key. A mismatch between the KSM-emitted bmc_name and the hostname label on Redfish metrics silently returns no data. That's not a great experience for users.
> 
> What I think I can fix is your concerns around K8s API calls on the hot path and 1-hour stale labels. They're artifacts of the TTL cache implementation, not of the in-operator approach itself. I think the way KSM does this is with runtime manager's informer-backed client.Client. In our implementation that client is already passed in, already watch-based, and reads are local in-memory lookups — zero API round trips, and labels reflect changes near-instantly.
> 
> The remaining concerns (restart to add a label dimension, separation of concerns) are right but I think are acceptable trade-offs given that topology labels like region and zone are stable once configured.

The `label_replace` example doesn't apply, since both metrics share `hostname`. The join is just `on(hostname) group_left(...)` with no renaming needed. Recording rules are also broadly supported on managed backends (Mimir, Thanos, Grafana Cloud, VictoriaMetrics all support them), and inline `group_left` in dashboard panels works fine in practice.

The concern I'd want addressed is cardinality. Sensor readings already fan out across `metric_id × type × unit × origin_context`, and every added topology label multiplies that. An info metric pays the cost once per BMC. The "restart to add a label dimension" tradeoff, plus the operator owning user specific topology vocabulary, are exactly what the KSM pattern exists to avoid.

asergeant01:
@afritzler Ok, I understand. You're right on the `label_replace` point — I misread the query. Since both metrics already carry `hostname` and KSM would emit with the same key, the join is just `on(hostname) group_left(...).`

On your cardinality point, I get the concern-  but the intended use case (topology labels like region/zone/rack that are 1:1 with hostname), adding them doesn't actually multiply cardinality — (hostname, metric_id, ..., region, zone) has the same number of unique series as without them, since region and zone are deterministic given hostname, e.g. `redfish_bmc_info{hostname="...", region="...", zone="..."}`. But your point about misconfiguration is valid — if someone maps a high-cardinality label by mistake there's no guard, and with an info metric you'd see the blast radius before it hits long-term storage.

A middle path that I think addresses your data model concern without the KSM dependency: instead of baking the labels into the operational metrics, the operator emits a redfish_bmc_info gauge (constant 1, labels from BMC resource) using the same informer cache already in place. Users join at query time with group_left. No schema coupling on the operational metrics, no KSM required, and cardinality is paid once per BMC.

Would that be an acceptable direction? Happy to rework the PR along those lines.

afritzler:
What I am asking myself here is: Which deployment scenario specifically motivates this change? If KSM is available + the fact that we here are 100% in the k8s types world -> we can solve that problem with 0 lines of code.

asergeant01:
I take your point, this simply eased use of consumption.

PR #929, opened 2026-06-04
Replace docker with containerd in default ignition probe template

### LINKED ISSUE #906: Fix server discovery in air-gapped environments
Server discovery currently does not work in air-gapped environments. 

`metalprobe` is running in a docker container, the docker runtime is not preinstalled in our current GardenLinux image and installed after startup— which is not possible without external network access.

There is no GardenLinux flavor that comes with docker, but the `gardener` or `chost` feature ship containerd/ctr pre-installed. This removes the need to install any packages and allows discovery to run in air-gapped enviromnents.
At least as long as the metalprobe image is available on a reachable OCI mirror, e.g. the same as the GardenLinux OS image that is booted on the node.

**Tested flow (on `metal-gardener_prod` flavor):**
```sh
systemctl enable --now containerd
ctr images pull <mirror>/ironcore-dev/metalprobe:latest
ctr run --rm --net-host --privileged <mirror>/ironcore-dev/metalprobe:latest metalprobe ./launch.sh \
  --server-uuid=<uuid> \
  --registry-url=<url>
```

The `launch.sh` entrypoint starts `lldpd` before invoking metalprobe, which is required since `lldpctl` communicates with the daemon on the host via Unix socket (due to `--net-host`).

**Open items:**
- The tested flavor (`metal-gardener_prod`, ~2.5 GB) is large. A more minimal flavor based on `metal` + `chost` would be preferable — but no such image is currently published on ghcr.io/gardenlinux/gardenlinux. This would need to be built or requested from the GardenLinux team.

xkonni:
gardenlinux metal-capi ships containerd via the chost feature set, so docker does not need to be installed at probe boot time. Switch to ctr for image pull and container run, and add the containerd mount units that wire a ramdisk into /var/lib/containerd before containerd starts.

this comes with the added benefit that it does not require any internet connectivity to install packages.

Fixes #906 

## Summary by CodeRabbit

* **Chores**
  * Migrated container runtime from Docker to containerd, enhancing system performance and reducing resource consumption.

### DIFF
--- a/config/manager/ignition-template.yaml
+++ b/config/manager/ignition-template.yaml
@@ -2,36 +2,45 @@ variant: fcos
 version: "1.3.0"
 systemd:
   units:
-    - name: docker-install.service
+    - name: var-lib-containerd.mount
       enabled: true
       contents: |-
         [Unit]
-        Description=Install Docker
-        Before=metalprobe.service
-        [Service]
-        Restart=on-failure
-        RestartSec=20
-        Type=oneshot
-        RemainAfterExit=yes
-        ExecStart=/usr/bin/apt-get update
-        ExecStart=/usr/bin/apt-get install docker.io docker-cli -y
+        Description=tmpfs for containerd image cache
+        DefaultDependencies=no
+        Before=containerd.service
+
+        [Mount]
+        What=tmpfs
+        Where=/var/lib/containerd
+        Type=tmpfs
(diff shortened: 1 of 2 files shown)


asergeant01 (approved):
Looks good to me.

stefanhipfel (approved):
lgtm

PR #951, opened 2026-06-15
Transition ServerMaintenance to Pending for missing Server resource

### LINKED ISSUE #367: When deleting `BMC`s the `ServerMaintenance` is not being deleted
**Describe the bug**
When I delete BMC's to rediscover the associated server maintenance is orphaned and kept around

**To Reproduce**
Create a BMC and a server maintenance for it. Delete the bmc.

**Expected behavior**
The server maintenance should have an owner reference on the bmc and be deleted along with the bmc

stefanhipfel:
Fixes #367

When a BMC is deleted, Kubernetes cascades deletion to owned `Server` resources. `ServerMaintenance` objects referencing those servers were left orphaned due to two bugs in the `ServerMaintenanceReconciler`:

- `reconcile()` returned silently on `NotFound` server before adding the finalizer — so the resource was never garbage collected
- `delete()` propagated `NotFound` as an error, blocking finalizer removal permanently

Both paths now handle a missing `Server` gracefully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reconciliation when a referenced Server is missing: ServerMaintenance now records the condition, transitions to **Pending**, and exits without further processing.

* **Tests**
  * Added coverage confirming that ServerMaintenance remains available and transitions to **Pending** when its referenced Server is deleted.

### DIFF
--- a/internal/controller/servermaintenance_controller.go
+++ b/internal/controller/servermaintenance_controller.go
@@ -73,6 +73,13 @@ func (r *ServerMaintenanceReconciler) reconcile(ctx context.Context, maintenance
 	server := &metalv1alpha1.Server{}
 	if err := r.Get(ctx, client.ObjectKey{Name: maintenance.Spec.ServerRef.Name}, server); err != nil {
 		if apierrors.IsNotFound(err) {
+			log.V(1).Info("Referenced Server not found, transitioning to Pending state",
+				"ServerMaintenance", client.ObjectKeyFromObject(maintenance),
+				"Server", maintenance.Spec.ServerRef.Name,
+			)
+			if modified, err := r.patchMaintenanceState(ctx, maintenance, metalv1alpha1.ServerMaintenanceStatePending); err != nil || modified {
+				return ctrl.Result{}, err
+			}
 			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{}, fmt.Errorf("failed to get Server: %w", err)
--- a/internal/controller/servermaintenance_controller_test.go
+++ b/internal/controller/servermaintenance_controller_test.go
@@ -537,6 +537,32 @@ var _ = Describe("ServerMaintenance Controller", func() {
 		Expect(k8sClient.Delete(ctx, serverClaim)).To(Succeed())
 	})
 
+	It("should transition ServerMaintenance to Pending state when its referenced Server no longer exists", func(ctx SpecContext) {
+		By("Creating a ServerMaintenance object")
+		serverMaintenance := &metalv1alpha1.ServerMaintenance{
+			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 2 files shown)


internal/controller/servermaintenance_controller.go
```diff
@@ -73,6 +73,10 @@ func (r *ServerMaintenanceReconciler) reconcile(ctx context.Context, maintenance
 	server := &metalv1alpha1.Server{}
 	if err := r.Get(ctx, client.ObjectKey{Name: maintenance.Spec.ServerRef.Name}, server); err != nil {
 		if apierrors.IsNotFound(err) {
+			log.V(1).Info("Referenced Server not found, deleting ServerMaintenance")
+			if err := r.Delete(ctx, maintenance); err != nil {
+				return ctrl.Result{}, client.IgnoreNotFound(err)
+			}
```
nagadeesh-nagaraja: dont you think this might cause any flaky issue to delete the maintenances? 
I think the owner of the servermaintenance should be responsible for deleting the servermaintenance.

nagadeesh-nagaraja (commented):
seems like we both kind of working on similar topic:

https://github.com/ironcore-dev/metal-operator/pull/950

I believe that the deletion should not be forced if in case its caused by flaky issue. 

the deletion should be handled by the creator of the maintenance?  let me know what you think.

internal/controller/servermaintenance_controller.go
stefanhipfel: why should we leave orphaned maintenances? Same way we don't leave orphaned servers around after their bmc got deleted.
No one will clean them up!

nagadeesh-nagaraja: maintenances are mostly created by someone..
example when BMCSettings or other similar operators creates it, it needs to be deleted by them as they are the owner of it. they should detect the ref server or BMC is gone and delete themself and hence delete the maintenance as well. (this is already done when the SET CRD detects the server is gone) 

if its created by the user (human), then we can delete them. as its easier for user. may be we can check if it has a owner ref, and if so do not delete it.. let the owner handle it?

afritzler (changes requested):
I would suggest not to delete resources in the reconciliation flow: Following analogy: if you create a `Pod` with `spec.nodeName` and the `Node` is not there, nobody will in k8s delete the `Pod`. It will transition into a `Pending` state. For our `ServerMaintenance` I would suggest that whoever creates the `ServerMaintenance` should be responsible deleting it.

stefanhipfel:
> ServerMaintenance

I don't fully agree, since people just don't delete their servermaintenances or just forget it. Can we at least put a different state to the serverMaintenance object? Otherwise it would keep `InMaintenance` forever even with the server gone, which is also not true!

afritzler (approved):

afritzler (approved):

PR #966, opened 2026-06-24
Split BMC into capability interfaces and an open vendor registry

### LINKED ISSUE #967: Make the `bmc` package extensible
## Problem

The `bmc` package today is closed for extension. Third parties that want to reuse it to add a new vendor, override one for a private OEM, or attach vendor-only methods, have to fork the package.

Two specific obstacles:

- `NewRedfishBMCClient` hard-codes a switch over Dell/HPE/Lenovo/Supermicro.
- The single 40-method `BMC` interface forces any vendor implementation to satisfy the full surface, and offers no place to plug in OEM-only methods.

## What we want

- Register a custom vendor (or override a built-in) without forking the package.
- Override individual methods on the standard Redfish implementation.
- Expose vendor-only methods that consumers can opt into via type assertion.
- Keep the existing API working. No churn in current callers.

afritzler:
# Proposed Changes

## Motivation

The `bmc` package had two extensibility problems:

1. **Closed for new vendors.** `NewRedfishBMCClient` hard-coded a switch over Dell/HPE/Lenovo/Supermicro. Adding a vendor or overriding one for a private OEM meant forking the package.
2. **One fat interface.** Every consumer took the full `BMC` interface even when it only needed a handful of methods. Vendor implementations had to satisfy the entire surface even to override a single method.

This change applies two Go interface segregations and an explicit registry passed through `Options`.

## What changes

Split `BMC` into capability interfaces (`PowerController`, `BootController`, `SystemInspector`, `BIOSManager`, `BMCSettingsManager`, `FirmwareUpdater`, `ManagerController`, `AccountManager`, `EventSubscriber`) and keep `BMC` as their union (plus `Logout()`) for backward compatibility.

Replace the hard-coded vendor switch with `Options.AdditionalVendors`, a `map[Manufacturer]VendorFactory` that is merged on top of `DefaultVendors()` inside `NewRedfishBMCClient`. Callers only need to supply the extra OEMs they want to add; registering the same key as a built-in overrides it. Export `Client()`, `Options()` and `Manufacturer()` on `RedfishBaseBMC` so third-party vendor implementations can embed it without reaching into unexported fields.

Narrow the consumers where it conveys a real role (`serverevents` on `EventSubscriber`, `helper` on `ManagerController`); leave the rest on `bmc.BMC`.

## Extending an existing vendor or adding a new one

```go
type ACMEBMC struct{ *bmc.RedfishBaseBMC }

var _ bmc.BMC = (*ACMEBMC)(nil)

// Override an existing method (delegate + patch, or roll your own against r.Client()).
func (r *ACMEBMC) GetSystemInfo(ctx context.Context, uri string) (bmc.SystemInfo, error) {
    info, err := r.RedfishBaseBMC.GetSystemInfo(ctx, uri)
    if err != nil {
      return info, err
    }
    if info.SKU == "" {
        info.SKU = "ACME-" + info.SerialNumber[:4]
    }
    return info, nil
}

// Add a new OEM-only method which is not part of bmc.BMC.
// Publish an extension interface so callers can discover it via type assertion.
type ThermalProfileGetter interface {
    GetACMEThermalProfile(ctx context.Context) (ThermalProfile, error)
}

func (r *ACMEBMC) GetACMEThermalProfile(ctx context.Context) (ThermalProfile, error) {
    // OEM-specific Redfish call via r.Client()
}
```

Register through `Options`. The built-in Dell/HPE/Lenovo/Supermicro factories stay active automatically:

```go
opts := bmc.Options{
  Endpoint: "...", Username: "...", Password: "...",
  AdditionalVendors: map[bmc.Manufacturer]bmc.VendorFactory{
      "ACME Corp": func(b *bmc.RedfishBaseBMC) bmc.BMC {
          return &ACMEBMC{RedfishBaseBMC: b}
      },
  },
}
client, _ := bmc.NewRedfishBMCClient(ctx, opts)

if tp, ok := client.(ThermalProfileGetter); ok {
    profile, _ := tp.GetACMEThermalProfile(ctx)
}
```

Fixes #967

### DIFF
--- a/bmc/bmc.go
+++ b/bmc/bmc.go
@@ -183,12 +183,6 @@ type EventSubscriber interface {
 	DeleteEventSubscription(ctx context.Context, uri string) error
 }
 
-// Logouter closes the BMC client connection.
-type Logouter interface {
-	// Logout closes the BMC client connection by logging out.
-	Logout()
-}
-
 // BMC is the union of every capability a Redfish BMC client may offer. It
 // exists so that constructors can return a single fully-featured value;
 // individual consumers should depend on the narrower capability interfaces
@@ -203,17 +197,20 @@ type BMC interface {
 	ManagerController
 	AccountManager
 	EventSubscriber
-	Logouter
+
+	// Logout closes the BMC client connection by logging out.
+	Logout()
 }
 
 // VendorFactory wraps a *RedfishBaseBMC in a vendor-specific implementation.
-// External packages can supply their own factories via Options.Vendors to
-// extend the BMC client with new manufacturers without modifying this package.
+// External packages can supply their own factories via Options.AdditionalVendors
+// to extend the BMC client with new manufacturers, or to override a built-in,
+// without modifying this package.
 type VendorFactory func(base *RedfishBaseBMC) BMC
 
 // DefaultVendors returns the built-in vendor factories for Dell, HPE, Lenovo
 // and Supermicro. Callers can copy and extend this map, or build their own
-// from scratch, and pass the result through Options.Vendors.
+// from scratch, and pass the extras through Options.AdditionalVendors.
 func DefaultVendors() map[Manufacturer]VendorFactory {
 	return map[Manufacturer]VendorFactory{
 		ManufacturerDell:       func(b *RedfishBaseBMC) BMC { return &DellRedfishBMC{RedfishBaseBMC: b} },
--- a/bmc/bmc.go
+++ b/bmc/bmc.go
@@ -32,8 +32,15 @@ const (
 	ComponentTypeBIOS ComponentType = "BIOS"
 )
 
-// BMC defines an interface for interacting with a Baseboard Management Controller.
-type BMC interface {
+// The BMC interface is intentionally split into smaller, capability-focused
+// interfaces below. Each consumer should accept the narrowest combination of
+// interfaces it actually uses, following the "accept interfaces, return
+// structs" idiom and io.Reader-style interface segregation. BMC itself is
+// retained as the union of all capabilities for callers (such as the BMC
+// factory) that need to hand out a fully-featured client.
+
+// PowerController manages the power state of a system.
+type PowerController interface {
 	// PowerOn powers on the system.
 	PowerOn(ctx context.Context, systemURI string) error
 
@@ -46,6 +53,12 @@ type BMC interface {
 	// Reset performs a reset on the system.
 	Reset(ctx context.Context, systemURI string, resetType schemas.ResetType) error
 
+	// WaitForServerPowerState waits for the server to reach the specified power state.
+	WaitForServerPowerState(ctx context.Context, systemURI string, powerState schemas.PowerState) error
+}
+
+// BootController manages boot overrides and boot order for a system.
+type BootController interface {
 	// SetBootOverride configures the system to network-boot on its next
 	// power-on, bypassing the persistent boot order. If persistent is false
 	// the override applies to a single boot only; if true it applies to every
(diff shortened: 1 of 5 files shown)


afritzler:
@coderabbitai review

bmc/redfish.go
```diff
@@ -129,18 +138,33 @@ func NewRedfishBMCClient(ctx context.Context, options Options) (BMC, error) {
...
-	default:
-		return base, nil
+	vendors := options.Vendors
+	if vendors == nil {
+		vendors = DefaultVendors()
 	}
+	if factory, ok := vendors[Manufacturer(manufacturer)]; ok {
+		return factory(base), nil
+	}
+	return base, nil
```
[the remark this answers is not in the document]
afritzler: Will address this later when this problem hits us.

afritzler:
@coderabbitai review

bmc/redfish.go
```diff
@@ -55,6 +56,13 @@ type Options struct {
 	ResourcePollingTimeout  time.Duration
 	PowerPollingInterval    time.Duration
 	PowerPollingTimeout     time.Duration
+
+	// Vendors maps a manufacturer string (as reported by Redfish) to a
+	// factory that wraps the base Redfish client in a vendor-specific
+	// implementation. When nil, NewRedfishBMCClient uses DefaultVendors().
+	// To extend the set with a private OEM, copy DefaultVendors() and add
+	// the additional entries.
+	Vendors map[Manufacturer]VendorFactory
```
stefanhipfel: @afritzler thanks for fixing this!!

 maybe consider making this additive rather than replace-the-whole-map?

 Suggestion: rename to AdditionalVendors and merge with DefaultVendors() inside NewRedfishBMCClient:

```
  vendors := DefaultVendors()
  for k, v := range options.AdditionalVendors {
      if v == nil {
          return nil, fmt.Errorf("nil vendor factory for manufacturer %q", k)
      }
      vendors[k] = v
  }

```

bmc/bmc.go
```diff
@@ -149,12 +172,68 @@ type BMC interface {
...
+type EventSubscriber interface {
+	// CreateEventSubscription creates an event subscription for the manager.
+	CreateEventSubscription(ctx context.Context, destination string, eventType schemas.EventFormatType, protocol schemas.DeliveryRetryPolicy) (string, error)
+
+	// DeleteEventSubscription deletes an event subscription for the manager.
+	DeleteEventSubscription(ctx context.Context, uri string) error
+}
+
+// Logouter closes the BMC client connection.
+type Logouter interface {
```
stefanhipfel: could we drop Logouter and keep Logout() directly on BMC?

bmc/redfish.go
```diff
@@ -129,18 +138,36 @@ func NewRedfishBMCClient(ctx context.Context, options Options) (BMC, error) {
...
+}
+
+// Client returns the underlying gofish API client. External vendor
+// implementations that embed *RedfishBaseBMC use this to perform OEM-specific
+// HTTP requests without re-establishing the connection.
+func (r *RedfishBaseBMC) Client() *gofish.APIClient {
+	return r.client
+}
+
+// ClientOptions Options returns the options the client was constructed with.
```
stefanhipfel: small typo adding one Options too many

bmc/redfish.go
```diff
@@ -129,18 +138,36 @@ func NewRedfishBMCClient(ctx context.Context, options Options) (BMC, error) {
...
+func (r *RedfishBaseBMC) ClientOptions() Options {
+	return r.options
+}
+
+// Manufacturer returns the manufacturer detected during connect, or the empty
+// string when detection failed (for example because no Systems were exposed
+// during endpoint discovery).
+func (r *RedfishBaseBMC) Manufacturer() string {
+	return r.manufacturer
```
stefanhipfel: maybe we could use `return Manufacturer(r.manufacturer)` instead here.

bmc/redfish.go
```diff
@@ -129,18 +138,36 @@ func NewRedfishBMCClient(ctx context.Context, options Options) (BMC, error) {
...
+// Client returns the underlying gofish API client. External vendor
+// implementations that embed *RedfishBaseBMC use this to perform OEM-specific
+// HTTP requests without re-establishing the connection.
+func (r *RedfishBaseBMC) Client() *gofish.APIClient {
+	return r.client
+}
+
+// ClientOptions Options returns the options the client was constructed with.
+func (r *RedfishBaseBMC) ClientOptions() Options {
+	return r.options
```
stefanhipfel, line 174: name could be confusing, as ClientOptions are not options for the redfish client but for the BMC's construction.

bmc/redfish.go
```diff
@@ -129,18 +138,36 @@ func NewRedfishBMCClient(ctx context.Context, options Options) (BMC, error) {
...
+	if vendors == nil {
+		vendors = DefaultVendors()
 	}
+	if factory, ok := vendors[Manufacturer(manufacturer)]; ok {
+		if factory == nil {
+			return nil, fmt.Errorf("nil vendor factory registered for manufacturer %q", manufacturer)
+		}
+		return factory(base), nil
+	}
+	return base, nil
```
stefanhipfel, line 162: maybe drop a log here, that the base client was used:

` "No vendor factory registered for manufacturer, using base Redfish implementation"`

bmc/bmc.go
```diff
@@ -149,12 +172,68 @@ type BMC interface {
...
+	BIOSManager
+	BMCSettingsManager
+	FirmwareUpdater
+	ManagerController
+	AccountManager
+	EventSubscriber
+	Logouter
+}
+
+// VendorFactory wraps a *RedfishBaseBMC in a vendor-specific implementation.
```
stefanhipfel, line 205: we could split the file in 2. for example put VendorFactory into a new file called vendor.go? better separation

afritzler:
@stefanhipfel thanks a lot for your feedback. I addressed your comments. Can you please have another look?

stefanhipfel (approved):

PR #972, opened 2026-06-25
Add `--default-discovered-server-reclaim-policy` flag

### LINKED ISSUE #971: Expose default reclaimPolicy as a controller flag
## Context

`Server.spec.reclaimPolicy` controls whether a server transitions to `Released` after use (`Retain`) or directly back to `Available` (`Recycle`, the current default).

The metal-maintenance-operator sanitization flow requires servers to pass through `Released` state — meaning `Retain` must be set on every `Server` object for sanitization to trigger.

## Problem

There is no way to set a cluster-wide default `reclaimPolicy`. It must be patched on every individual `Server` object, which is not manageable at scale.

## Request

Expose `--default-server-reclaim-policy` (or similar) as a controller manager flag, so it can be set via the metal-operator helm chart values and applied as the default for all servers.

## Reference

- [`Server.spec.reclaimPolicy`](https://github.com/ironcore-dev/metal-operator/blob/main/api/v1alpha1/server_types.go#L116)
- [metal-maintenance-operator](https://github.com/ironcore-dev/metal-maintenance-operator) — depends on `Retain` to trigger sanitization

xkonni:
Adds \`--default-discovered-server-reclaim-policy\` to \`BMCReconciler\` (default: \`Recycle\`).

When the BMC controller discovers a new server, it sets \`spec.reclaimPolicy\` from this flag — but only when the field is not already set, preserving any manually configured value.

Operators running metal-maintenance-operator for sanitization can set this to \`Retain\` to ensure servers transition through the \`Released\` state without patching every \`Server\` object individually.

Closes #971

## Summary by CodeRabbit

* **New Features**
  * Added a new startup flag to set the default reclaim policy for newly discovered servers.
  * Discovered servers without a reclaim policy now automatically receive the configured default.
* **Bug Fixes**
  * Startup now validates the reclaim policy value and stops the process if an unsupported option is provided.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -90,7 +90,7 @@ func main() { // nolint: gocyclo
 		webhookPort                        int
 		enforceFirstBoot                   bool
 		enforcePowerOff                    bool
-		defaultServerReclaimPolicy         string
+		serverDiscoveryReclaimPolicy       string
 		discoveryIgnitionPath              string
 		serverResyncInterval               time.Duration
 		maintenanceResyncInterval          time.Duration
@@ -160,9 +160,9 @@ func main() { // nolint: gocyclo
 		"Enforce the first boot probing of a Server even if it is powered on in the Initial state.")
 	flag.BoolVar(&enforcePowerOff, "enforce-power-off", false,
 		"Enforce the power off of a Server when graceful shutdown fails.")
-	flag.StringVar(&defaultServerReclaimPolicy, "default-server-reclaim-policy",
-		string(metalv1alpha1.ServerReclaimPolicyRecycle),
-		"Default reclaim policy for servers when spec.reclaimPolicy is unset (Recycle or Retain).")
+	flag.StringVar(&serverDiscoveryReclaimPolicy, "server-discovery-reclaim-policy", "",
+		"Reclaim policy to set on servers at discovery time (Recycle or Retain). "+
+			"If unset, the API-level default (Recycle) applies.")
 	flag.IntVar(&webhookPort, "webhook-port", 9443, "The port to use for webhook server.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
@@ -192,6 +192,16 @@ func main() { // nolint: gocyclo
 	flag.Parse()
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	if serverDiscoveryReclaimPolicy != "" {
+		switch metalv1alpha1.ServerReclaimPolicy(serverDiscoveryReclaimPolicy) {
+		case metalv1alpha1.ServerReclaimPolicyRecycle, metalv1alpha1.ServerReclaimPolicyRetain:
+		default:
+			setupLog.Error(nil, "Invalid server discovery reclaim policy, must be 'Recycle' or 'Retain'",
+				"serverDiscoveryReclaimPolicy", serverDiscoveryReclaimPolicy)
+			os.Exit(1)
+		}
+	}
 
 	// Handle protocol and TLS certificate validation flags with backward compatibility
 	var effectiveProtocol metalv1alpha1.ProtocolScheme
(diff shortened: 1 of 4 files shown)


dist/chart/values.yaml
```diff
@@ -12,6 +12,7 @@ controllerManager:
       - "--leader-elect"
       - "--metrics-bind-address=:8443"
       - "--health-probe-bind-address=:8081"
+      - "--default-server-reclaim-policy=Recycle"
```
afritzler: We don't need to set the flag to `Recycle` here because that will be defaulted.

cmd/main.go
```diff
@@ -159,6 +160,9 @@ func main() { // nolint: gocyclo
 		"Enforce the first boot probing of a Server even if it is powered on in the Initial state.")
 	flag.BoolVar(&enforcePowerOff, "enforce-power-off", false,
 		"Enforce the power off of a Server when graceful shutdown fails.")
+	flag.StringVar(&defaultServerReclaimPolicy, "default-server-reclaim-policy",
```
afritzler: I would call the flag maybe something like: `server-discovery-replaim-policy` Since we have a default defined in the type itself.

internal/controller/server_controller.go
```diff
@@ -504,7 +505,11 @@ func (r *ServerReconciler) handleReservedState(ctx context.Context, bmcClient bm
 			return modified, err
 		}
 
-		switch server.Spec.ReclaimPolicy {
+		reclaimPolicy := server.Spec.ReclaimPolicy
+		if reclaimPolicy == "" {
+			reclaimPolicy = r.DefaultReclaimPolicy
+		}
```
afritzler: This one is not needed.

afritzler (changes requested):
We just need to add a flag which can actually be a bool to put all servers when detected by the BMCReconciler into Retain mode. The rest is handled by defaulting it.

internal/controller/server_controller.go
```diff
@@ -504,7 +505,11 @@ func (r *ServerReconciler) handleReservedState(ctx context.Context, bmcClient bm
...
+		switch reclaimPolicy {
```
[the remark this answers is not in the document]
xkonni: The +kubebuilder:default=Recycle marker on spec.reclaimPolicy is intentional — it ensures newly created Servers have well-defined behaviour regardless of operator configuration. The flag covers the remaining case: objects where reclaimPolicy is genuinely empty (objects created before the default was introduced, or manually cleared). Removing the API default to make the flag apply universally would be a breaking change — existing clusters relying on Recycle without setting the field explicitly would silently change behaviour if the operator flag is set to Retain. If cluster-wide enforcement for all objects is needed, that belongs to a mutating webhook, not a controller flag.

internal/controller/bmc_controller.go
```diff
@@ -60,6 +60,9 @@ type BMCReconciler struct {
 	// DNSRecordTemplatePath is the path to the file containing the DNSRecord template.
 	DNSRecordTemplate string
 	Conditions        *conditionutils.Accessor
+	// DefaultDiscoveredServerReclaimPolicy sets spec.reclaimPolicy on newly discovered servers.
+	// If empty, the API-level default (Recycle) applies.
```
afritzler, line 64: This comment can be removed here.

afritzler (changes requested):
Can we also update the documentation describing the flag and it's purpose?

afritzler:
After thinking about this a bit here: Wouldn't a readiness check would be a better fit here? E.g. following problem: I want my hypervisor `Server` to have a `Retain` policy and the storage node in `Recycle` (no wipe) since I will take care of that myself.

xkonni:
Closing in favour of ironcore-dev/metal-maintenance-operator#96. After discussion, label-selector-based `ReclaimPolicy` assignment is a better fit for metal-maintenance-operator, which already has the per-server context (roles, labels) needed to make per-class decisions. A single cluster-wide default in metal-operator is too coarse.

PR #978, opened 2026-07-01
Updating webhooks for maintenance plan

asergeant01:
## Summary

This PR updates the validating webhooks for `BIOSSettings`, `BIOSVersion`, `BMCSettings`, and
`BMCVersion` to support the upcoming **metal-maintenance-operator**, which introduces a
`MaintenancePlanRun` controller that creates these CRs programmatically to drive firmware and
configuration upgrades across a fleet of servers.

### Problem

The existing webhook duplicate-detection logic blocked any two CRs from targeting the same server
or BMC ref, regardless of version. This was too strict for the maintenance plan use case, where the
operator needs to create a new CR for a _different target version_ on the same server — for example,
to upgrade BIOS firmware from `P71` to `P79`. The old logic would reject this as a duplicate.

Additionally, the webhooks were needlessly re-running the full duplicate check on every update even
when the `serverRef`/`bmcRef` field hadn't changed, which could never introduce a new duplicate and
was wasted work.

### Changes

**`biossettings_webhook.go`, `biosversion_webhook.go`, `bmcsettings_webhook.go`, `bmcversion_webhook.go`**

- **Dual-condition duplicate detection**: a CR is now only rejected as a duplicate when both the
  server/BMC ref **and** the target `version` match an existing record. Two CRs targeting the same
  server with different versions are explicitly allowed, enabling the maintenance plan operator to
  stage sequential upgrades.

- **Early return on unchanged ref**: `ValidateUpdate` now skips the duplicate check entirely when
  the `serverRef`/`bmcRef` has not changed between old and new object, since no new duplicate can
  be introduced in that case.

**`bmcversion_controller.go`**

- Minor: removes redundant error wrapping in the `resetBMC` call path (the inner error already
  contains sufficient context).

### Tests

All four webhook test files have been updated:

- Fixed existing tests that assumed the old single-condition duplicate logic
- Added new cases covering: deny when both ref and version match; allow when ref changes but version
  differs; allow when ref is unchanged but version changes

### Context

The `metal-maintenance-operator` `MaintenancePlanRun` controller creates `BIOSSettings`,
`BIOSVersion`, `BMCSettings`, and `BMCVersion` CRs on behalf of a `MaintenancePlan`. Each stage in
a plan targets a specific `version`, and multiple stages across a plan run may reference the same
server/BMC with different versions. Without this change, the webhooks would reject those creates and
block the maintenance pipeline entirely.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined webhook validation so updates are only blocked when both the reference and version match an existing entry.
  * Allowed updates to keep the same reference without being incorrectly flagged as duplicates.
  * Improved error handling during BMC reconciliation to avoid returning misleading errors when no real failure occurred.

### DIFF
--- a/internal/controller/biossettings_controller.go
+++ b/internal/controller/biossettings_controller.go
@@ -1023,6 +1023,9 @@ func (r *BIOSSettingsReconciler) handleAppliedState(ctx context.Context, bmcClie
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to get BIOS version and settings diff: %w", err)
 	}
+	if settings.Status.State == metalv1alpha1.BIOSSettingsStateApplied && len(settingsDiff) == 0 {
+		return ctrl.Result{}, nil
+	}
 	if len(settingsDiff) > 0 {
 		log.V(1).Info("Found BIOS setting difference after applied state", "SettingsDiff", settingsDiff)
 		return ctrl.Result{}, r.updateStatus(ctx, settings, metalv1alpha1.BIOSSettingsStatePending, nil)
@@ -1490,9 +1493,7 @@ func (r *BIOSSettingsReconciler) enqueueBiosSettingsByServerRefs(ctx context.Con
 		if settings.Spec.ServerMaintenanceRef == nil ||
 			(server.Spec.ServerMaintenanceRef != nil &&
 				(server.Spec.ServerMaintenanceRef.Name != settings.Spec.ServerMaintenanceRef.Name ||
-					server.Spec.ServerMaintenanceRef.Namespace != settings.Spec.ServerMaintenanceRef.Namespace)) ||
-			settings.Status.State == metalv1alpha1.BIOSSettingsStateApplied ||
-			settings.Status.State == metalv1alpha1.BIOSSettingsStateFailed {
+					server.Spec.ServerMaintenanceRef.Namespace != settings.Spec.ServerMaintenanceRef.Namespace)) {
 			continue
 		}
 		reqs = append(reqs, ctrl.Request{NamespacedName: types.NamespacedName{Name: settings.Name}})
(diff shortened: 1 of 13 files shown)


internal/controller/biossettings_controller.go
```diff
@@ -1023,6 +1023,9 @@ func (r *BIOSSettingsReconciler) handleAppliedState(ctx context.Context, bmcClie
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to get BIOS version and settings diff: %w", err)
 	}
+	if settings.Status.State == metalv1alpha1.BIOSSettingsStateApplied && len(settingsDiff) == 0 {
+		return ctrl.Result{}, nil
+	}
```
nagadeesh-nagaraja, line 1028: what is the logic for this?? when handleAppliedState function is called, it is already in `BIOSSettingsStateApplied `state. and we return `ctrl.Result{}, nil` anyway if settingsDiff ==0

internal/webhook/v1alpha1/biossettings_webhook.go
```diff
@@ -69,6 +69,13 @@ func (v *BIOSSettingsCustomValidator) ValidateUpdate(ctx context.Context, oldObj
 		}
 	}
 
+	// If neither the serverRef nor the version changed, no new duplicate can be introduced.
+	if oldObj.Spec.ServerRef != nil && newObj.Spec.ServerRef != nil &&
+		oldObj.Spec.ServerRef.Name == newObj.Spec.ServerRef.Name &&
```
nagadeesh-nagaraja, line 74: i think ServerRef is immutable :D so this check is redundant.

this logic basically says that version needs to change everytime i update the spec??? dont seem to be right.

afritzler:
@asergeant01 can we close this PR as the maintenance plan topic has now been moved to the maintenance operator project?

afritzler:
Closing as `ServerMaintenance` + `BMC*` and `BIOS*` resource have been moved to the metal-maintenance-operator.

PR #990, opened 2026-07-09
Complete all `ServerMaintenances` before removing the maintenance approval

### LINKED ISSUE #989: ServerMaintenances delete the approval without considering other waiting mentenances
**Which component is affected?**
 Server, ServerClaim, ServerMaintenance

**Describe the bug**
When multiple serverMaintenances are waiting on the server. approval provided needs to be used for all waiting maintenances before deleting the approval. 
We have a mechanism to handle multiple maintenances in available state to reserve state transition, where we wait for all maintenances to complete before claiming the server and moving to reserved state.  however, this is not honored in case of owner approval.

**To reproduce**
1. create multiple owner approval maintenances on a reserved server.
2. on the serverclaim, approve the maintenances. 
3. check that after one serverMaintenance the approval is removed and others wait for re-approval

**Expected behavior**
All waiting serverMaintenances needs to run through before joining the production cluster. that way the disruption of services are minimal. 

**Actual behavior**
one maintenance finishes, and deletes the approval on the serverclaim. leaving rest of the maintenances unfinished. 

**Environment**
- metal-operator version/commit: latest
- Kubernetes version (`kubectl version`): latest
- Deployment method (Helm / kustomize / local `make run`): NA
- Hardware / BMC type (if relevant): NA

nagadeesh-nagaraja:
# Proposed Changes
consistent behavior when multiple maintenances are in queue when server is in available/reserved state

Fixes #989 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved maintenance handling so queued maintenance requests are processed in sequence without briefly leaving Maintenance between items.
  * Prevented maintenance labels from being cleared too early when other pending maintenances still exist.
  * Preserved the server’s current state more reliably during maintenance handoffs, including reserved and approval-based workflows.

* **Tests**
  * Updated and expanded automated coverage for queued maintenance scenarios to verify continuous Maintenance behavior and correct label cleanup.

### DIFF
--- a/internal/controller/server_controller.go
+++ b/internal/controller/server_controller.go
@@ -600,10 +600,6 @@ func (r *ServerReconciler) hasPendingMaintenances(ctx context.Context, server *m
 		}
 	}
 
-	// For OwnerApproval maintenances, only stay in Maintenance if they can actually proceed.
-	// With no ServerClaim, they are auto-approved (same as Enforced). With a ServerClaim,
-	// check that the approval label is still present — if revoked externally, exit Maintenance
-	// to avoid getting stuck.
 	if pendingOwnerApproval {
 		if server.Spec.ServerClaimRef == nil {
 			return true, nil
@@ -626,16 +622,12 @@ func (r *ServerReconciler) hasPendingMaintenances(ctx context.Context, server *m
 func (r *ServerReconciler) handleMaintenanceState(ctx context.Context, bmcClient bmc.BMC, server *metalv1alpha1.Server) (bool, error) {
 	log := ctrl.LoggerFrom(ctx)
 	if server.Spec.ServerMaintenanceRef == nil {
-		// Before exiting maintenance, check if other maintenances are queued for this server.
-		// If so, stay in Maintenance so the next one can claim the ref without a Reserved -> maintenance state bounce.
 		hasPending, err := r.hasPendingMaintenances(ctx, server)
 		if err != nil {
 			return false, fmt.Errorf("failed to check pending maintenances: %w", err)
 		}
 		if hasPending {
 			log.V(1).Info("Other maintenances are pending, staying in Maintenance state", "Server", server.Name)
-			// Refresh system info between maintenances so the next one sees up-to-date
-			// hardware state (e.g. BIOS version after a firmware update).
 			if err := r.updateServerStatusFromSystemInfo(ctx, bmcClient, server); err != nil {
 				return false, fmt.Errorf("failed to update server status system info between maintenances: %w", err)
 			}
--- a/internal/controller/server_controller.go
+++ b/internal/controller/server_controller.go
@@ -580,9 +580,68 @@ func (r *ServerReconciler) handleReleasedState(ctx context.Context, bmcClient bm
 	return r.patchServerState(ctx, server, metalv1alpha1.ServerStateAvailable)
 }
 
+func (r *ServerReconciler) hasPendingMaintenances(ctx context.Context, server *metalv1alpha1.Server) (bool, error) {
+	maintenanceList := &metalv1alpha1.ServerMaintenanceList{}
+	if err := r.List(ctx, maintenanceList, client.MatchingFields{serverRefField: server.Name}); err != nil {
+		return false, fmt.Errorf("failed to list ServerMaintenances: %w", err)
+	}
+
+	var pendingOwnerApproval bool
+	for i := range maintenanceList.Items {
+		m := &maintenanceList.Items[i]
+		if !m.DeletionTimestamp.IsZero() {
+			continue
+		}
+		if m.Spec.Policy == metalv1alpha1.ServerMaintenancePolicyEnforced {
+			return true, nil
+		}
+		if m.Spec.Policy == metalv1alpha1.ServerMaintenancePolicyOwnerApproval {
+			pendingOwnerApproval = true
+		}
+	}
+
+	// For OwnerApproval maintenances, only stay in Maintenance if they can actually proceed.
+	// With no ServerClaim, they are auto-approved (same as Enforced). With a ServerClaim,
+	// check that the approval label is still present — if revoked externally, exit Maintenance
+	// to avoid getting stuck.
+	if pendingOwnerApproval {
+		if server.Spec.ServerClaimRef == nil {
+			return true, nil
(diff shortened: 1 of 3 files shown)


internal/controller/server_controller.go
```diff
@@ -580,9 +580,68 @@ func (r *ServerReconciler) handleReleasedState(ctx context.Context, bmcClient bm
...
+			return false, fmt.Errorf("failed to check pending maintenances: %w", err)
+		}
+		if hasPending {
+			log.V(1).Info("Other maintenances are pending, staying in Maintenance state", "Server", server.Name)
+			// Refresh system info between maintenances so the next one sees up-to-date
+			// hardware state (e.g. BIOS version after a firmware update).
+			if err := r.updateServerStatusFromSystemInfo(ctx, bmcClient, server); err != nil {
+				return false, fmt.Errorf("failed to update server status system info between maintenances: %w", err)
+			}
+			return false, nil
```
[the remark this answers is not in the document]
nagadeesh-nagaraja: workflow is:
when maintenance is done, its ref is removed on server..servermaintenance is deleted. 
on servercontroller side, it checks other maintenance pending and ready to takeover, so waits..
setbootOverride will be called when the next servermaintenance patches itself and its spec will be used for next

atd9876 (approved):
LGTM

internal/controller/server_controller.go
```diff
@@ -580,9 +580,68 @@ func (r *ServerReconciler) handleReleasedState(ctx context.Context, bmcClient bm
 	return r.patchServerState(ctx, server, metalv1alpha1.ServerStateAvailable)
 }
 
+func (r *ServerReconciler) hasPendingMaintenances(ctx context.Context, server *metalv1alpha1.Server) (bool, error) {
```
afritzler, line 583: We stick with the approval labels regardless of the maintenance state. E.g. if the maintenance object which is still around has completed.

internal/controller/server_controller.go
```diff
@@ -580,9 +580,68 @@ func (r *ServerReconciler) handleReleasedState(ctx context.Context, bmcClient bm
...
+		}
+		if m.Spec.Policy == metalv1alpha1.ServerMaintenancePolicyOwnerApproval {
+			pendingOwnerApproval = true
+		}
+	}
+
+	// For OwnerApproval maintenances, only stay in Maintenance if they can actually proceed.
+	// With no ServerClaim, they are auto-approved (same as Enforced). With a ServerClaim,
+	// check that the approval label is still present — if revoked externally, exit Maintenance
+	// to avoid getting stuck.
```
afritzler: Can we drop this comment as this is not exactly describing what the code below is doing?

internal/controller/servermaintenance_controller.go
```diff
@@ -428,18 +428,36 @@ func (r *ServerMaintenanceReconciler) cleanup(ctx context.Context, maintenance *
...
-		}
-		serverClaimBase := serverClaim.DeepCopy()
-		metautils.DeleteLabels(serverClaim, []string{
-			metalv1alpha1.ServerMaintenanceApprovedLabelKey,
-			metalv1alpha1.ServerMaintenanceNeededLabelKey,
-		})
-		if err := r.Patch(ctx, serverClaim, client.MergeFrom(serverClaimBase)); err != nil {
-			return fmt.Errorf("failed to patch ServerClaim annotations: %w", err)
+		serverMaintenancesList := &metalv1alpha1.ServerMaintenanceList{}
+		if err := clientutils.ListAndFilter(ctx, r.Client, serverMaintenancesList, func(object client.Object) (bool, error) {
```
afritzler: Why not using `MatchingFields{serverRefField: server.Name}` instead of filtering over all maintenance objects? Of course make sure that the index setup has been done correctly.

afritzler (changes requested):

internal/controller/server_controller.go
nagadeesh-nagaraja: dont think I understand this comment. Could you please help me understand?
this is needed so that the server will not switch between maintenance -> reserved -> maintenance in short period of time.

afritzler:
In the `hasPending` branch of `handleMaintenanceState`, the server stays in Maintenance but `SetBootOverride` is never called. The normal reconcile path re-asserts the persistent network boot override to guard against non-operator reboots during maintenance. Could we add the same `SetBootOverride` call before returning from the `hasPending` branch to maintain that safety guarantee between queued maintenances? Also, the `cleanup` path uses `clientutils.ListAndFilter` while `hasPendingMaintenances` uses the `serverRefField` index. Consider using `MatchingFields` in `cleanup` as well for consistency and to avoid scanning all ServerMaintenances on every cleanup.

internal/controller/server_controller.go
```diff
@@ -580,9 +580,64 @@ func (r *ServerReconciler) handleReleasedState(ctx context.Context, bmcClient bm
...
+	}
+
+	return false, nil
+}
+
 func (r *ServerReconciler) handleMaintenanceState(ctx context.Context, bmcClient bmc.BMC, server *metalv1alpha1.Server) (bool, error) {
 	log := ctrl.LoggerFrom(ctx)
 	if server.Spec.ServerMaintenanceRef == nil {
+		// Before exiting maintenance, check if other maintenances are queued for this server.
+		// If so, stay in Maintenance so the next one can claim the ref without a Reserved -> maintenance state bounce.
```
afritzler: Comment not needed as the code is self explanatory here.

internal/controller/server_controller.go
```diff
@@ -580,9 +580,64 @@ func (r *ServerReconciler) handleReleasedState(ctx context.Context, bmcClient bm
...
+		hasPending, err := r.hasPendingMaintenances(ctx, server)
+		if err != nil {
+			return false, fmt.Errorf("failed to check pending maintenances: %w", err)
+		}
+		if hasPending {
+			log.V(1).Info("Other maintenances are pending, staying in Maintenance state", "Server", server.Name)
+			// Refresh system info between maintenances so the next one sees up-to-date
+			// hardware state (e.g. BIOS version after a firmware update).
```
afritzler: Same here.

afritzler (changes requested):

afritzler (approved):

PR #998, opened 2026-07-13
Keep BootSourceOverrideTarget as Pxe when clearing override

afritzler:
# Proposed Changes

HPE iLO rejects setting BootSourceOverrideEnabled while the target is unset/None, which surfaced when the maintenance override stopped being re-asserted every reconcile. Keep Pxe (ignored once Disabled) to satisfy iLO's ordering requirement.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved boot override clearing compatibility with HPE iLO systems.
  * Disabling a boot override now preserves PXE as the boot target instead of sending an unsupported “None” value.

### DIFF
--- a/bmc/redfish_hpe.go
+++ b/bmc/redfish_hpe.go
@@ -20,6 +20,36 @@ type HPERedfishBMC struct {
 	*RedfishBaseBMC
 }
 
+// hpeClearedBootOverride disables the boot override using Pxe as the target.
+var hpeClearedBootOverride = schemas.Boot{
+	BootSourceOverrideEnabled: schemas.DisabledBootSourceOverrideEnabled,
+	// HPE iLO rejects setting BootSourceOverrideEnabled while the target is
+	// unset/None. The target is irrelevant once Disabled, so keep Pxe to satisfy
+	// iLO's ordering requirement.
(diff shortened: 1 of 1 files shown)


bmc/redfish.go
```diff
@@ -76,7 +76,9 @@ type RedfishBaseBMC struct {
 
 var clearedBootOverride = schemas.Boot{
 	BootSourceOverrideEnabled: schemas.DisabledBootSourceOverrideEnabled,
-	BootSourceOverrideTarget:  schemas.NoneBootSource,
+	// Keep Pxe instead of None: HPE iLO rejects setting BootSourceOverrideEnabled
+	// when the target is unset/None. The target is irrelevant once Disabled.
+	BootSourceOverrideTarget: schemas.PxeBootSource,
```
nagadeesh-nagaraja: I would recommend moving this to redfish_hpe.go 

overriding in the base for a HPE problem will create issue when it fails in other vendor :D

xkonni (approved):

bmc/redfish.go
afritzler: Good point. Addressed that issue.

nagadeesh-nagaraja (approved):

PR #999, opened 2026-07-13
Make bmcutils package public

### LINKED ISSUE #991: Move BMCUtils package to public
BMCUtils package

Since we are moving the user management (BMCUser) to the metal-maintenance-operator, we need the same utils function there as well. Hence we should move our BMCUtils out of internal.

**Describe the feature**
Make BMCUtils package public, so the maintenance-operator can import it as well.

stefanhipfel:
Moves `internal/bmcutils` to the repo root so it can be imported by the metal-maintenance-operator.

Did not move it to pkg/bmcutils, since we do not use pkg yet in this repo.

Fixes #991

## Summary by CodeRabbit

* **New Features**
  * Introduced shared BMC utilities for credential extraction from Kubernetes secrets, endpoint/address resolution, protocol-specific BMC client creation, optional registry configuration, and optional connectivity checks.
  * Added support for deriving server names from BMC details and performing SSH-based BMC reset operations.
* **Build Improvements**
  * Updated the container build to include the local `pkg/` utilities in the builder stage.
* **Refactor**
  * Standardized BMC handling to use the shared utilities across console, controllers, and command tooling.


afritzler:
How about moving this to the `bmc` package?

stefanhipfel:
> How about moving this to the `bmc` package?

Should we maybe move those 2 into `pkg`, since we already have `internal`? If we put it in bmc, I would then call the package just `utils`.

afritzler:
`pkg/bmcutils` would make a valid alternative IMO.

afritzler (approved):

PR #985, opened 2026-07-03
Omit BootSourceOverrideMode on Supermicro BMCs that don't advertise it

afritzler:
# Proposed Changes

Omit BootSourceOverrideMode on Supermicro BMCs that don't advertise it.

Fixes #982

PR #1000, opened 2026-07-13
Revert maintenance boot-override persistence and ClearBootOverride

afritzler:
# Proposed Changes

The server controller no longer sets a persistent network-boot override on maintenance entry or clears boot overrides on available/reserved/maintenance-exit. `SetBootOverride` collapses back to a single-shot signature and `ClearBootOverride` is removed from the BootController interface and all implementations.

Reverts the persistence feature from #939 and the follow-ups that built on it (#985 persistent no-op, #996 set-once-at-entry, #998 HPE ClearBootOverride).

## Summary by CodeRabbit

* **Behavior Changes**
  * PXE boot overrides now apply only to the next power-on.
  * Removed support for persistent boot overrides and explicit override clearing.
  * Maintenance transitions no longer automatically set or clear boot overrides.
  * Local and hardware-specific Redfish implementations now follow the one-time PXE boot behavior.
* **Tests**
  * Updated boot override coverage to verify one-time PXE boot configuration.

### DIFF
--- a/bmc/bmc.go
+++ b/bmc/bmc.go
@@ -59,16 +59,10 @@ type PowerController interface {
 
 // BootController manages boot overrides and boot order for a system.
 type BootController interface {
-	// SetBootOverride configures the system to network-boot on its next
-	// power-on, bypassing the persistent boot order. If persistent is false
-	// the override applies to a single boot only; if true it applies to every
-	// subsequent boot until ClearBootOverride is called.
-	SetBootOverride(ctx context.Context, systemURI string, persistent bool) error
-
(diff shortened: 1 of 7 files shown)


nagadeesh-nagaraja (approved):

xkonni (approved):
lgtm

nagadeesh-nagaraja:
just thought of this, if these are required and its working as expected in supermicro computers, we can just disable the two methods for all other vendors for now and re write them as required later :)

afritzler:
No they are actually not necessary for Supermicro.

PR #1039, opened 2026-07-22
Require BMC-confirmed PowerState=Off before releasing a ServerClaim

fwiesel:
## Summary

Split into two independently-green commits:

**Commit 1 — `54612f9` Safety: require BMC-confirmed `PowerState=Off` before release**

When a claim was deleted, `ensureServerPoweredOffWithoutBootConfig` set
`Spec.Power=Off` and cleared `Spec.BootConfigurationRef`, then on the next
reconcile short-circuited on those spec fields alone and released the
ServerClaim — even if the BMC still reported `PowerState=Unknown`. The
Server could be handed to the next tenant without confirming the host was
powered down.

Gates the reclaim/release paths on the observed `Status.PowerState=Off`.
Adds `MockServer.ForcePowerState` to simulate a stuck BMC in tests.

**Commit 2 — `4bc3808` Observability: report the wait, stop blocking a worker**

When the BMC cannot confirm `PowerState=Off`, a claimed Server correctly
stays bound, but nothing showed why — the stuck claim looked like an
operator bug and had to be diagnosed from controller logs.

Adds a dedicated `WaitingForPowerOff` condition (visible via
`kubectl describe server`, `Ready` left untouched). Issues `PowerOff`
without blocking in `WaitForServerPowerState` on the reclaim/release
paths so an unhealthy BMC no longer holds a controller worker for the
full polling timeout.

## Summary by CodeRabbit

* **New Features**
  * Added clearer status reporting while a server is waiting for the BMC to confirm power-off.
  * Servers now remain in their current lifecycle state until power-off is confirmed, preventing premature release or reassignment.
  * Once confirmation is received, normal cleanup, release, and future claims proceed automatically.

* **Bug Fixes**
  * Improved recovery from BMCs that accept power commands but do not immediately update their reported power state.
  * Prevented stale server claim information from being restored during maintenance updates.

### DIFF
--- a/bmc/mock/server/server.go
+++ b/bmc/mock/server/server.go
@@ -158,6 +158,11 @@ type MockServer struct {
 	actionHandlers    []actionHandler       // ordered POST action dispatch table (first match wins)
 	onCreate          map[string]memberHook // collection URL suffix → hook called after a member is added
 	onDelete          map[string]memberHook // collection URL suffix → hook called before a member is removed
+	// stuckPowerState pins the PowerState of a given system path to a specific
+	// value. Subsequent doPowerOff / doPowerOn / doPowerReset invocations on
+	// that path leave PowerState untouched, modelling an unhealthy BMC that
+	// accepts reset commands but never converges to the requested state.
+	stuckPowerState map[string]string
 }
 
 // loadAccountsFromEmbedded seeds the authentication store by reading the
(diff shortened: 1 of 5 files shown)


afritzler:
I would propose to solve this problem a little bit simpler: https://github.com/ironcore-dev/metal-operator/pull/1048

afritzler:
Closing as this has been addressed by https://github.com/ironcore-dev/metal-operator/pull/1048

PR #1042, opened 2026-07-22
Deprecate Server.spec.power field

### LINKED ISSUE #1041: Deprecate `Server.spec.power`
## Problem

`Server.spec.power` is the field that drives the desired power state of a server. Today it is a `spec` field that mixes two distinct concerns, and is written almost exclusively by controllers rather than by users:

- **Controller-internal desired state.** The `ServerReconciler` writes it as it walks the state machine: `PowerOn` during discovery, `PowerOff` when a claim is released/recycled, and `PowerOff` during the initial/first-boot guard.
- **Surrogate input from owning resources.** The `ServerClaimReconciler` copies `claim.Spec.Power` onto `server.Spec.Power`, and the `ServerMaintenanceReconciler` copies `maintenance.Spec.ServerPower` onto `server.Spec.Power`.

Because it lives on `spec`, every such controller write is a `Patch` against the `Server` object; a write to the user-facing desired state that races with the actual owner of the `Server` and re-triggers the server reconciler. The field is effectively an internal actuator variable exposed as user-facing API.

The only genuine observed value is `status.powerState` (the `ServerPowerState` reported by the BMC). `spec.power` does not add a user-facing knob that isn't already expressed more precisely by the
`ServerClaim`, `ServerMaintenance`, and server state machine.

afritzler:
# Proposed Changes

Deprecate Server.spec.power field.

Fixes #1041

## Summary by CodeRabbit

* **Documentation**
  * Marked the server `power` field as deprecated across the API schema, reference docs, and Go API docs.
  * Clarified it is a controller-managed internal actuator value that may race with the actual observed power state.
  * Added guidance to use server claims, maintenance settings, and the server status/state-machine for power management.
  * Warned that new server resources should not set this field, as it will be removed in a future release.


afritzler:
@coderabbitai please create an issue to nilify the `spec.power` field on the `Server` resource and remove any logic ensuring the power state via this field.

adracus (approved):

PR #1058, opened 2026-07-27
Wire metaldata DaemonSet image through values

### LINKED ISSUE #1057: helm: metaldata DaemonSet image is hardcoded and not configurable via values — helm install always fails
## Problem

The `metaldata` DaemonSet template in `dist/chart/templates/extras/metaldata.yaml`
is generated by `kubebuilder edit --plugins=helm/v2-alpha` with a hardcoded image
reference and no `imagePullPolicy` field:

```yaml
image: metaldata:latest
# no imagePullPolicy
```

This causes two problems:

1. **Kubernetes always pulls from Docker Hub** — when `imagePullPolicy` is absent
   and the tag is `latest`, Kubernetes defaults the policy to `Always`. The image
   `docker.io/library/metaldata:latest` does not exist on Docker Hub. Every
   `helm install` fails with:

   ```
   Failed to pull image "metaldata:latest": pull access denied, repository does not exist
   ```

2. **The image is not configurable via Helm values** — there are no
   `metaldata.image.*` values in `values.yaml`. Users cannot set the repository,
   tag, or pull policy via `--set` or a custom `values.yaml`. This is inconsistent
   with `manager.image.*` which is fully configurable.

## Impact

- Every `helm install` from `ghcr.io/ironcore-dev/charts/metal-operator` fails
  on the `metaldata` DaemonSet
- The `dist/install.yaml` raw `kubectl apply` path also lacks `imagePullPolicy`
- Air-gapped and private registry deployments have no way to redirect the image
- The `publish-chart.yml` CI workflow does not set the metaldata image at publish
  time, so published charts have no correct image reference

## Root Cause

`kubebuilder helm/v2-alpha` generates values templating only for the primary
controller manager image. Secondary workloads placed in `config/metaldata/` are
copied verbatim into `dist/chart/templates/extras/` with no values wiring. The
plugin has no mechanism to override this behaviour — a post-generation patch in
the `make helm` Makefile target is required.

## Expected Behaviour

The metaldata image should be configurable via Helm values and default to the
correct GHCR repository, matching the pattern used by `manager.image.*`:

```yaml
metaldata:
  image:
    repository: ghcr.io/ironcore-dev/metaldata
    # tag defaults to Chart.AppVersion
    pullPolicy: IfNotPresent
```

And `publish-chart.yml` should set both `manager.image.*` and `metaldata.image.*`
at publish time so the published chart is immediately deployable.

atd9876:
# Proposed Changes

- Add post-generation `sed`/`printf` patches to `make helm` that wire the `metaldata`
  DaemonSet image through `values.yaml` after every `kubebuilder` run (including `--force`)
- Add `metaldata.image.repository` and `metaldata.image.tag` to `publish-chart.yml`
  so published charts deploy the correct image

Fixes #1057

## Summary by CodeRabbit

* **New Features**
  * Helm charts now support configurable Metaldata image repository, tag (defaulting to the chart app version), and pull policy.
  * Published charts automatically carry the release-specific Metaldata image settings.

* **Bug Fixes**
  * Improved Helm chart packaging to remove the hardcoded Metaldata image reference and image pull policy, ensuring generated values are applied consistently.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -262,6 +262,18 @@ undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.
 .PHONY: helm
 helm: manifests kubebuilder
 	"$(KUBEBUILDER)" edit --plugins=helm/v2-alpha
+	@sed -i '/^        imagePullPolicy: /d' \
+	  dist/chart/templates/extras/metaldata.yaml
+	@grep -qF '        image: metaldata:latest' \
+	  dist/chart/templates/extras/metaldata.yaml || { \
+	  echo "unexpected metaldata template; image patch not applied" >&2; exit 1; }
+	@sed -i \
+	  's|image: metaldata:latest|image: "{{ .Values.metaldata.image.repository }}:{{ .Values.metaldata.image.tag \| default .Chart.AppVersion }}"\n        {{- with .Values.metaldata.image.pullPolicy }}\n        imagePullPolicy: {{ . }}\n        {{- end }}|' \
+	  dist/chart/templates/extras/metaldata.yaml
+	@awk '/^metaldata:/{f=1} f && /^[ \t]+image:/{found=1; exit} END{exit !found}' \
+	  dist/chart/values.yaml || \
+	  printf '\n## Configure the metaldata DaemonSet\n##\nmetaldata:\n  image:\n    repository: ghcr.io/ironcore-dev/metaldata\n    # tag: ""\n    pullPolicy: IfNotPresent\n' \
+	  >> dist/chart/values.yaml
 
 ##@ Dependencies
 
--- a/.github/workflows/publish-chart.yml
+++ b/.github/workflows/publish-chart.yml
@@ -85,6 +85,8 @@ jobs:
           IMAGE_TAG=sha-$(echo ${{ github.sha }} | cut -c1-7)
           yq -i '.manager.image.repository = "ghcr.io/ironcore-dev/metal-operator-controller-manager"' dist/chart/values.yaml
           yq -i '.manager.image.tag = "'$IMAGE_TAG'"' dist/chart/values.yaml
+          yq -i '.metaldata.image.repository = "ghcr.io/ironcore-dev/metaldata"' dist/chart/values.yaml
+          yq -i '.metaldata.image.tag = "'$IMAGE_TAG'"' dist/chart/values.yaml
 
       - name: Package Helm chart with crds folder in template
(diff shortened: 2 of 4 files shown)


atd9876:
@coderabbitai review

asergeant01 (approved):

Makefile
```diff
@@ -262,6 +262,18 @@ undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.
 .PHONY: helm
 helm: manifests kubebuilder
 	"$(KUBEBUILDER)" edit --plugins=helm/v2-alpha
+	@sed -i '/^        imagePullPolicy: /d' \
```
afritzler, line 265: Why exactly do we need this one here?

afritzler (changes requested):

Makefile
atd9876: It line removes the static imagePullPolicy: Always line that kustomize emits into metaldata.yaml

Without this deletion, the generated template would contain both the static line from kustomize and the values-driven block we inject in the next sed step, resulting in a duplicate.

afritzler: Why not just add this to: https://github.com/ironcore-dev/metal-operator/blob/main/config/metaldata/daemonset.yaml

atd9876: The sed deletion is needed because of how helm/v2-alpha handles non-primary workloads (anything in templates/extras/). The plugin copies these verbatim from the kustomize output — it generates values templating only for the primary controller manager image. For metaldata, the plugin emits a static imagePullPolicy field from the kustomize source, and we then inject a values-driven {{- with .Values.metaldata.image.pullPolicy }} block in its place so users can override it.

If we only set imagePullPolicy in daemonset.yaml without the sed patch, the chart would have a hardcoded imagePullPolicy: IfNotPresent with no way to override it via Helm values. The sed deletion removes the static field so only the values-driven block remains.

This is a limitation of the helm/v2-alpha plugin for secondary workloads — it doesn't generate values wiring for them the way it does for the manager. This was noted in the issue https://github.com/ironcore-dev/metal-operator/issues/1057

afritzler: Ok got it. Well, lets do it for now that way. However it feels like we might think about 2 options moving forward/follow up:
1. Extend the helm plugin upstream to support a more flexible chart generation for extra components (hard)
2. Use a tool like https://github.com/arttor/helmify to generate a own chart for the `metaldata` component

afritzler (approved):

PR #1072, opened 2026-08-03
Add parallel operator instances via --watch-filter flag and label-scoped cache

### LINKED ISSUE #1079: Support Parallel Operator Instances via Watch-Filter
## Summary

Support running **multiple metal-operator instances against a single API server**, where each instance reconciles only the objects that match its ownership filter. This is a pure predicate/label filter: there is no automatic assignment, no hashing, no rebalancing. An instance either owns objects carrying its label value (`metal.ironcore.dev/watch-filter: <name>`) or, when unconfigured, objects carrying no watch-filter label at all.

Instances must not interfere with each other: no duplicate reconciles, no fighting over status/labels/finalizers, no cross-instance BMC/Redfish
operations.

afritzler:
# Proposed Changes

Add parallel operator instances via --watch-filter flag and label-scoped cache.

Fixes #1079

## Summary by CodeRabbit

- **New Features**
  - Added watch-filter support for selecting resources by label.
  - Filtered controller instances reconcile matching resources and propagate filters to related resources.
  - Added filter validation and isolated leader-election identifiers for filtered instances.

- **Documentation**
  - Added configuration guidance, including the SSH reset timeout setting.
  - Added conceptual documentation covering filtering, multi-instance deployments, labeling, observability, and migration considerations.
  - Added navigation links to the new documentation pages.


internal/controller/bmcuser_controller.go
```diff
@@ -262,6 +262,7 @@ func (r *BMCUserReconciler) createBMCSecretForUser(ctx context.Context, user *me
 		Immutable: new(true),
 	}
 	op, err := controllerutil.CreateOrPatch(ctx, r.Client, secret, func() error {
+		metalv1alpha1.PropagateShardLabel(secret, user)
```
nikParasyr: I think this can be easier done through wrapping the client (check issue)

nikParasyr (commented):
I didn't go in depth. My input is:
- propagation is probably easier done through wrapping the client, that will also ensure that we always propagate the shard label on all metal-operator resources created by a shard
- in the current implementation, BiosSettingSet (and others) are not updated to propagate:
  - not sure if this is intentionally left out because they will be moved to maintanance operator
  - but currently for example BiosSettingSet is only catched by the default shard that doesnt see servers on other shards

internal/controller/bmcuser_controller.go
afritzler: I would keep the implementation as simple as possible for now. The `*Set` resource are deprecated and will move out of the `metal-operator` with the next release. Since we only stamp out 4 resources as child objects, it is easier to do it via those 4 helper calls instead of maintaining an embedded client.

api/v1alpha1/constants.go
```diff
@@ -70,6 +70,12 @@ const (
 	// when set on it.
 	OperationAnnotationRotateCredentials = "rotate-credentials"
 
+	// WatchFilterLabel selects the operator instance that owns a resource. An
+	// instance started with --watch-filter=<value> only watches and reconciles
+	// resources carrying this label with that value. An instance without
+	// --watch-filter only handles resources that do not carry the label at all.
+	WatchFilterLabel = "metal.ironcore.dev/watch-filter"
```
nikParasyr, line 77: CAPI's label is called watch-filter because it really is a watch/event filter — it's applied through predicates ([ref1](https://github.com/kubernetes-sigs/cluster-api/blob/e4e236c764cf8a520deeda830667edb229311329/core/main.go#L571), [ref2](https://github.com/kubernetes-sigs/cluster-api/blob/e4e236c764cf8a520deeda830667edb229311329/core/reconcilers/cluster/cluster_controller.go#L119))
Here we're doing a cache filter instead, which has different semantics (it affects reads/what's in the cache, not just triggering), so watch-filter would be a bit misleading.
If `shard` was an "abstract terminology" that anyone can put their own meaning and therefore better not to use (and probably rightly so) then something like `cache-filter` would be better.

internal/controller/bmcuser_controller.go
nikParasyr: My reasoning for the wrapper was:
- I was counting all the *Set resources, so I had 10+ call sites rather than 4 — I didn't know they were on their way out.
- Consistency across repos. I hadn't gone through maintenance-operator or boot-operator, but assumed a similar size, and a shared wrapper would give all three the same mechanism.
- Helpers are opt-in per call site, so a child resource added six months from now silently produces an unstamped object — which lands in the default shard rather than erroring. Catching that relies on review and tests.

That said, at 4 call sites the wrapper is more machinery than this needs, so I'm happy with the current approach. Easy enough to revisit if the count grows or the other repos end up diverging.

nikParasyr (commented):
I see that you are not propagating the label on the ServerBootConfiguration which is used afaik by the boot-operator to create HTTP/PXE boot configs.
Is that because boot-operator will be removed later on or you consider this "cross controller" watch-filter to be out-of-scope?

api/v1alpha1/constants.go
afritzler: The predicate approach gates events, not reads: with having a lot of reconcilers each instance would still cache every object and could Get/List foreign resources (e.g. a claim selecting another instance's server), plus a forgotten. `WithEventFilter` on one future controller silently breaks isolation. The cache selector makes mis-scoped access structurally impossible and drops memory/watch traffic to the instance's share. Acknowledged difference vs CAPI's mechanics. 

I kept the watch-filter name for the operator-facing contract since users know it from CAPI. The real consequence of cache scoping is finalizer handover on relabel; I'll document that relabels must happen after          a finalizer cleanup. Happy to rename to `--cache-filter` if the label name trips people up, though the CAPI familiarity was the point.

nikParasyr:
@afritzler what is the status of this (have been away for quite a while)?
From my side it is LGTM

adracus (approved):

PR #1088, opened 2026-08-07
Refactor BMCVersionSet/BIOSVersionSet: dedupe-safe child creation and leaner deletion paths

afritzler:
# Proposed Changes

Refactor BMCVersionSet/BIOSVersionSet: dedupe-safe child creation and leaner deletion paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of BIOS and BMC version resources during deletion.
  * Excluded servers and BMCs marked for deletion from selection results.
  * Improved reconciliation when server or BMC labels change.
  * Prevented unnecessary template updates for versions with active upgrades.
  * Improved handling of orphaned version resources and ownership changes.

* **Improvements**
  * Child resource names are generated consistently and remain valid within Kubernetes naming limits.
  * Improved retry handling when resources are removed during reconciliation.

### DIFF
--- a/internal/controller/biosversionset_controller.go
+++ b/internal/controller/biosversionset_controller.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/ironcore-dev/controller-utils/clientutils"
 	metalv1alpha1 "github.com/ironcore-dev/metal-operator/api/v1alpha1"
+	metalutil "github.com/ironcore-dev/metal-operator/internal/util"
 )
 
 const (
@@ -41,6 +42,7 @@ type BIOSVersionSetReconciler struct {
 // +kubebuilder:rbac:groups=metal.ironcore.dev,resources=biosversionsets/finalizers,verbs=update
 // +kubebuilder:rbac:groups=metal.ironcore.dev,resources=biosversions,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=metal.ironcore.dev,resources=servers,verbs=get;list;watch
+// +kubebuilder:rbac:groups=metal.ironcore.dev,resources=servermaintenances,verbs=get;list;watch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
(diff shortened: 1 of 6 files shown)


nagadeesh-nagaraja:
why are we making so many changes to the deprecated resources in metal-operator?

isnt it better to make this changes in metal-mentainance-operator directly once its ported?

internal/controller/biosversionset_controller.go
```diff
@@ -63,35 +61,20 @@ func (r *BIOSVersionSetReconciler) reconcileExists(ctx context.Context, versionS
...
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to get owned BIOSVersions: %w", err)
 	}
-
-	status := r.getOwnedBIOSVersionSetStatus(versions)
-	if status.AvailableBIOSVersion != (status.CompletedBIOSVersion+status.FailedBIOSVersion) ||
-		versionSet.Status.AvailableBIOSVersion != status.AvailableBIOSVersion {
-		if err = r.patchStatus(ctx, status, versionSet); err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to patch BIOSVersionSet status: %w", err)
-		}
```
nagadeesh-nagaraja, line 85: why do not want to report the latest status on the SET Resouce? 
the deletion might be blocked for long time and hence to better to show the updated status here.

internal/controller/biosversionset_controller.go
```diff
@@ -272,7 +208,8 @@ func (r *BIOSVersionSetReconciler) patchBIOSVersionFromTemplate(ctx context.Cont
 	var pendingPatchingVersion bool
 	var errs []error
 	for _, version := range versions.Items {
-		if version.Status.State == metalv1alpha1.BIOSVersionStateInProgress {
+		if version.Status.State == metalv1alpha1.BIOSVersionStateInProgress && version.Status.UpgradeTask != nil {
+			log.V(1).Info("Skipping BIOSVersion spec patching as it is InProgress with an active UpgradeTask", "BIOSVersion", version.Name)
```
nagadeesh-nagaraja: the biosVersion controller does active operation ( like resetting the BMC) when maintenance is approved.

In my opinion, its safer to stop once the maintenance is approved rather than task created. this method is also consistent across controllers, easier to follow rather than do different logic in different controller.

internal/controller/biosversionset_controller.go
afritzler: The set's finalizer is removed immediately now, so the set disappears quickly. There's no long deletion to report status on. The wait moved to the children, which keep their own status while their finalizer holds them during active maintenance. This also fixes the old failure mode where a stuck InProgress child made the set undeletable forever. I would keep this one as is since this is just a UX change (let the user know) but for a very short period of time.

internal/controller/biosversionset_controller.go
afritzler: Agreed. The child is `InProgress` with `UpgradeTask == nil` through approval and the BMC reset, so the current gate allows template changes mid-flight. I'll change it to stop patching once maintenance is approved instead of waiting for the task.

internal/controller/biosversionset_controller.go
nagadeesh-nagaraja, line 199: this logic will throw error rather than gracefully waiting, because the webhook throws error when you try to delete object which is in active Inprogress state. I would keep this logic to avoid unnecessary errored reconcile.

internal/controller/biosversionset_controller.go
nagadeesh-nagaraja: i see a tracking/debugging problem with force removing the SET object while leaving the child hanging...
 
take following case: 
1. SET was created 
2. Child got created by SET
3. Child where is progress
4. I deleted the SET and created new SET
5. now we have old child which might be stuck inProgress state which is harder to track and delete and clean up
6. the new SET object will face issue because already exisiting abandoned child is actively holding the server. 


if the SET was not forced deleted, it would show why it was not deleted and gives clear indication on how many child are InProgress blocking the clean delete.

nagadeesh-nagaraja:
I still feel this PR is better candicate to directly go to the metal-mentainance operator. 
I am at the very end to create PR for porting the BIOS*/BMC* resources. we can just move this PR post that? I do not see the advantage in merging it here and then porting it to metal-maintenance-operator aswell?

afritzler:
@nagadeesh-nagaraja fair point. I am closing this PR here in the metal-operator and we might want to revisit this topics in the maintenance operator later.

PR #1097, opened 2026-08-12
Extract ServerClaim scheduling into a dedicated scheduler controller

### LINKED ISSUE #1096: Factor Out ServerClaim Scheduling into a Dedicated Scheduler Controller
## Summary

When a claim picks its server via a label selector, the ServerClaim reconciler currently does the placement itself: it scans all servers, filters out the ones that don't match or aren't claimable, and grabs the first suitable one. That's scheduling work sitting in the middle of a controller whose actual job is the claim lifecycle.

We should move placement into its own ServerClaim scheduler controller.

## Proposal

The scheduler watches claims that aren't bound to a server yet. It evaluates selectors, checks whether a server is claimable (state, power, maintenance, conflicting claims), claims the best match, and records the binding. Since a binding is immutable once made, the scheduler is done with a claim the moment it is bound. No rescheduling, no eviction handling.

The scheduler completes the binding in both directions: it claims the server (with optimistic locking) and then sets the server reference on the claim. That way the claim reconciler never has to search for the server. It reads the reference and gets on with the lifecycle. The two writes can't be atomic, so the scheduler must be idempotent: if it already claimed the server for this claim, it just completes the claim-side reference on the next pass.

The claim reconciler stops doing any of that. It only sees claims that already point at a server and takes care of boot configuration, power state, and cleanup.

## Why

- Two concerns, two controllers. Each loop gets smaller and easier to reason about.
- Placement becomes the natural home for future scheduling improvements (priorities, spreading, cordon awareness) without touching the claim lifecycle.
- The scheduler can be scaled, tuned, or disabled on its own.

## Separation of concerns

The two controllers stay out of each other's way through event predicates. The scheduler only observes claims without a server reference; it ignores everything already bound. The claim reconciler gets the inverse filter and only observes bound claims. Since a binding is immutable, a claim crosses the boundary exactly once, from scheduler to reconciler, and never goes back. Both filters must use the same definition of "bound" so no claim is missed or handled by both.

## Acceptance criteria

- A dedicated scheduler controller binds unbound claims to servers.
- The claim reconciler no longer contains any server selection or selector matching logic.
- Existing test coverage for the scheduling behavior is moved over and passes.

afritzler:
# Proposed Changes

Extract ServerClaim scheduling into a dedicated scheduler controller.

Fixes #1096 

## Summary by CodeRabbit

* **New Features**
  * Added automatic scheduling of server claims to eligible servers using explicit references or label selectors.
  * Added concurrent scheduling while preventing duplicate server assignments.
  * Claims now wait for assigned servers to confirm reservation before boot configuration and power actions proceed.

* **Bug Fixes**
  * Improved handling of unavailable, cordoned, tainted, maintained, powered-on, or already-claimed servers.
  * Added safer conflict handling during concurrent assignment.
  * Preserved claim state and restored power settings when servers are parked and returned to service.
  * Prevented changes to server assignments after claims are bound.


afritzler:
@coderabbitai review

internal/controller/index_fields.go
```diff
@@ -30,6 +31,32 @@ func RegisterIndexFields(ctx context.Context, indexer client.FieldIndexer) error
 		return err
 	}
 
+	if err := indexer.IndexField(ctx, &metalv1alpha1.Server{}, serverClaimRefField, func(rawObj client.Object) []string {
+		server, ok := rawObj.(*metalv1alpha1.Server)
+		if !ok {
+			return nil
+		}
+		if ref := server.Spec.ServerClaimRef; ref != nil {
+			return []string{ref.Namespace + "/" + ref.Name}
```
adracus: How about using
```suggestion
			return []string{(client.ObjectKey{Namespace: ref.Namespace, Name: ref.Name}).String()}
```

(same when doing the actual `List` with the `FieldSelector`

adracus (approved):
Minor improvement, otherwise good to go!

PR #1138, opened 2026-08-26
Remove OwnerReference from BMC to Server

### LINKED ISSUE #1137: Remove BMC OwnerReference from discovered Servers
BMC management, server discovery

Discovered `Server`s are owned by their `BMC` via a controller OwnerReference. Deleting or recreating a BMC garbage collects all its Servers and takes their claims, maintenance state, and labels down with them. 

**Describe the feature**

A Server is physical inventory. It exists whether or not a BMC object is around, and its identity is already tied to the BMC via `Spec.BMCRef`. The Server should outlive its BMC: deleting a BMC must not cascade.

**Proposed API or behavior changes**

- No OwnerReference between `BMC` and `Server`. `Server.Spec.BMCRef` is the link.
- Deleting a `BMC` leaves its `Server`s in place.
- Consider a condition on `Server` when its BMC is gone, so orphaned boxes aren't silently stale.

**Additional context**

Blocked use cases: BMC replacement/migration without wiping inventory.

afritzler:
# Proposed Changes

Remove OwnerReference from BMC to Server.

Fixes #1137 

## Summary by CodeRabbit

* **New Features**
  * Added startup migrations to safely update existing server resources.
  * Added an option to skip migrations during startup.

* **Changes**
  * Server resources are no longer automatically owned by BMC resources.
  * BMC reconciliation no longer manages event subscriptions or deletion finalizers.
  * BMC updates now continue to be triggered when referenced servers change.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -42,6 +42,7 @@ import (
 	"github.com/ironcore-dev/metal-operator/internal/api/macdb"
 	"github.com/ironcore-dev/metal-operator/internal/controller"
 	metalmetrics "github.com/ironcore-dev/metal-operator/internal/metrics"
+	"github.com/ironcore-dev/metal-operator/internal/migration"
 	"github.com/ironcore-dev/metal-operator/internal/registry"
 	// +kubebuilder:scaffold:imports
 )
@@ -102,6 +103,7 @@ func main() { // nolint: gocyclo
 		serverClaimMaxConcurrentReconciles int
 		dnsRecordTemplatePath              string
 		defaultFailedAutoRetryCount        int
+		skipMigrations                     bool
 	)
 
 	flag.IntVar(&serverMaxConcurrentReconciles, "server-max-concurrent-reconciles", 5,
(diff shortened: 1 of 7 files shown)


stefanhipfel:
Thanks @afritzler. would it make sense to add a label to the servers created by the bmc controller for faster lookup?
so one could find all servers that are related to a bmc via: ` r.List(ctx, serverList, client.MatchingLabels{"metal.ironcore.dev/bmc-name": bmc.Name})`

afritzler:
> Thanks @afritzler. would it make sense to add a label to the servers created by the bmc controller for faster lookup? so one could find all servers that are related to a bmc via: ` r.List(ctx, serverList, client.MatchingLabels{"metal.ironcore.dev/bmc-name": bmc.Name})`

We have a `fieldIndex` for the `bmcRef` for `Server` objects. So the lookup will be super fast.

adracus (commented):
Why don't we add something like migrations as done in `ironcore-net`:
https://github.com/ironcore-dev/ironcore-net/blob/main/utils/migration/migration.go

This way, the migration happens only once during startup and not during each reconciliation:
https://github.com/ironcore-dev/ironcore-net/blob/main/cmd/apinetlet/main.go#L309

afritzler:
> Why don't we add something like migrations as done in `ironcore-net`: https://github.com/ironcore-dev/ironcore-net/blob/main/utils/migration/migration.go
> 
> This way, the migration happens only once during startup and not during each reconciliation: https://github.com/ironcore-dev/ironcore-net/blob/main/cmd/apinetlet/main.go#L309

That is actually a good idea. Maybe we should also handle the `Endpoint` ownerref removal in the same way then.

adracus (approved):

PR #1106, opened 2026-08-18
BMC Update for Dell FIrmware upgrade

nagadeesh-nagaraja:
# Proposed Changes

- Add Dell-specific Redfish BMC methods: `InstallFirmwareFromRepository`, `GetRepositoryUpdateList`, `ListJobs`, `GetJob`
- Introduce `dellRepositoryUpdateRequestBody` with correct field encodings (`ApplySameVersions`/`ApplyDowngradeVersions` as `"True"`/`"False"`, `IgnoreCertWarning` as `"On"`/`"Off"`)
- Treat all `InstallFromRepository` POST errors as fatal (request may have reached BMC before connection dropped)
- Fix SSH reset fallback in `BMCReconciler`: schedule SSH on any connectivity failure, not only `*schemas.Error` 5xx (gofish returns plain-text 503 without wrapping)
- Extend mock server: Dell repository-update state machine, job step simulation, `WithResourceOverride`/`WithSystemOverride` options
- Add unit tests: field encoding, dropped-connection fatal error, mock server end-to-end flow
- Fix `errcheck` lint: assign `conn.Close()` return value in test hijacker

# Fixes

Fixes #1105

PR #1142, opened 2026-08-27
Add FirmwareUpdaterDell interface and Dell Redfish implementation

### LINKED ISSUE #1105: BMC package update to support new API calls for dell firmware upgrade
sub issue for: [#169](https://github.com/ironcore-dev/metal-maintenance-operator/issues/169)

support DELL OME firmware upgrade API calls in the BMC package

xkonni:
contains minor linter fix

- Replaces #1106 
- Fixes #1105

## Summary by CodeRabbit

- **New Features**
  - Added support for Dell repository-based firmware updates, including package detection, installation, update listings, and job tracking.
  - Added visibility into Dell update job progress, completion, and failure states.
- **Bug Fixes**
  - Improved handling of Dell firmware attributes and update responses.
  - Added support for varied completion-value formats and recognized catalog mismatch responses.
- **Tests**
  - Added comprehensive coverage for Dell repository updates and job lifecycle scenarios, including successful and failed updates.

### DIFF
--- a/bmc/redfish_dell.go
+++ b/bmc/redfish_dell.go
@@ -675,17 +675,13 @@ func (j *DellJob) IsCompleted() bool {
 	return j.State == dellJobStateCompleted
 }
 
-// IsFailed reports whether the job has reached a failed terminal state, either
-// via a known failure JobState or via failure-indicating text in Message
-// (mirroring Dell's reference scripting examples, which classify jobs by
-// substring-matching Message rather than JobState alone).
+// IsFailed reports whether the job has reached a known terminal failure state.
 func (j *DellJob) IsFailed() bool {
 	switch j.State {
 	case dellJobStateFailed, dellJobStateCompletedWithErrors, dellJobStateRebootFailed:
 		return true
 	}
-	msg := strings.ToLower(j.Message)
-	return strings.Contains(msg, "fail") || strings.Contains(msg, "invalid") || strings.Contains(msg, "unable")
+	return false
 }
 
 // IsTerminal reports whether the job has reached any terminal state, success
@@ -919,9 +915,6 @@ func (r *DellRedfishBMC) ListJobs(_ context.Context, UUID string) ([]string, err
 	if err != nil {
 		return nil, fmt.Errorf("failed to read jobs collection response body: %w", err)
 	}
-	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("failed to list jobs %v, statusCode %v", string(rawBody), resp.StatusCode)
-	}
 
 	var collection dellJobsCollection
 	if err := json.Unmarshal(rawBody, &collection); err != nil {
@@ -954,9 +947,6 @@ func (r *DellRedfishBMC) GetJob(_ context.Context, UUID string, jobID string) (*
 	rawBody, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read job response body: %w", err)
-	}
-	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("failed to get job %v, statusCode %v", string(rawBody), resp.StatusCode)
 	}
 
 	var dj dellJob
--- a/bmc/redfish_dell.go
+++ b/bmc/redfish_dell.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"maps"
 	"net/http"
+	"path"
 	"strings"
 
 	"github.com/stmcginnis/gofish/schemas"
@@ -455,8 +456,6 @@ func (r *DellRedfishBMC) CheckBMCAttributes(ctx context.Context, bmcUUID string,
 	return checkAttributes(attrs, filteredAttr)
 }
 
-// --- Firmware upgrade overrides ---
-
 func (r *DellRedfishBMC) dellBuildRequestBody(parameters *schemas.UpdateServiceSimpleUpdateParameters) *SimpleUpdateRequestBody {
 	body := &SimpleUpdateRequestBody{}
 	body.RedfishOperationApplyTime = schemas.ImmediateOperationApplyTime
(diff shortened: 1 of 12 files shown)


internal/controller/bmc_controller.go
```diff
@@ -267,7 +267,7 @@ func (r *BMCReconciler) discoverServers(ctx context.Context, bmcClient bmc.BMC,
 			server.Spec.SystemUUID = strings.ToLower(s.UUID)
 			server.Spec.SystemURI = s.URI
 			server.Spec.BMCRef = &corev1.LocalObjectReference{Name: bmcObj.Name}
-			return nil
+			return controllerutil.SetControllerReference(bmcObj, server, r.Scheme)
```
atd9876: Setting a controller owner-reference from BMC → Server changes deletion semantics.

Today BMCReconciler.delete() is a no-op, so deleting a BMC leaves its Server orphaned. After this change, deleting a BMC will cascade-delete the Server (GC sets deletionTimestamp, then the ServerFinalizer cleanup runs). That's arguably the correct lifecycle, and it's also what drives the Or(Succeed(), IsNotFound) / IgnoreNotFound churn in the controller tests — but it's a meaningful operational change that isn't mentioned in the PR description.

bmc/redfish_dell.go
```diff
@@ -561,3 +561,415 @@ func (r *DellRedfishBMC) dellMatchesComponentFilter(fw *schemas.SoftwareInventor
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close() // nolint: errcheck
+
+	rawBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read jobs collection response body: %w", err)
+	}
+	if resp.StatusCode != http.StatusOK {
```
atd9876: These resp.StatusCode != http.StatusOK checks are effectively dead code with gofish v0.24.0: client.Get funnels through runRequestWithHeaders, which already converts any non-2xx/204 into a returned error (with a nil *http.Response) — so if we get here, err was nil and the status is necessarily 200. GetRepositoryUpdateList in this same file relies on that behavior (inspecting the error via errors.As), so this manual check is inconsistent with the sibling method. Consider dropping both blocks (here and in GetJob at L958)

internal/controller/bmc_controller.go
afritzler: That part here needs to go as we decoupled the ownerref from BMC and Server.

afritzler (changes requested):

bmc/redfish_dell.go
```diff
@@ -561,3 +561,415 @@ func (r *DellRedfishBMC) dellMatchesComponentFilter(fw *schemas.SoftwareInventor
...
+// via a known failure JobState or via failure-indicating text in Message
+// (mirroring Dell's reference scripting examples, which classify jobs by
+// substring-matching Message rather than JobState alone).
+func (j *DellJob) IsFailed() bool {
+	switch j.State {
+	case dellJobStateFailed, dellJobStateCompletedWithErrors, dellJobStateRebootFailed:
+		return true
+	}
+	msg := strings.ToLower(j.Message)
+	return strings.Contains(msg, "fail") || strings.Contains(msg, "invalid") || strings.Contains(msg, "unable")
```
atd9876: The Message substring fallback is broad: strings.Contains(msg, "fail"/"invalid"/"unable") is a case-insensitive substring can misclassify jobs — e.g. a still-Running job whose message contains "unable to reach share, retrying" is reported as failed (the "Running state with a failure message" → true test documents this), and innocent phrasing like "no failures" / "no invalid packages" matches too.

bmc/redfish_dell.go
```diff
@@ -561,3 +561,415 @@ func (r *DellRedfishBMC) dellMatchesComponentFilter(fw *schemas.SoftwareInventor
...
+	IgnoreCertWarning bool
+	// ApplyUpdate, if true, applies the update; if false, performs a dry-run check only.
+	ApplyUpdate bool
+	// RebootNeeded, if true, allows the BMC to reboot the system to apply updates that require it.
+	RebootNeeded bool
+	// ApplySameVersions, if true, re-applies packages already at the same version.
+	ApplySameVersions bool
+	// ApplyDowngradeVersions, if true, allows applying packages older than the currently installed version.
+	ApplyDowngradeVersions bool
+}
```
davidgrun, line 592: we should think about ClearJobQueueBeforeUpdate bool or something like this. Dell iDRAC maintains a job queue that can accumulate completed/failed jobs over time. A full job queue can prevent new firmware updates from being triggered or cause unexpected behavior. on idrac9+ ist should be possible via /redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellJobService/Actions/DellJobService.DeleteJobQueue

bmc/redfish_dell.go
```diff
@@ -561,3 +561,415 @@ func (r *DellRedfishBMC) dellMatchesComponentFilter(fw *schemas.SoftwareInventor
...
+	resp, err := client.Post(target, map[string]any{})
+	if err != nil {
+		// As in InstallFirmwareFromRepository, gofish surfaces non-2xx/204
+		// responses as a *schemas.Error rather than a response we can inspect.
+		// Treat a "no catalog match" message or the specific iDRAC SUP099 error
+		// (no applicable updates found) as "nothing pending" rather than a hard failure.
+		var redfishErr *schemas.Error
+		if errors.As(err, &redfishErr) {
+			msg := strings.ToLower(redfishErr.Message)
+			if strings.Contains(msg, "match catalog") {
```
davidgrun, line 736: should be checked in live environment strings.Contains(msg, "not found") || strings.Contains(eiMsg, "match catalog")

atd9876 (approved):
Lgtm

bmc/redfish_dell.go
```diff
@@ -561,3 +561,415 @@ func (r *DellRedfishBMC) dellMatchesComponentFilter(fw *schemas.SoftwareInventor
...
+// Known Dell iDRAC JobState values. Dell does not publish a formal enum for
+// this field; these are the values commonly observed on iDRAC OEM Jobs and
+// used by Dell's reference scripting examples. Verify against real hardware
+// before relying on this exhaustively.
+const (
+	dellJobStateCompleted           = "Completed"
+	dellJobStateCompletedWithErrors = "CompletedWithErrors"
+	dellJobStateFailed              = "Failed"
+	dellJobStateRebootFailed        = "RebootFailed"
+)
```
davidgrun, line 613: should we think about the dellJobStatePending = "Pending"
some of the jobs will be planed with this and we should consider how we deal with the situation if this job state do not change. Just to ensure the controller runs not with an  "forever update"

bmc/redfish_dell.go
xkonni: Good idea for a follow-up but out of scope for this PR. The current PR adds the basic install-from-repository flow; job queue management is a separate concern.

bmc/redfish_dell.go
xkonni: The broad "not found" substring was already removed in this PR. The handler now only swallows IDRAC.2.1.SUP099 (via ExtendedInfos) and the "match catalog" substring. Happy to adjust "match catalog" further once we can verify on real hardware.

bmc/redfish_dell.go
xkonni: The Pending state already flows through verbatim via toRepositoryJob → RepositoryJob.State → FirmwareUpdate.status.checkJob.state, so it's visible in the CRD status today.

internal/controller/bmc_controller.go
```diff
@@ -489,19 +489,15 @@ func (r *BMCReconciler) resetBMC(ctx context.Context, bmcObj *metalv1alpha1.BMC,
 		return fmt.Errorf("failed to set BMC resetting condition: %w", err)
 	}
 	if bmcClient == nil {
-		// No Redfish client available. If the connection error was a 5xx (e.g. 503
```
afritzler, line 492: Should those changes be part of this PR? Looks like a rebase problem to me.

afritzler (changes requested):

internal/controller/bmc_controller.go
xkonni: this was part of the coderabbit review saying that "any connectivity failure should warrant a ssh retry" but I agree this is out of scope for this PR and will remove it.

afritzler (approved):

PR #1164, opened 2026-09-09
BMC interface: add FetchETags and write-only drift detection helpers

### LINKED ISSUE #1163: BMC Interface: Additions to support drift detection in BMCSettings
# Summary
Add ETag support to the BMC interface to enable efficient drift detection in the metal-maintenance-operator's BMCSettings controller. Rather than issuing a full value-map GET on every reconcile, the controller can use a lightweight ETag comparison to determine whether any attribute has changed, falling back to a full GET only when needed.

# Motivation
HPE iLO and Lenovo XCC expose ETags on their BIOS/BMC settings resources. The metal-maintenance-operator can use these to short-circuit reconciliation: if the ETag on a resource hasn't changed since the last apply, no attribute under that resource can have drifted, and the full GetBMCAttributeValues GET can be skipped entirely.

This also enables correct handling of two edge cases that are not addressable without ETag tracking:

Write-only attributes (e.g. passwords): not returned in GET responses, so a policy-based decision is needed on when to re-apply them. An ETag change on the owning resource is the signal that something may have drifted.
POST-created resources (e.g. event subscriptions): need idempotency tracking to avoid re-creating on every reconcile. The stored value fingerprint serves as the idempotency key.

# What this adds
FetchETags(ctx, uris) (map[string]string, error) on the BMC interface — returns the current ETag for each resource URI, or nil if ETags are not available for this vendor
IsPost bool on ApplyResult — flags resources created via POST so callers know the URI is ephemeral and not suitable for ETag-based drift detection
Vendor implementations:
HPE / Lenovo: active ETag fetch via shared httpFetchETags helper
Dell / base Redfish: returns nil (controller falls back to existing full GET path)
Supporting mock server helpers (SetResourceETag, GetResourceETag, atomic saveResourceAndBumpETag) for testing drift scenarios

# Relationship to metal-maintenance-operator
This PR is the interface layer. The controller logic (classifying keys, computing diffs using ETags, handling write-only and POST-prefix keys) lives in a follow-on PR in the metal-maintenance-operator repository.

atd9876:
# Proposed Changes

bmc.go

Add FetchETags(ctx, uris) (map[string]string, error) to the BMC interface
Add IsPost bool to ApplyResult to flag POST-created ephemeral resources

redfish.go

Default FetchETags implementation returns nil (callers fall back to full value-map GET)

redfish_dell.go

FetchETags returns nil — Dell DellAttributes are a monolithic resource with no per-attribute ETag signal
SetBMCAttributesImmediately now captures URI + ETag per key after each successful PATCH
CheckBMCAttributes skips POST/PATCH-prefix keys (not Dell OEM registry entries)

redfish_hpe.go / redfish_lenovo.go

FetchETags overrides using shared httpFetchETags helper — HPE iLO and Lenovo XCC both expose ETags on BIOS settings resources

oem_helpers.go

httpFetchETags: shared GET-based ETag fetcher with proper status code handling (404 → empty string, non-2xx → error)
httpBasedUpdateBMCAttributes: pre-PATCH GET now fails fast on error/non-2xx rather than silently proceeding without If-Match; PATCH ETag fallback-GET on missing response header
httpBasedGetBMCSettingAttribute: returns nil for keys where all desired fields are present but null (write-only PATCH-prefix semantics)
checkAttributes: skips PasswordAttributeType and WriteOnly entries (value cannot be verified via GET)

redfish_local.go

SetBMCAttributesImmediately and FetchETags implementations for local/test Redfish

server.go

SetResourceETag / GetResourceETag / SetBMCSettingAttr helpers for test drift simulation
saveResourceAndBumpETag: atomic resource write + ETag bump under one mutex (eliminates GET/ETag race window)
bumpETag: handles both hyphen-delimited (res-3 → res-4) and bare numeric suffixes (v1 → v2)

Fixes #1163

## Summary by CodeRabbit

* **New Features**
  * BMC setting updates now report affected resources, current ETags, and whether changes were applied through temporary resources.
  * Added ETag retrieval where supported to help coordinate updates safely.
  * Immediate updates now provide detailed per-setting results across supported BMC platforms.

* **Bug Fixes**
  * Improved handling of write-only and password-protected settings.
  * Enhanced POST and PATCH processing with response validation and resource tracking.
  * Added safer handling for unavailable resources and unsupported ETag information.


atd9876:
@coderabbitai review

afritzler (approved):

PR #1172, opened 2026-09-18
Add upgrade guide for v0.7 to v0.8

afritzler:
# Proposed Changes

Add upgrade guide for v0.7 to v0.8.

## Summary by CodeRabbit

- **Documentation**
  - Added an Upgrades section to the documentation navigation.
  - Added general upgrade guidance, including release-note checks, resource cleanup, CRD handling, and finalizer considerations.
  - Added a v0.7-to-v0.8 migration guide covering removed resources and fields, scheduler changes, operator upgrade methods, CRD cleanup, and webhook configuration updates.


docs/upgrades/v0.7-to-v0.8.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+### 2. Delete the removed resources while v0.7 is still running
+
+Let in-flight BIOS/BMC upgrades and settings updates finish first, then delete all custom resources
+of the removed kinds:
+
+```bash
+kubectl delete bmcusers.metal.ironcore.dev --all
+kubectl delete servermaintenances.metal.ironcore.dev --all -A
```
nagadeesh-nagaraja, line 47: nit-pick: might be better to list this in the end of the list??? deleting parent (like biossettings/version) help auto clean up.. parent might recreate it if deleted earlier.

docs/upgrades/v0.7-to-v0.8.md
```diff
@@ -0,0 +1,115 @@
...
+kubectl delete biosversions.metal.ironcore.dev --all
+kubectl delete bmcsettingssets.metal.ironcore.dev --all
+kubectl delete bmcsettings.metal.ironcore.dev --all
+kubectl delete bmcversionsets.metal.ironcore.dev --all
+kubectl delete bmcversions.metal.ironcore.dev --all
+```
+
+Wait until all instances are gone. Owned children (e.g. `BIOSSettings` created by a
+`BIOSSettingsSet`) are garbage-collected automatically. If those children are still needed,
+recreate them standalone before deleting the sets.
```
nagadeesh-nagaraja, line 60: may be worth to mention that if they are actively performing update/upgrade, we either need to wait to complete or we need to force delete them (making sure it wont leave the physical server in weird state.

docs/upgrades/v0.7-to-v0.8.md
```diff
@@ -0,0 +1,115 @@
...
+  -o jsonpath='{range .items[?(@.status.state=="Maintenance")]}{.metadata.name}{"\n"}{end}'
+```
+
+### 3. Clean up your manifests
+
+Remove all manifests of the removed kinds from your GitOps/IaC repositories, and drop these fields
+from the remaining resources:
+
+- `spec.power` and `spec.biosSettingsRef` from `Server`
+- `spec.bmcSettingsRef` from `BMC`
```
nagadeesh-nagaraja, line 75: is there a way to make this automatic?? 
like the reconciler automatically removes/set them to null

may be with intermediate release 0.7.5 or something?

docs/upgrades/v0.7-to-v0.8.md
afritzler: The list already deletes each Set before its children, and the Set controller stops reconciling once the Set has a deletion timestamp, so it can't recreate children after that. The child deletes are also redundant: owner references let GC remove them automatically (noted in the paragraph below the commands). I kept them only to catch stragglers from already-deleted Sets, so I'd leave the order as is.

docs/upgrades/v0.7-to-v0.8.md
afritzler: Step 2 already starts with "let in-flight upgrades finish first", so the waiting part is covered. The force-delete caveat is fair, I'll add a note to the finalizer patch below that you should only strip finalizers once the update is done or safely aborted, otherwise you can leave the box mid-flash.

docs/upgrades/v0.7-to-v0.8.md
afritzler: That's exactly the problem: the manual step exists because of people's GitOps manifests, and no reconciler or intermediate release can touch those. The stored values on the objects already get pruned by the apiserver on the next write, so making the operator strip them wouldn't remove any manual work for users. I'd keep the guide as is.

adracus (approved):

